PR #31, opened 2022-10-06
fix provider address

marksoper:
_Add a description of the problem this PR addresses and an overview of how this PR works_.

**Checklist**

* [x] I have added these changes to the changelog (or it's not applicable).

PR #38, opened 2022-10-17
v0.2.0: Refactor cluster resources

erademacher:
- Consolidated create specs and update specs into configs
- Added some basic validation to Create and Update
- Removed the cluster data source in favor of the resource
- Updated and pruned examples
- Reran `go generate` to format examples and update docs
- Fixed debug address in main.go
- Bumped version to 0.2.0

The [TF design principles document](https://www.terraform.io/plugin/hashicorp-provider-design-principles#resource-and-attribute-schema-should-closely-match-the-underlying-api) says, "A Terraform resource and associated schema should follow the naming and structure of the API, unless the it degrades the user experience or works in a way counter to a user's expectations of Terraform". The cluster resource doesn't need to include both the create and upgrade specs, plus config, in a single object. It's cleaner and more idiomatic to combine these fields, which this PR attempts to do.

The most important changes are in cluster_resource.go.

Acceptance tests pass, but are very basic. I've also got some manual testing underway, but that involves creating and modifying dedicated clusters, which takes a while. Serverless seems to work as expected.

**Checklist**

* [ ] I have added these changes to the changelog (or it's not applicable).

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38)


erademacher:
@marksoper The commit template has a reference to a changelog, but I don't see one anywhere. Am I just missing it, or should I update the template?

fantapop (commented):
> Reviewed 1 of 19 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38)* status: 1 of 19 files reviewed, 4 unresolved discussions (waiting on @erademacher, @jenngeorge, @marksoper, @prafull01, and @sean-)

___
*[`main.go` line 55 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEbxs7R9TnNc9ajmwgR:-NEbxs7R9TnNc9ajmwgS:b-sb8j3i) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/main.go#L55)):*
> ```Go
> 	opts := providerserver.ServeOpts{
> 		Debug:   debug,
> 		Address: "registry.terraform.io/cockroachdb/cockroach",
> ```

Are you sure this is the right thing?  This url 404s.  The previous value which I believe @marksoper updated last week resolves.



_Code quote:_
```Go
registry.terraform.io/providers/cockroachdb/cockroach
```
___
*[`examples/workflows/cockroach_serverless_cluster/main.tf` line 30 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEbzIj0BN29HH7ZSCEZ:-NEbzIj0BN29HH7ZSCE_:b7oplc6) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/examples/workflows/cockroach_serverless_cluster/main.tf#L30)):*
> ```HCL
> }
> 
> variable "cloud_provider_region" {
> ```

cloud_provider_regions should be used here since this is a list. 
___
*[`examples/workflows/cockroach_serverless_cluster/main.tf` line 31 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEbz1gD1tqEQ9y5jW8a:-NEbz1gD1tqEQ9y5jW8b:b76gou4) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/examples/workflows/cockroach_serverless_cluster/main.tf#L31)):*
> ```HCL
> 
> variable "cloud_provider_region" {
>   type     = list(any)
> ```

can this be `list(string)` (syntax?) or something more specific?
___
*[`examples/workflows/cockroach_serverless_cluster/main.tf` line 56 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEc-bSD2-b6F8Cdpy6_:-NEc-bSD2-b6F8Cdpy6a:b-r5bqqv) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/examples/workflows/cockroach_serverless_cluster/main.tf#L56)):*
> ```HCL
>   regions = [
>     {
>       name = var.cloud_provider_region[0]
> ```

Would it be a better example if we transformed all of cloud_provider_regions into a list of regions instead of just picking the first one?

marksoper:
I think the changelog reference is just boilerplate that can be removed

marksoper:
*[`main.go` line 55 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEbxs7R9TnNc9ajmwgR:-NEdF-xO3JYtWwEkvkTm:b-5cakuk) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/main.go#L55)):*
> registry.terraform.io/providers/cockroachdb/cockroach
This looks like the correct URL, the new value as of last week's update. It does resolve for me, not sure why you're getting a 404 @fantapop

marksoper:
*[`main.go` line 55 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEbxs7R9TnNc9ajmwgR:-NEdFKdi9Nes1EYWCzdu:b-lzpnpo) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/main.go#L55)):*
<details><summary><i>Previously, marksoper (Mark Soper) wrote…</i></summary><blockquote>

> registry.terraform.io/providers/cockroachdb/cockroach
This looks like the correct URL, the new value as of last week's update. It does resolve for me, not sure why you're getting a 404 @fantapop 
</blockquote></details>

https://github.com/cockroachdb/terraform-provider-cockroach/pull/31/files

marksoper:
@erademacher By removing the data source are you thinking that `terraform import` will be the way to work with resources originally provisioned outside of Terraform?

marksoper:
*[`examples/workflows/cockroach_serverless_cluster/main.tf` line 56 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEc-bSD2-b6F8Cdpy6_:-NEdHdXa99hJtTGtJ0iO:b6yds6r) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/examples/workflows/cockroach_serverless_cluster/main.tf#L56)):*
<details><summary><i>Previously, fantapop (Christopher Fitzner) wrote…</i></summary><blockquote>

Would it be a better example if we transformed all of cloud_provider_regions into a list of regions instead of just picking the first one?
</blockquote></details>

It's Serverless so only a single region is supported.

marksoper:
<details><summary><i>Previously, marksoper (Mark Soper) wrote…</i></summary><blockquote>

@erademacher By removing the data source are you thinking that `terraform import` will be the way to work with resources originally provisioned outside of Terraform?
</blockquote></details>

The pattern I'm seeing with other providers is that a given resource has both a TF Resource and a TF Data Source - e.g. MongoDB Atlas https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/data-sources/advanced_cluster

@erademacher are you concerned about the reliability of the Data Source defs we have in place - if not we could leave them in place as a way to access CC resources in a read-only way if they're sure they don't want to bring the resource under TF management but do want to be able to refer to it.

erademacher (commented):
I added the data source back in. We should figure out what's goin on with the provider address.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-:-NEgwi9kBQg1U8c9fxzP:b4aihnl)* status: 1 of 21 files reviewed, 5 unresolved discussions (waiting on @fantapop, @jenngeorge, @marksoper, @prafull01, and @sean-)

*[a discussion](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEdFQ4d9vsaT1yKbWOx:-NEgx3Nx7OY_FeX14scC:bkdlquv) (no related file):*
<details><summary><i>Previously, marksoper (Mark Soper) wrote…</i></summary><blockquote>

The pattern I'm seeing with other providers is that a given resource has both a TF Resource and a TF Data Source - e.g. MongoDB Atlas https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/data-sources/advanced_cluster

@erademacher are you concerned about the reliability of the Data Source defs we have in place - if not we could leave them in place as a way to access CC resources in a read-only way if they're sure they don't want to bring the resource under TF management but do want to be able to refer to it.
</blockquote></details>

I've added it back in and updated the schema. Not sure why the cloud provider was marked as required before.

___
___
*[`main.go` line 55 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEbxs7R9TnNc9ajmwgR:-NEgjwEY95I3gOThZMVr:b-xco91l) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/main.go#L55)):*
<details><summary><i>Previously, marksoper (Mark Soper) wrote…</i></summary><blockquote>

https://github.com/cockroachdb/terraform-provider-cockroach/pull/31/files
</blockquote></details>

That's very weird, because unless I remove the "/providers", I get this:
```
unable to validate ServeOpts: unable to validate Address: expected hostname/namespace/type format, got: registry.terraform.io/providers/cockroachdb/cockroach
```
___
*[`examples/workflows/cockroach_serverless_cluster/main.tf` line 30 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEbzIj0BN29HH7ZSCEZ:-NEgwEmC1cOiWh0YmhYb:bt2vr1b) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/examples/workflows/cockroach_serverless_cluster/main.tf#L30)):*
<details><summary><i>Previously, fantapop (Christopher Fitzner) wrote…</i></summary><blockquote>

cloud_provider_regions should be used here since this is a list. 
</blockquote></details>

Fixed.
___
*[`examples/workflows/cockroach_serverless_cluster/main.tf` line 31 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEbz1gD1tqEQ9y5jW8a:-NEgwFheEOEB7kWKDHDI:b-896fix) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/examples/workflows/cockroach_serverless_cluster/main.tf#L31)):*
<details><summary><i>Previously, fantapop (Christopher Fitzner) wrote…</i></summary><blockquote>

can this be `list(string)` (syntax?) or something more specific?
</blockquote></details>

Done.
___
*[`examples/workflows/cockroach_serverless_cluster/main.tf` line 56 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEc-bSD2-b6F8Cdpy6_:-NEgwGwLEOy06YxuCM0U:b6pk1cn) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/examples/workflows/cockroach_serverless_cluster/main.tf#L56)):*
>It's Serverless so only a single region is supported.

That was my thinking, but the API spec does take a list of regions, with the goal of supporting multiregion serverless when it does roll out. I'll change it back to a list, but leave it with only one value.

erademacher (commented):
Looks like I missed an update to the network acceptance test setup. I'll fix that tomorrow. Then, I'd be happy to go over this PR with anyone who wants a rundown.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-:-NEhbSUMAsUmyOrKUyeZ:b-x0pbsh)* status: 1 of 22 files reviewed, 5 unresolved discussions (waiting on @fantapop, @jenngeorge, @marksoper, @prafull01, and @sean-)

fantapop:
*[`main.go` line 55 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEbxs7R9TnNc9ajmwgR:-NEiyndrCHIpkFdn-AbH:bjq0pcx) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/main.go#L55)):*

That's very weird, because unless I remove the "/providers", I get this:
```
unable to validate ServeOpts: unable to validate Address: expected hostname/namespace/type format, got: registry.terraform.io/providers/cockroachdb/cockroach
```
</blockquote></details>

I agree its very weird. It seems like maybe its not intended to be a routable address.  I've looked at a few other examples and none of them resolve.  Including the aws provider, vercel and planetscale.

Here is the description of that field.  It seems to indicate that its used only for matching like an ID

https://github.com/hashicorp/terraform-plugin-sdk/blob/main/plugin/serve.go#L83-L88

fantapop:
*[`examples/workflows/cockroach_serverless_cluster/main.tf` line 56 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEc-bSD2-b6F8Cdpy6_:-NEizZJL0Mez4bl9HVhW:bigwnhb) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/examples/workflows/cockroach_serverless_cluster/main.tf#L56)):*

>It's Serverless so only a single region is supported.

That was my thinking, but the API spec does take a list of regions, with the goal of supporting multiregion serverless when it does roll out. I'll change it back to a list, but leave it with only one value.
</blockquote></details>

sounds good. FWIW, I think its nice to provide this example using a list even though we only support one currently.  This is probably the code that someone wants to use in their own templates since that field is a list.

fantapop (commented):
I'd be interested in a walk through if you find the time.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-:-NEj-Is2CUbmwWbPf0mA:b-nfon2a)* status: 1 of 22 files reviewed, 3 unresolved discussions (waiting on @erademacher, @jenngeorge, @marksoper, @prafull01, and @sean-)

erademacher (commented):
Cool, we can walk through it at watercooler today.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-:-NEkyCd-Ag9X8EvtvDLs:b-3mz8c6)* status: 1 of 22 files reviewed, 1 unresolved discussion (waiting on @jenngeorge, @marksoper, @prafull01, and @sean-)

___
*[`main.go` line 55 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEbxs7R9TnNc9ajmwgR:-NEl1OTl6jacRx26U8h-:b4gm716) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/main.go#L55)):*
<details><summary><i>Previously, fantapop (Christopher Fitzner) wrote…</i></summary><blockquote>

I agree its very weird. It seems like maybe its not intended to be a routable address.  I've looked at a few other examples and none of them resolve.  Including the aws provider, vercel and planetscale.

Here is the description of that field.  It seems to indicate that its used only for matching like an ID

https://github.com/hashicorp/terraform-plugin-sdk/blob/main/plugin/serve.go#L83-L88
</blockquote></details>

Thanks for looking that up. I'm going to leave it as-is then. I guess a URN doesn't have to be a URL.

jenngeorge (commented):
Testing the examples with the updated provider worked for me! 
Used `terraform apply` to create clusters "jenn-tf-dedicated" and "jenn-tf-serverless", update the spend limit for "jenn-tf-serverless", and add regions to "jenn-tf-dedicated".

> Reviewed 6 of 19 files at r1, 3 of 12 files at r2, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-:-NEnga90ALlM3NJ5KbFo:bx1s36e)* status: 11 of 22 files reviewed, 4 unresolved discussions (waiting on @erademacher, @fantapop, @marksoper, @prafull01, and @sean-)

___
*[`Makefile` line 3 at r5](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEnZ4R6-ona_YbeAj4A:-NEnZ4R6-ona_YbeAj4B:bdy98k8) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/23e3a1c20526f0424a5ae4356a35e69d786b858f/Makefile#L3)):*
> ```Makefile
> TEST?=$$(go list ./... | grep -v 'vendor')
> HOSTNAME=registry.terraform.io
> NAMESPACE=hashicorp
> ```

Is the namespace for provider address "registry.terraform.io/cockroachdb/cockroach" "cockroachdb" instead of "hashicorp"? 

I have a `.terraformrc` file with `plugin_cache_dir = "$HOME/.terraform.d/plugins"` and `terraform init` did not find the new binary from `make install` until I changed `NAMESPACE=cockroachdb` here.
___
*[`internal/provider/cluster_resource_test.go` line 58 at r5](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEnrA_E8lEOBl4mlu7L:-NEnrA_E8lEOBl4mlu7M:bww36sh) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/23e3a1c20526f0424a5ae4356a35e69d786b858f/internal/provider/cluster_resource_test.go#L58)):*
> ```Go
> 	t.Parallel()
> 	var (
> 		clusterName  = fmt.Sprintf("crdb-dedicated-%s", GenerateRandomString(4))
> ```

nit: I cancelled running tests, then saw emails that I had created some "crdb-testname-xxxx" clusters and deleted them manually. What do you think about prefixing test cluster names with something like "tftest-" so they're easy to find if abandoned?
___
*[`internal/provider/cockroach_cluster_data_source.go` line 45 at r5](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEnqbQ-E2T80TL257Ou:-NEnqbQ-E2T80TL257Ov:b9f7et0) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/23e3a1c20526f0424a5ae4356a35e69d786b858f/internal/provider/cockroach_cluster_data_source.go#L45)):*
> ```Go
> 			"cockroach_version": {
> 				Type:     types.StringType,
> 				Computed: true,
> ```

Did removing `tfsdk.UseStateForUnknown()` result in the all the attribute values reading `(known after apply)` in the plan? It's bit long and looks like more changes than will happen. Why not use the state values for most of these attributes?
```
  # data.cockroach_cluster.cockroach will be read during apply
  # (depends on a resource or a module with changes pending)
 <= data "cockroach_cluster" "cockroach" {
      + account_id        = (known after apply)
      + cloud_provider    = (known after apply)
      + cockroach_version = (known after apply)
      + creator_id        = (known after apply)
      + dedicated         = {
          + disk_iops        = (known after apply)
          + machine_type     = (known after apply)
          + memory_gib       = (known after apply)
          + num_virtual_cpus = (known after apply)
          + storage_gib      = (known after apply)
        }
      + id                = "[redacted]"
      + name              = (known after apply)
      + operation_status  = (known after apply)
      + plan              = (known after apply)
      + regions           = [
        ] -> (known after apply)
      + serverless        = {
          + routing_id  = (known after apply)
          + spend_limit = (known after apply)
        }
      + state             = (known after apply)
    }
```

jenngeorge (commented):
I posted too soon -- adding regions succeeded for my "jenn-tf-dedicated" cluster, but the terraform output shows these provider errors:
```
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to cockroach_cluster.cockroach, provider
│ "provider[\"registry.terraform.io/cockroachdb/cockroach\"]" produced an unexpected new value: .regions[1].name:
│ was cty.StringVal("us-west2"), but now cty.StringVal("us-east4").
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to cockroach_cluster.cockroach, provider
│ "provider[\"registry.terraform.io/cockroachdb/cockroach\"]" produced an unexpected new value: .regions[2].name:
│ was cty.StringVal("us-east4"), but now cty.StringVal("us-west2").
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```
Here is the relevant part of the plan (From the default single "us-east1" region, I set `cloud_provider_region = ["us-east1", "us-west2", "us-east4"]`):
```
  # cockroach_cluster.cockroach will be updated in-place
  ~ resource "cockroach_cluster" "cockroach" {
        id                = "cb914bf8-ea58-4d91-83df-06d186c777fd"
        name              = "jenn-tf-dedicated"
      ~ regions           = [
          + {
              + name       = "us-west2"
              + node_count = 1
              + sql_dns    = (known after apply)
              + ui_dns     = (known after apply)
            },
          + {
              + name       = "us-east4"
              + node_count = 1
              + sql_dns    = (known after apply)
              + ui_dns     = (known after apply)
            },
            # (1 unchanged element hidden)
        ]
    }
```
Maybe we need to sort by region name somewhere?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-:-NEnwfp97cWFLkl4OXYZ:bkqqokx)* status: 11 of 22 files reviewed, 4 unresolved discussions (waiting on @erademacher, @fantapop, @marksoper, @prafull01, and @sean-)

marksoper:
>  Provider produced inconsistent result after apply
I can help look into this @jenngeorge - I've been looking for sort requirements suggested for Terraform config arrays. I don't think that order-is-important matching could work because users wouldn't know how to pre-order their arrays in their configs not knowing what the sort function is in the provider.

marksoper:
*[`Makefile` line 3 at r5](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEnZ4R6-ona_YbeAj4A:-NEqqnL067aqhU2jySrL:bzadb0s) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/23e3a1c20526f0424a5ae4356a35e69d786b858f/Makefile#L3)):*
<details><summary><i>Previously, jenngeorge (Jennifer Georgevich) wrote…</i></summary><blockquote>

Is the namespace for provider address "registry.terraform.io/cockroachdb/cockroach" "cockroachdb" instead of "hashicorp"? 

I have a `.terraformrc` file with `plugin_cache_dir = "$HOME/.terraform.d/plugins"` and `terraform init` did not find the new binary from `make install` until I changed `NAMESPACE=cockroachdb` here.
</blockquote></details>

If I recall this is configuration for the destination of the local dev build, doesn't apply to the registered provider. But the naming of the NAMESPACE in the Makefile may need to correspond to the actual registry URL  - i.e. the dir structure locally has to have the same path as the registry for terraform to find it locally. I changed the registry setting in order to make registration work, but never changed the NAMESPACE value. So this would explain why it didn't work for you @jenngeorge and why your change fixed it.

marksoper:
*[`main.go` line 55 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEbxs7R9TnNc9ajmwgR:-NEqsDVS1fhoO8QR_-6A:b-qc9wah) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/main.go#L55)):*

Thanks for looking that up. I'm going to leave it as-is then. I guess a URN doesn't have to be a URL.
</blockquote></details>

Agreed that the correct value should be "registry.terraform.io/cockroachdb/cockroach" (without the /providers/). This is an "id" not a url. 
Btw our currently published provider has a totally incorrect value. The value in the main branch is a fix I applied only 9 days ago, and we have re-published since then.
https://github.com/cockroachdb/terraform-provider-cockroach/pull/31/files

marksoper:
*[`docs/resources/cluster.md` line 46 at r5](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEukalDAYe2eo4dmJaA:-NEukalDAYe2eo4dmJaB:bqd26k) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/23e3a1c20526f0424a5ae4356a35e69d786b858f/docs/resources/cluster.md#L46)):*
> ```Markdown
> - `name` (String)
> - `node_count` (Number)
> - `sql_dns` (String)
> ```

should sql_dns and ui_dns be "Computed" instead of "Required"?

marksoper:
*[`examples/workflows/cockroach_serverless_cluster/main.tf` line 56 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEc-bSD2-b6F8Cdpy6_:-NEupiqV6Tz_hqjqI-pe:b-ohqtgo) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/53d6c2aff3c14bba2dff905ddf9b01a89e878c72/examples/workflows/cockroach_serverless_cluster/main.tf#L56)):*
<details><summary><i>Previously, fantapop (Christopher Fitzner) wrote…</i></summary><blockquote>

sounds good. FWIW, I think its nice to provide this example using a list even though we only support one currently.  This is probably the code that someone wants to use in their own templates since that field is a list. 
</blockquote></details>

good point @fantapop I agree

marksoper:
*[`internal/provider/cluster_resource.go` line 108 at r5](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEuqXJv4H1thGgdSPBR:-NEuqXJv4H1thGgdSPBS:bh95b8a) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/23e3a1c20526f0424a5ae4356a35e69d786b858f/internal/provider/cluster_resource.go#L108)):*
> ```Go
> 						},
> 					},
> 					"memory_gib": {
> ```

@erademacher should this also have the "Optional: true" line?

marksoper:
*[`internal/provider/cockroach_cluster_data_source.go` line 45 at r5](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEnqbQ-E2T80TL257Ou:-NEuvEajCS0n4HKg0tAp:b-iow8ul) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/23e3a1c20526f0424a5ae4356a35e69d786b858f/internal/provider/cockroach_cluster_data_source.go#L45)):*
<details><summary><i>Previously, jenngeorge (Jennifer Georgevich) wrote…</i></summary><blockquote>

Did removing `tfsdk.UseStateForUnknown()` result in the all the attribute values reading `(known after apply)` in the plan? It's bit long and looks like more changes than will happen. Why not use the state values for most of these attributes?
```
  # data.cockroach_cluster.cockroach will be read during apply
  # (depends on a resource or a module with changes pending)
 <= data "cockroach_cluster" "cockroach" {
      + account_id        = (known after apply)
      + cloud_provider    = (known after apply)
      + cockroach_version = (known after apply)
      + creator_id        = (known after apply)
      + dedicated         = {
          + disk_iops        = (known after apply)
          + machine_type     = (known after apply)
          + memory_gib       = (known after apply)
          + num_virtual_cpus = (known after apply)
          + storage_gib      = (known after apply)
        }
      + id                = "[redacted]"
      + name              = (known after apply)
      + operation_status  = (known after apply)
      + plan              = (known after apply)
      + regions           = [
        ] -> (known after apply)
      + serverless        = {
          + routing_id  = (known after apply)
          + spend_limit = (known after apply)
        }
      + state             = (known after apply)
    }
```
</blockquote></details>

In the "data source", as opposed to the "resource", everything except the id is (known after apply) by definition because data sources are strictly read-only representations of the resource.

marksoper (approved):
Approve w/ several suggestions/questions ...

erademacher (commented):
>Maybe we need to sort by region name somewhere?

Good catch. I added a sort by region name according to the plan, along with some tests.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-:-NEwJNYo-xZCwCW8PyjN:b-2buxjn)* status: 9 of 24 files reviewed, 7 unresolved discussions (waiting on @fantapop, @jenngeorge, @jlinder, @marksoper, @prafull01, and @sean-)

*[a discussion](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEqw-xl96-YmXbxLe6c:-NEwJc_0DFp31H99i8WR:bod0ehy) (no related file):*
<details><summary><i>Previously, marksoper (Mark Soper) wrote…</i></summary><blockquote>

From @jlinder on how to publish a new version of the provider:

If you push a new tag for the version number you want to publish, it will automatically build and publish the code at the SHA that was tagged. So if you push the tag v0.2.0 or v0.1.1, a GitHub Actions workflow will automatically start and do the publishing. You’ll be able to watch the logs from the running workflow on this page.

https://github.com/cockroachdb/terraform-provider-cockroach/actions/workflows/release.yml

Cc: @erademacher @jenngeorge @fantapop 
</blockquote></details>

Noted. Thanks.

___
___
*[`Makefile` line 3 at r5](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEnZ4R6-ona_YbeAj4A:-NEwJfit2zqUCkbSrusO:bt2vr1b) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/23e3a1c20526f0424a5ae4356a35e69d786b858f/Makefile#L3)):*
<details><summary><i>Previously, marksoper (Mark Soper) wrote…</i></summary><blockquote>

If I recall this is configuration for the destination of the local dev build, doesn't apply to the registered provider. But the naming of the NAMESPACE in the Makefile may need to correspond to the actual registry URL  - i.e. the dir structure locally has to have the same path as the registry for terraform to find it locally. I changed the registry setting in order to make registration work, but never changed the NAMESPACE value. So this would explain why it didn't work for you @jenngeorge and why your change fixed it.
</blockquote></details>

Fixed.
___
*[`docs/resources/cluster.md` line 46 at r5](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEukalDAYe2eo4dmJaA:-NEwJlho0TGxmwLCMeiv:bfafxdk) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/23e3a1c20526f0424a5ae4356a35e69d786b858f/docs/resources/cluster.md#L46)):*
<details><summary><i>Previously, marksoper (Mark Soper) wrote…</i></summary><blockquote>

should sql_dns and ui_dns be "Computed" instead of "Required"?
</blockquote></details>

They are computed. The docs don't seem to reflect that, though, even after I regenerate them. I'll investigate.
___
*[`internal/provider/cluster_resource.go` line 108 at r5](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEuqXJv4H1thGgdSPBR:-NEwIuJm27Y5yMDyHLaL:b-w7sz0r) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/23e3a1c20526f0424a5ae4356a35e69d786b858f/internal/provider/cluster_resource.go#L108)):*
<details><summary><i>Previously, marksoper (Mark Soper) wrote…</i></summary><blockquote>

@erademacher should this also have the "Optional: true" line?
</blockquote></details>

No, because memory is dependent on machine type. You don't get to arbitrarily choose your RAM.
___
*[`internal/provider/cluster_resource_test.go` line 58 at r5](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEnrA_E8lEOBl4mlu7L:-NEwJ-rjB2LMjYZfnZ53:b-luduzo) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/23e3a1c20526f0424a5ae4356a35e69d786b858f/internal/provider/cluster_resource_test.go#L58)):*
<details><summary><i>Previously, jenngeorge (Jennifer Georgevich) wrote…</i></summary><blockquote>

nit: I cancelled running tests, then saw emails that I had created some "crdb-testname-xxxx" clusters and deleted them manually. What do you think about prefixing test cluster names with something like "tftest-" so they're easy to find if abandoned?
</blockquote></details>

Updated. I had to shorten the random portion of some of them to avoid the 20-character limit. Apparently, that limit has been lifted in console. We should update the API too. I'll put something in JIRA.
___
*[`internal/provider/cockroach_cluster_data_source.go` line 45 at r5](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEnqbQ-E2T80TL257Ou:-NEwJAZx6gQfVLH51Qg2:b-mc876x) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/23e3a1c20526f0424a5ae4356a35e69d786b858f/internal/provider/cockroach_cluster_data_source.go#L45)):*
<details><summary><i>Previously, marksoper (Mark Soper) wrote…</i></summary><blockquote>

In the "data source", as opposed to the "resource", everything except the id is (known after apply) by definition because data sources are strictly read-only representations of the resource.
</blockquote></details>

I added `UseStateForUnknown` for the immutable fields: cloud_provider, creator_id, and name.

erademacher (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38)* status: 9 of 24 files reviewed, 7 unresolved discussions (waiting on @fantapop, @jenngeorge, @jlinder, @marksoper, @prafull01, and @sean-)

___
*[`docs/resources/cluster.md` line 46 at r5](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-NEukalDAYe2eo4dmJaA:-NEwLSGh6GBd4VSId975:bchhcrv) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/23e3a1c20526f0424a5ae4356a35e69d786b858f/docs/resources/cluster.md#L46)):*

They are computed. The docs don't seem to reflect that, though, even after I regenerate them. I'll investigate.
</blockquote></details>

Looks like it was a bug in the doc generator. I updated it, reran it, and it looks fixed now. Thanks for catching that!

erademacher (commented):
Sorry, I'm getting sloppy with my commit structure. Probably still best if you review it squashed.

Today's changes:
- Sorted regions according to plan
- Made some adjustments to the schema to add RequiresReplace and UseStateForUnknown where applicable
- Added an undocumented COCKROACH_SERVER variable so we can test on staging (and avoid triggering weird alerts in prod)
- Fixed the namespace used for local provider resolution
- Changed the names of the clusters created by acceptance tests
- Updated serverless tests to use regions available in staging
- Regenerated docs after updating the tool

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-:-NEwUBCh5GjfO96yZc78:b-i713wo)* status: 9 of 26 files reviewed, 7 unresolved discussions (waiting on @fantapop, @jenngeorge, @jlinder, @marksoper, @prafull01, and @sean-)

erademacher:
Hey @jlinder, I added an env variable to point the tests to a different server. I'd like to point CI at staging, but I'm not sure where to manage secrets (i.e. the API key). Is that something you could help with? Or is that something @marksoper manages?

jlinder:
Hi @erademacher, Up til now it's probably been something Mark has managed. 

Is this a change to https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/.github/workflows/ci.yaml? If so, do you want CI to always run against staging or will it run against two different servers for every run? If it's always against the staging env and the code is using the `COCKROACH_API_KEY` secret (like in [this line](https://github.com/cockroachdb/terraform-provider-cockroach/blob/362bcc4031e3975403aec397f257551c7d26851b/.github/workflows/ci.yaml#L46)), then the value of the GitHub Actions secret can be updated with the new value. And if you need another secret added, we another one can be added to the GitHub Actions secrets for the repository.

erademacher (commented):
I'm gonna take silence as complicity on this one. If anyone has additional review feedback, I can address it in a separate PR.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-:-NFK6AfJ4cZPNH4Fqiqd:b7xxdkp)* status: 9 of 26 files reviewed, 7 unresolved discussions (waiting on @fantapop, @jenngeorge, @jlinder, @marksoper, @prafull01, and @sean-)

jenngeorge (commented):
:lgtm: sorry for late reply!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/38#-:-NFKF9lmAKHiAVu-gv1e:b3q30qw)* status: 9 of 26 files reviewed, 7 unresolved discussions

examples/resources/cockroach_serverless_cluster/cockroach_serverless_cluster.tf
```diff
@@ -1,10 +1,14 @@
...
-            regions = ["us-east1"]
-            spend_limit = 1
-        }
+  name           = "cockroach-serverless"
+  cloud_provider = "GCP"
+  create_spec = {
+    serverless = {
+      spend_limit = 1
     }
+  }
```
prafull01, line 8: This example is incorrect, according to new schema it should be 
```
serverless = {
       spend_limit = 1
}
```

internal/provider/cluster_resource.go
```diff
@@ -72,194 +76,110 @@ func (r clusterResourceType) GetSchema(_ context.Context) (tfsdk.Schema, diag.Di
...
+				Required: true,
 				PlanModifiers: tfsdk.AttributePlanModifiers{
 					tfsdk.UseStateForUnknown(),
 				},
 				Attributes: tfsdk.ListNestedAttributes(map[string]tfsdk.Attribute{
 					"name": {
-						Computed: true,
+						Required: true,
 						Type:     types.StringType,
 					},
```
prafull01, line 161: Earlier we had `create_spec` and `update_spec` differently. `update_spec` has the only fields which are allowed to be updated in cluster. However now we only have `serverless` and `dedicated` field, hence we have to add the validation which are not allowed to be updated. For example: the `region` is not allowed to be updated. Our TF provider should properly throw error that you are not allowed to be updated.  

TF provider is crashing when we try to update the region:
```
cockroach_cluster.cockroach: Modifying... [id=d161cece-663a-448e-964e-c4217db6fe9c]
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to cockroach_cluster.cockroach, provider "provider[\"registry.terraform.io/cockroachdb/cockroach\"]" produced an unexpected new
│ value: .regions[0].name: was cty.StringVal("us-east2"), but now cty.StringVal("us-east1").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```

prafull01:
@erademacher Sorry for the late review. I was on leave. Added few comments, please take a look.

internal/provider/cluster_resource.go
```diff
@@ -72,194 +76,110 @@ func (r clusterResourceType) GetSchema(_ context.Context) (tfsdk.Schema, diag.Di
...
+							tfsdk.UseStateForUnknown(),
+						},
+					},
+					"machine_type": {
+						Type:     types.StringType,
+						Optional: true,
 						Computed: true,
 						PlanModifiers: tfsdk.AttributePlanModifiers{
 							tfsdk.UseStateForUnknown(),
 						},
```
prafull01, line 140: When I tried to update the `machine_type` in dedicated cluster the TF provided gave inconsistent result. I have updated the `machine_type     = "n1-standard-4" -> "n1-standard-8"`

```
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to cockroach_cluster.cockroach, provider "provider[\"registry.terraform.io/cockroachdb/cockroach\"]" produced an unexpected new
│ value: .dedicated.memory_gib: was cty.NumberIntVal(15), but now cty.NumberIntVal(30).
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to cockroach_cluster.cockroach, provider "provider[\"registry.terraform.io/cockroachdb/cockroach\"]" produced an unexpected new
│ value: .dedicated.num_virtual_cpus: was cty.NumberIntVal(4), but now cty.NumberIntVal(8).
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```

PR #39, opened 2022-10-27
Clean up SQL User and Allowlist resources

erademacher:
- Fixed import by composing a unique ID for both resources
- Renamed the old id to cluster_id
- Implemented Read with an API call instead of copying the state
- Cleaned up error message formatting
- Updated test provider to use COCKROACH_SERVER
- Consolidated SQL user models
- Reran `go generate`
- Pull misconfigured provider error into util function

### DIFF
--- a/internal/provider/networking_resource.go
+++ b/internal/provider/networking_resource.go
@@ -75,10 +75,7 @@ type allowListResource struct {
 
 func (n allowListResource) Create(ctx context.Context, req tfsdk.CreateResourceRequest, resp *tfsdk.CreateResourceResponse) {
 	if !n.provider.configured {
-		resp.Diagnostics.AddError(
-			"Provider not configured",
-			"The provider hasn't been configured before apply, likely because it depends on an unknown value from another resource. This leads to weird stuff happening, so we'd prefer if you didn't do that. Thanks!",
-		)
+		addConfigureProviderErr(&resp.Diagnostics)
 		return
 	}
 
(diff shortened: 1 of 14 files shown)


jenngeorge (approved):
:lgtm_strong: 

> Reviewed 14 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/39#-:-NFQsuOHDHJ3htoA2Qj9:b-uwkowd)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @ryanluu12345)

internal/provider/networking_resource.go
```diff
@@ -75,59 +89,61 @@ type allowListResource struct {
...
-	var plan AllowlistEntry
-	diags := req.Config.Get(ctx, &plan)
+	var entry AllowlistEntry
+	diags := req.Config.Get(ctx, &entry)
 	resp.Diagnostics.Append(diags...)
+	// Create a unique ID (required by terraform framework) by combining
+	// the cluster ID and full CIDR address.
+	entry.ID = types.String{
+		Value: fmt.Sprintf(allowListIDFmt, entry.ClusterId.Value, entry.CidrIp.Value, entry.CidrMask.Value),
+	}
```
prafull01, line 103: We are creating this unique id which can be non-unique in certain cases. Consider a scenario:
1. We have created an IPAllowlist with `192.168.1.1/32` and cluster id: `cdb0012d-1ce2-4341-a0dd-e8a649affc58`. The unique id will be `cdb0012d-1ce2-4341-a0dd-e8a649affc58:192.168.1.1/32`
2. Now lets say we have updated the CIDR IP to `192.168.1.2/32`, then the unique id will not change but the IP address will change.
3. Now we try to add the prev CIDR address again with new name `192.168.1.1/32`, then the unique id for this new IPAllowlist resource and prev resource will be same

erademacher: Unique doesn't mean immutable. It's computed and will change when the address changes. The Terraform ID (cockroach_cluster.cluster_name) is unique and immutable. The provider ID (cdb0012d-1ce2-4341-a0dd-e8a649affc58:192.168.1.1/32) can change, and that's OK. It's only actually used as an identifier during acceptance testing and `terraform import` anyway.

erademacher:
TFTR!

internal/provider/networking_resource.go
erademacher: Sorry, I misspoke. It won't change when the address changes because it isn't handled in Update. I'll fix in a followup PR.

PR #42, opened 2022-10-28
Mock tests

erademacher:
Split out the existing resource tests into acceptance tests and
integration tests. They do the same thing, but the integration
tests use gomock and don't take 45m and an extra AWS bill to run.

I configured the CI workflow to use integration tests, though we
can talk about adding the acceptance tests back as a prerelease
check. I'd need to read through some docs to figure out how to do
that.

Bumped Go to 1.18 to make use of generics for HookGlobal.


erademacher:
These tests are failing after dropping my other two in-review commits. I'm going to rebase again to fix it.

erademacher:
I'm not sure how to force the github workflow to use Go 1.18. It's set in go.mod and build.yaml. Unless you guys know off the top of your head or I can figure it out quickly, I'll just revert to a HookGlobal without generics.

prafull01 (approved):
Overall looks good. However, we have to write acceptance tests which run periodically and on pre-release. These acceptance tests should also include the update of the resources where the most of the issues in TF provider gets introduced while changing the TF state. We can handle those in separate PR.
:lgtm:

> Reviewed 3 of 3 files at r1, 10 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/42#-:-NFY2JrG-sYUTN1q0TSc:bbxvoa7)* status: 13 of 15 files reviewed, all discussions resolved (waiting on @jenngeorge and @ryanluu12345)

erademacher:
TFTR!

> However, we have to write acceptance tests which run periodically and on pre-release. These acceptance tests should also include the update of the resources where the most of the issues in TF provider gets introduced while changing the TF state. We can handle those in separate PR.

Totally agree. I'm going to add update tests next, and I'll look into running full acceptance tests nightly and during prerelease.

PR #44, opened 2022-11-01
Private endpoint connection resource

erademacher:
- Added new private_endpoint_connection resource
- Added tests
- Consolidated some reads for resources where we need to poll
  for create/update status

Acceptance tests are currently skipped. Added a comment elaborating
why.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44)


erademacher:
This is based on the existing EndpointServices PR, so feel free to only look at the last commit.

I won't merge this until I'm able to test manually. Still waiting on shared AWS credentials to do that.

prafull01 (commented):
> Reviewed 17 of 17 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44)* status: all files reviewed, 3 unresolved discussions (waiting on @erademacher, @marksoper, and @ryanluu12345)

___
*[`internal/provider/private_endpoint_connection_resource.go` line 126 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44#-NFtL6Ri-NkbJHKVPJih:-NFtL6Ri-NkbJHKVPJii:bc4a846) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/366208d139a31f2899fce569f87d34611233a0d0/internal/provider/private_endpoint_connection_resource.go#L126)):*
> ```Go
> 		return
> 	}
> ```

Same check if the cluster type is not AWS, the return error from here
___
*[`internal/provider/private_endpoint_services_resource.go` line 139 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44#-NFtMsrLC2dT-23meUx5:-NFtMsrLC2dT-23meUx6:b-h5gmn7) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/366208d139a31f2899fce569f87d34611233a0d0/internal/provider/private_endpoint_services_resource.go#L139)):*
> ```Go
> 		return
> 	}
> ```

Only for AWS type cluster
___
*[`internal/provider/private_endpoint_services_resource.go` line 182 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44#-NFtN7IREua2tt-ODRzH:-NFtN7IREua2tt-ODRzI:bu6afw7) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/366208d139a31f2899fce569f87d34611233a0d0/internal/provider/private_endpoint_services_resource.go#L182)):*
> ```Go
> 	apiResp, _, err := n.provider.service.ListPrivateEndpointServices(ctx, state.ClusterID.Value)
> 	if err != nil {
> 		resp.Diagnostics.AddError("Couldn't retrieve endpoint services",
> ```

Instead of returning an error in case of PrivateEndpointServices not found. We should print the error and remove the PrivateEndpointServices resource from the terraform state as it no longer present

ryanluu12345 (approved):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44)* status: all files reviewed, 4 unresolved discussions (waiting on @erademacher and @marksoper)

___
*[`internal/provider/private_endpoint_connection_resource.go` line 238 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44#-NFtuE535XmMx7t9fHI2:-NFtuE535XmMx7t9fHI3:b-6wqb99) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/2a8b883f296934d60842e6d797a8a01bbd6bba27/internal/provider/private_endpoint_connection_resource.go#L238)):*
> ```Go
> }
> 
> func waitForEndpointConnectionCreatedFunc(ctx context.Context, clusterID, endpointID string, cl client.Service, connection *client.AwsEndpointConnection) resource.RetryFunc {
> ```

I noticed this pattern in each of the `provider/*resource` files. I am wondering if you've considered making a generic `wait` function that can take in a `client.List` method. Hmm I guess there may be challenges since the conditions you check for each of these resources is different.

ryanluu12345 (commented):
:lgtm: except for @prafull01 's comments about restricting this to just AWS dedicated clusters.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44#-:-NFtv5wy1jBvbQCCnGlX:b-lmbia1)* status: all files reviewed, 4 unresolved discussions (waiting on @erademacher, @marksoper, and @prafull01)

___
*[`internal/provider/private_endpoint_services_resource.go` line 182 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44#-NFtN7IREua2tt-ODRzH:-NFtuuT53t_24ygwaUTm:bm8kwe5) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/366208d139a31f2899fce569f87d34611233a0d0/internal/provider/private_endpoint_services_resource.go#L182)):*
<details><summary><i>Previously, prafull01 (Prafull Ladha) wrote…</i></summary><blockquote>

Instead of returning an error in case of PrivateEndpointServices not found. We should print the error and remove the PrivateEndpointServices resource from the terraform state as it no longer present
</blockquote></details>

I think if the `PrivateEndpointServices` are not found, it won't error. If you mean that "not found" means there are no resources. In that case, I'd expect there to be no services in the Terraform state. However, I think this error relates to the listing from the API failing, which would semantically be an error.

erademacher (commented):
TFTRs! I was able to get this all working in conjunction with the AWS provider. I added it as a workflow example. If you wouldn't mind taking a quick look at that, I'd appreciate it.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44#-:-NFuUIub01FRpptCcocu:b-bd9g8u)* status: 16 of 19 files reviewed, 4 unresolved discussions (waiting on @marksoper, @prafull01, and @ryanluu12345)

___
*[`internal/provider/private_endpoint_connection_resource.go` line 126 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44#-NFtL6Ri-NkbJHKVPJih:-NFuUbfO-A7YM9ycxeJG:b-896fix) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/366208d139a31f2899fce569f87d34611233a0d0/internal/provider/private_endpoint_connection_resource.go#L126)):*
<details><summary><i>Previously, prafull01 (Prafull Ladha) wrote…</i></summary><blockquote>

Same check if the cluster type is not AWS, the return error from here
</blockquote></details>

Done.
___
*[`internal/provider/private_endpoint_connection_resource.go` line 238 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44#-NFtuE535XmMx7t9fHI2:-NFuUfvK6Un_mOj85vk4:b-o64tgl) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/2a8b883f296934d60842e6d797a8a01bbd6bba27/internal/provider/private_endpoint_connection_resource.go#L238)):*
<details><summary><i>Previously, ryanluu12345 wrote…</i></summary><blockquote>

I noticed this pattern in each of the `provider/*resource` files. I am wondering if you've considered making a generic `wait` function that can take in a `client.List` method. Hmm I guess there may be challenges since the conditions you check for each of these resources is different.
</blockquote></details>

Yeah, they're different enough that I don't think it's worth trying to consolidate into one function.
___
*[`internal/provider/private_endpoint_services_resource.go` line 139 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44#-NFtMsrLC2dT-23meUx5:-NFuUtbZ0LP3J4tIhW8G:bg9hjbi) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/366208d139a31f2899fce569f87d34611233a0d0/internal/provider/private_endpoint_services_resource.go#L139)):*
<details><summary><i>Previously, prafull01 (Prafull Ladha) wrote…</i></summary><blockquote>

Only for AWS type cluster
</blockquote></details>

I added an AWS check here before merging that last PR :)

erademacher (commented):
Well, it looks like I actually need to configure separate VPCs for each region. I'll update the example, but I can't get to it until tomorrow.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44#-:-NFucXPO1J28GPoBJrIm:b-ofwdd1)* status: 16 of 19 files reviewed, 4 unresolved discussions (waiting on @marksoper, @prafull01, and @ryanluu12345)

erademacher (commented):
I changed the example to single-region. It's possible to create connections in multiple regions, but it requires aliasing multiple AWS providers and it's just more complicated than it's worth.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44#-:-NFuzYrh4_9WP1oKh1NU:bcnt1jy)* status: 16 of 19 files reviewed, 4 unresolved discussions (waiting on @marksoper, @prafull01, and @ryanluu12345)

prafull01 (approved):
> Reviewed 2 of 3 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44)* status: all files reviewed, 2 unresolved discussions (waiting on @marksoper and @ryanluu12345)

erademacher (commented):
Thanks Prafull!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/44#-:-NFy3sPi174K3z44naLE:byjmupx)* status: all files reviewed, 2 unresolved discussions (waiting on @marksoper, @prafull01, and @ryanluu12345)

PR #45, opened 2022-11-03
Fixed endpoint connection import and annotated example

erademacher:
Just a quick change to support `import` for connections, even though it probably won't be used often. Follows the same pattern as SQL users. Tested manually, and tested import for services too. I also added a warning comment to the example, since region is configured at the provider level for AWS.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/45)

### DIFF
--- a/docs/resources/private_endpoint_connection.md
+++ b/docs/resources/private_endpoint_connection.md
@@ -23,7 +23,7 @@ AWS PrivateLink Endpoint Connection
 ### Read-Only
 
 - `cloud_provider` (String)
-- `id` (String) Required by Terraform. Will always equal endpoint_id.
+- `id` (String) Used with `terrform import`. Format is "<cluster ID>:<endpoint ID>"
 - `region_name` (String)
 - `service_id` (String)
 
--- a/examples/workflows/aws_privatelink/main.tf
+++ b/examples/workflows/aws_privatelink/main.tf
@@ -70,8 +70,13 @@ provider "cockroach" {
 }
 
 provider "aws" {
-  # see https://registry.terraform.io/providers/hashicorp/aws/latest/docs
-  # for configuration steps
+  # See https://registry.terraform.io/providers/hashicorp/aws/latest/docs
+  # for configuration steps.
+
+  # Please don't use a variable for region in production! The AWS provider won't
+  # be able to find any resources if this value changes and you'll get
+  # into a weird state. Be sure to run `terraform destroy` before changing
+  # this value.
   region = var.aws_region
 }
 
@@ -83,7 +88,7 @@ resource "cockroach_cluster" "cockroach" {
     machine_type = var.machine_type
   }
   regions = [{
-    name = var.aws_region
+    name       = var.aws_region
     node_count = var.cluster_nodes
   }]
 }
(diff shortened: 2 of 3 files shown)


ryanluu12345 (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/45)* status: all files reviewed, 3 unresolved discussions (waiting on @erademacher, @marksoper, and @prafull01)

___
*[`examples/workflows/aws_privatelink/main.tf` line 152 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/45#-NFyOhm757ySNLVnCKsM:-NFyOhm757ySNLVnCKsN:b32a7qq) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/35c72f2678d860256299fd1390febc36488c79fe/examples/workflows/aws_privatelink/main.tf#L152)):*
> ```HCL
>   # the cockroach_private_endpoint_connection resource.
>   #
>   # private_dns_enabled = true
> ```

Is there a reason we call this `private_dns` and not `internal_dns`. The latter would match how it's currently referenced in console/intrusion?
___
*[`internal/provider/private_endpoint_connection_resource.go` line 34 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/45#-NFyP5Se4DaDXw_X-WNs:-NFyP5Se4DaDXw_X-WNt:bjj98g9) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/35c72f2678d860256299fd1390febc36488c79fe/internal/provider/private_endpoint_connection_resource.go#L34)):*
> ```Go
> type privateEndpointConnectionResourceType struct{}
> 
> // clusterID:endpointID
> ```

Did you end up verifying that endpoint IDs are globally unique? Not just unique to a service.
___
*[`internal/provider/private_endpoint_connection_resource.go` line 250 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/45#-NFyPDnoAKi3DLsgvxjs:-NFyPDnp3VDD1LWwCxUH:b-hvdew3) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/35c72f2678d860256299fd1390febc36488c79fe/internal/provider/private_endpoint_connection_resource.go#L250)):*
> ```Go
> 	// work, we need to deserialize an ID back into endpoint ID and cluster ID.
> 	matches := privateEndpointConnectionIDRegex.FindStringSubmatch(req.ID)
> 	if len(matches) != 3 {
> ```

Magic number. Why does it expect 3 to continue onward? Maybe refactor to a descriptive variable or comment on why 3.

ryanluu12345 (approved):
:lgtm: except the comment about the magic number for the number of matches.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/45#-:-NFyPZo03nNc_Q43FVnq:b-f6fbn)* status: all files reviewed, 3 unresolved discussions (waiting on @erademacher, @marksoper, and @prafull01)

prafull01 (approved):

PR #48, opened 2022-11-03
example naming updates

erademacher:


examples/workflows/aws_privatelink/main.tf
```diff
@@ -36,7 +40,7 @@ variable "storage_gib" {
 variable "machine_type" {
   type     = string
   nullable = false
-  default  = "n1-standard-2"
+  default  = "m5.large"
```
fantapop, line 43: is this the smallest we can suggest?  Running this will cost someone real money so we should try to be sensitive to that.  I noticed this banner in the terraform sensitive variables example. It might be nice if we could add a disclaimer somewhere.
Screen Shot 2022-11-03 at 3 48 58 PM

examples/workflows/aws_privatelink/main.tf
```diff
@@ -80,7 +84,7 @@ provider "aws" {
   region = var.aws_region
 }
 
-resource "cockroach_cluster" "cockroach" {
+resource "cockroach_cluster" "example" {
```
fantapop, line 87: "example" isn't bad.  The naming convention link I shared suggests naming it "this" if there is only one of a thing.  Other ideas for naming here are "dedicated"  or "serverless" for the cluster.  In real life for a sql_user I'd probably name it something like "admin" or "schema_owner" or "readonly" or "Evan" but example seems fine too.  If you don't feel inspired by any of these ideas I'm fine keeping it example everywhere.

examples/workflows/aws_privatelink/main.tf
```diff
@@ -9,6 +9,10 @@ variable "sql_user_name" {
   default  = "maxroach"
 }
 
+# Remember that even variables marked sensitive will show up
```
fantapop, line 12: These seem great. Thanks for adding

examples/workflows/aws_privatelink/main.tf
erademacher: That's our default 2-vcpu cluster and the smallest of the standard options. All dedicated clusters cost money, and there is no dedicated free tier, but I guess we could add something to the readme.

jenngeorge (approved):
LGTM!

examples/workflows/aws_privatelink/main.tf
erademacher: Added that disclaimer!

fantapop: great!

fantapop (approved):

erademacher:
TFTRs!

PR #49, opened 2022-11-03
Require spend limit

erademacher:
Users can still pass in a value of zero. It just can't be null, to avoid state errors. We might be able to make it optional again after implementing ModifyResource, but that's a larger change for another day.


jenngeorge (approved):
LGTM!

erademacher:
TFTR!

PR #51, opened 2022-11-04
Fix state updates on read for missing resources

erademacher:
Apparently, using an error will prevent any state changes. We need to use a warning instead. I also noticed that we're not correctly updating state in Read for allowlist entries and endpoint services, so fixed that.


fantapop (approved):

erademacher:
TFTR!

PR #53, opened 2022-11-04
Fix cluster state on initial creation

erademacher:
Previously, we attempted to copy over cluster state while polling for status, but accidentally overrode the local pointer address instead of the object itself. This change copies the cluster over. Added test cases for cluster and endpoint services, which behaves the same way.


jenngeorge (approved):
LGTM

marksoper (approved):

erademacher:
TFTR!

PR #57, opened 2022-11-09
Add link to Terraform registry

ianjevans:
Fixes CC-8334.


marksoper (commented):
@ianjevans I think I wrote this originally a while back when the repo was more oriented toward provider development, so could use a redesign.

- Users intending to install the provider and get started should find very clear instructions. The new link does that but it's not obvious from the associated `terraform-provider-cockroach` text. Could we make it more explicit.
- The "Terraform provider" link is informative but we don't really want people going there, maybe we take that out.
- The CC API links are pretty prominent - e.g. the "getting started" phrase might be confused with "getting started with the CC API Terraform provider"
- More info on Terraform could be made less prominent or removed.

I think I wrote this originally a while back when the repo was more oriented toward provider development, so could use a redesign.

erademacher (approved):

marksoper (approved):

marksoper:
@ianjevans looks good, feel free to merge, thank you

PR #58, opened 2022-11-10
Fix inconsistent result error when using cockroach_version

erademacher:
Transform version strings from API results into the format we expect for cluster creation.

Also moved the dedicated test cluster to GCP and us-central1 since it's a little bit cheaper and closer to our control plane.


jenngeorge (approved):
LGTM

erademacher:
TFTR!

PR #60, opened 2022-11-16
Work around race condition for endpoint connections

erademacher:
Since the CCAPI job to accept endpoint connections is async, we can
get into a state where the connection still shows as PENDING_ACCEPTANCE
after it's been accepted. The old wait logic looked for ACCEPTED or
PENDING, and failed out with other statuses, including
PENDING_ACCEPTANCE. This change interprets that status the same as
PENDING, and also lowers the timeout for connections (which usually
only take a minute or less) to 5 minutes.

### DIFF
--- a/internal/provider/private_endpoint_connection_resource.go
+++ b/internal/provider/private_endpoint_connection_resource.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
+	"time"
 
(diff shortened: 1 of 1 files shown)


ryanluu12345 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/60#-:-NH0zhX_1s4iqW2qP8Ff:bg7igkv)* status: 0 of 1 files reviewed, all discussions resolved

PR #61, opened 2022-11-16
Update timeouts and clean up constants

erademacher:
Cluster update operations can take a pretty long time, especially multiregion clusters. We had an SE bump into the timeout, even though it was still chugging along. Bumped it to 2 hours. Dropped the endpoint connection timeout to 10 minutes (usually completes in ~30s). Left cluster and endpoint service creation unchanged.

Generally refactored timeout constants, and also a couple unrelated cluster state constants that were hard coded and at risk of going out of sync with the API.

Also opportunistically fixed a fixed a segfault that happened if a call to GetCluster fails in the wait loop, trying to copy the value of a nil cluster pointer.

### DIFF
--- a/internal/provider/cluster_resource.go
+++ b/internal/provider/cluster_resource.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"regexp"
 	"sort"
+	"time"
 
 	"github.com/cockroachdb/cockroach-cloud-sdk-go/pkg/client"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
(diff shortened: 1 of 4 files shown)


ryanluu12345 (approved):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/61)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksoper)

PR #63, opened 2022-12-01
Update terraform-provider-framework to 0.17.0

erademacher:
This commit updates the terraform-provider-framework and supporting modules. Those updates include a ton of breaking changes, and thus required a significant amount of refactoring.

The refactoring is mostly straightforward, and there should be no functional changes. The biggest differences are around how resources and datasources are created and configured. Still, the diff should be easy enough to read.


ryanluu12345 (approved):
>What're the benefits of the newer version of the `terraform-provider-framework`? Also, what are the differences for how a >user would use this? Or do the examples still work exactly the same?

No functional changes. The examples don't need to change. This update fixes some stability issues, notably around config validation. This will address a lot of Jenn's bug bash feedback.

>Nit: why not `&plan.CockroachVersion.ValueString()`?

Because function return values aren't addressable.
https://utcc.utoronto.ca/~cks/space/blog/programming/GoAddressableValues

ryanluu12345 (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/63)* status: 2 of 13 files reviewed, 2 unresolved discussions (waiting on @erademacher, @jenngeorge, and @prafull01)

___
*[`internal/provider/networking_resource.go` line 97 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/63#-NIcRdV3CfVPXCMCWNqQ:-NIcRdV3CfVPXCMCWNqR:b-hvfv2g) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/34c5b18b182d560b0781af9fb0f9a0917100b195/internal/provider/networking_resource.go#L97)):*
> ```Go
> 
> func (r *allowListResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
> 	resp.TypeName = req.ProviderTypeName + "_allow_list"
> ```

Small Nit: constants for all of these suffixes for resources instead of hardcoded strings

erademacher:
TFTR!

Issue #65, opened 2022-12-17
Dedicated example throws an "unsupported version" error


ianjevans:
I got an "unsupported version" error when running `terraform apply` using the dedicated workflow example:
https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/examples/workflows/cockroach_dedicated_cluster/main.tf

```
cockroach_cluster.example: Creating...
╷
│ Error: Error creating cluster
│
│   with cockroach_cluster.example,
│   on main.tf line 89, in resource "cockroach_cluster" "example":
│   89: resource "cockroach_cluster" "example" {
│
│ Could not create cluster: invalid argument: unsupported version: v22.1.10
```

I used the following `terraform.tfvars` file:

~~~
cluster_name = "ian-dim-dog"
sql_user_name = "maxroach"
sql_user_password = "NotAGoodPassword"
cloud_provider = "GCP"
cloud_provider_regions = ["us-west2"]
cluster_node_count = 3
storage_gib = 15
machine_type = "n1-standard-2"
allow_list_name = "Ian's home network"
cidr_ip = "x.x.x.x"
cidr_mask = 32
~~~

The example `main.tf` specifies the default version as v22.1. Changing it to v22.2 allowed the plan to successfully apply.

marksoper:
I'm working on repro'ing this ...

marksoper:
<img width="967" alt="image" src="https://user-images.githubusercontent.com/193360/208987569-1460426c-4918-4f66-8ea1-eca740d3e038.png">

I got the above error message when using this main.tfvars file with the examples/workflows/cockroach_dedicated_cluster/main.tf file.

```
cluster_name = "soper-version-test"
cloud_provider = "AWS"
cloud_provider_regions = [
    "us-east-1",
    "us-east-2",
    "us-west-2"
]
machine_type = "m5.large"
```

erademacher:
I can't repro Ian's issue. When I use "v22.1", it sends the correct API request with that version string. I am seeing the same behavior as Mark, though, which suggests that the cluster might be auto-updating before marking itself as ready, or something like that. Digging in now.

erademacher:
Now it's working fine. Very strange.

ianjevans:
Closing as a "can't reproduce."

erademacher:
[...]

PR #66, opened 2022-12-17
Bump default CRDB dedicated version to v22.2

ianjevans:
Now that v22.2 is GA, bump the default version to v22.2 in the dedicated workflow example.


ianjevans:
This relates to #65 as a workaround, but users should be able to make v22.1 clusters the default value in their own configs.

erademacher (approved):

PR #67, opened 2022-12-20
[CC-8720]CMEK resource

erademacher:
This commit adds a new CMEK resource, which primarily manages the configuration of customer-managed encryption keys (CMEK). CMEK is enabled for a cluster when the resource is created, and keys can be rotated by updating their specs.

Due to circular dependencies (new regions depend on CMEK config, which depends on role policy, which depends on cluster), new regions must be managed by the CMEK resource. That means the cluster resource and CMEK resource need to keep track of which regions they "own" and reconcile that with updates and the actual state of the cluster. That's the bulk of the complexity here.

Destroying a CMEK resource is a no-op, since CMEK cannot be disabled. That means that a user could be left with "floating" regions. I need to figure out a way to block targeted destroy operations on CMEK resources while allowing them to be removed from state when the parent cluster is destroyed. We could attempt to remove additional regions on destroy, but that would take significantly longer.

Happy to do a walkthrough and discuss tradeoffs at a watercooler.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67)


adwittumuluri:
Hey evan, [here](https://github.com/cockroachlabs/managed-service/blob/ac0632e1ce8368a5c4ad771b806652546a285c21/intrusion/server/kube.go#L255) is where we Acquire a cloud provider account by ID. Seems like that's [plumbed](https://github.com/cockroachlabs/managed-service/blob/ac0632e1ce8368a5c4ad771b806652546a285c21/console/consolepb/admin.proto#L52-L58) all the way through to the admin console, and can be set via the admin-cli.

```
admin-cli cluster create [other args] --account-id [account id]
```

Seems like we're very vocal in the comments that it's for testing purposes, though if it makes the TF flow a lot easier, then we can argue against that.

ryanluu12345 (commented):
Evan, also wanted to give a reminder for a diagram that shows the flow for how this feature works, that way we can better visualize the dependencies.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-:-NJqgfrH4dQbwDZjaImT:bw05n18)* status: 0 of 13 files reviewed, all discussions resolved (waiting on @adwittumuluri, @jenngeorge, @marksoper, and @prafull01)

ryanluu12345 (commented):
Overall, this was not super hard to follow. Thanks again for the session today explaining the whole flow and reasonings for different regions lists that needed to be reconciled.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-:-NJqsOJS0116ZpNGX9-o:b-q47ilx)* status: 0 of 13 files reviewed, 4 unresolved discussions (waiting on @adwittumuluri, @erademacher, @jenngeorge, @marksoper, and @prafull01)

___
*[`examples/workflows/cockroach_cmek/main.tf` line 21 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-NJqqJu59_KEZJZcEgv8:-NJqqJu59_KEZJZcEgv9:bbcxode) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/0894069f7b1c6380069a8d6b4a030be56e86720e/examples/workflows/cockroach_cmek/main.tf#L21)):*
> ```HCL
>   type     = string
>   nullable = false
>   default  = "us-west2"
> ```

I believe for `AWS` the region would be `us-west-2`.
___
*[`internal/provider/cluster_resource.go` line 274 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-NJqqnBO-267FFJkV-eg:-NJqqnBO-267FFJkV-eh:b-2cuku) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/0894069f7b1c6380069a8d6b4a030be56e86720e/internal/provider/cluster_resource.go#L274)):*
> ```Go
> 			}
> 			dedicated.Hardware = hardware
> 			if cfg.PrivateNetworkVisibility.ValueBool() {
> ```

Wouldn't we need another CCAPI release for this to be exposed and usable? Or is this currently exposed in production?
___
*[`internal/provider/cluster_resource.go` line 551 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-NJqrEml6X1fL_DGqS3Q:-NJqrEml6X1fL_DGqS3R:b-u13qdz) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/0894069f7b1c6380069a8d6b4a030be56e86720e/internal/provider/cluster_resource.go#L551)):*
> ```Go
> 		var privateNetworkVisibility bool
> 		if networkVisibility, ok := clusterObj.AdditionalProperties["network_visibility"].(string); ok {
> 			privateNetworkVisibility = networkVisibility == "NETWORK_VISIBILITY_PRIVATE"
> ```

Nit: `NETWORK_VISIBILITY_PRIVATE` as a constant since it's reused elsewhere.
___
*[`internal/provider/cluster_resource.go` line 620 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-NJqrh-r4KCu7M21YlWb:-NJqrh-r4KCu7M21YlWc:byiaw2k) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/0894069f7b1c6380069a8d6b4a030be56e86720e/internal/provider/cluster_resource.go#L620)):*
> ```Go
> //
> // A nil return value means no region update is required.
> func reconcileRegionUpdate(ctx context.Context, state, plan []Region, clusterID string, service client.Service) (*map[string]int32, diag.Diagnostics) {
> ```

What are scenarios where these different region lists can go out of sync? Will this method address cases where we could go out of sync?

erademacher (commented):
I'm opening this back up before the single/multi key stuff because Twilio needs something to test with now. It shouldn't be a breaking change anyway.

I updated the Go SDK so we aren't doing anything janky with `AdditionalProperties` to get the cluster set up for private network visibility.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-:-NL28v0e5AiQRWYBYb5n:b-ezultn)* status: 0 of 17 files reviewed, 4 unresolved discussions (waiting on @adwittumuluri, @jenngeorge, @marksoper, @prafull01, and @ryanluu12345)

___
*[`examples/workflows/cockroach_cmek/main.tf` line 21 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-NJqqJu59_KEZJZcEgv8:-NL289uO5xCUdaCrXL_B:b-k2rwtz) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/0894069f7b1c6380069a8d6b4a030be56e86720e/examples/workflows/cockroach_cmek/main.tf#L21)):*
<details><summary><i>Previously, ryanluu12345 wrote…</i></summary><blockquote>

I believe for `AWS` the region would be `us-west-2`.
</blockquote></details>

Good catch. Fixed.
___
*[`internal/provider/cluster_resource.go` line 274 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-NJqqnBO-267FFJkV-eg:-NL28DAF5HmjcJzcqbS7:b-p9uf5v) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/0894069f7b1c6380069a8d6b4a030be56e86720e/internal/provider/cluster_resource.go#L274)):*
<details><summary><i>Previously, ryanluu12345 wrote…</i></summary><blockquote>

Wouldn't we need another CCAPI release for this to be exposed and usable? Or is this currently exposed in production?
</blockquote></details>

No, it's currently exposed in production. API releases are only needed for breaking changes.
___
*[`internal/provider/cluster_resource.go` line 551 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-NJqrEml6X1fL_DGqS3Q:-NL28L3wD3ZbQd_xd-xU:bm2lhmm) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/0894069f7b1c6380069a8d6b4a030be56e86720e/internal/provider/cluster_resource.go#L551)):*
<details><summary><i>Previously, ryanluu12345 wrote…</i></summary><blockquote>

Nit: `NETWORK_VISIBILITY_PRIVATE` as a constant since it's reused elsewhere.
</blockquote></details>

Addressed with the Go SDK update.
___
*[`internal/provider/cluster_resource.go` line 620 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-NJqrh-r4KCu7M21YlWb:-NL28MwRAZf3_WiSIGOv:bn7w678) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/0894069f7b1c6380069a8d6b4a030be56e86720e/internal/provider/cluster_resource.go#L620)):*
<details><summary><i>Previously, ryanluu12345 wrote…</i></summary><blockquote>

What are scenarios where these different region lists can go out of sync? Will this method address cases where we could go out of sync?
</blockquote></details>

Cluster and CMEK region lists go out of sync when new regions are added after CMEK is enabled. That's the whole point of this method. The only thing that it doesn't address is when users add or remove regions outside of Terraform, which they aren't supposed to do. We have no way to assign them ownership under this system. We won't accidentally remove them, though.

ryanluu12345 (approved):
> Reviewed 8 of 13 files at r1, 7 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67)* status: 15 of 17 files reviewed, 2 unresolved discussions (waiting on @adwittumuluri, @erademacher, @jenngeorge, @marksoper, and @prafull01)

___
*[`docs/data-sources/cluster.md` line 42 at r4](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-NL35gtl1gwv76ev0Ifs:-NL35gtl1gwv76ev0Ift:b9vts4q) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e588edb5ae231aa87e879bf7453805594f6db1ce/docs/data-sources/cluster.md#L42)):*
> ```Markdown
> - `memory_gib` (Number)
> - `num_virtual_cpus` (Number)
> - `private_network_visibility` (Boolean)
> ```

Wondering why you made the schema for this a boolean and not a string or enum (if supported)? The original decision for `NetworkVisibility` as an enum is to make it more extensible to future networking types without having to introduce a breaking type change.

If this is not too much of a concern and can be changed later with minimal impact, :lgtm:
___
*[`internal/provider/cluster_resource.go` line 620 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-NJqrh-r4KCu7M21YlWb:-NL36JaDDvZ9Lpmnmluj:ba5cr3e) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/0894069f7b1c6380069a8d6b4a030be56e86720e/internal/provider/cluster_resource.go#L620)):*

Cluster and CMEK region lists go out of sync when new regions are added after CMEK is enabled. That's the whole point of this method. The only thing that it doesn't address is when users add or remove regions outside of Terraform, which they aren't supposed to do. We have no way to assign them ownership under this system. We won't accidentally remove them, though.
</blockquote></details>

Thanks for the explanation. Looks good to me then.

marksoper (approved):

marksoper:
example and docs look good @erademacher - I will test locally asap

erademacher (commented):
TFTRs! I'm gonna err on the side of expediency and get this out the door.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-:-NL3GzZLArTxDiZgBdZ4:b372zqc)* status: 15 of 17 files reviewed, 2 unresolved discussions (waiting on @adwittumuluri, @jenngeorge, @prafull01, and @ryanluu12345)

___
*[`docs/data-sources/cluster.md` line 42 at r4](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/67#-NL35gtl1gwv76ev0Ifs:-NL3H4wX6-DryPVhpxNP:b-lgljks) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e588edb5ae231aa87e879bf7453805594f6db1ce/docs/data-sources/cluster.md#L42)):*
<details><summary><i>Previously, ryanluu12345 wrote…</i></summary><blockquote>

Wondering why you made the schema for this a boolean and not a string or enum (if supported)? The original decision for `NetworkVisibility` as an enum is to make it more extensible to future networking types without having to introduce a breaking type change.

If this is not too much of a concern and can be changed later with minimal impact, :lgtm:
</blockquote></details>

Simplicity, mostly. At this stage, it's relatively OK to make breaking changes. I can definitely change it to a string later.

PR #69, opened 2022-12-22
update manifest protocol version

erademacher:
This was reported by our friends at Nightfall. Our registry manifest says we use protocol version 5.0 (which is what the TF SDK v2 uses), but we're using the plugin framework, which uses protocol version 6.0.


erademacher:
TFTR!

PR #70, opened 2023-01-05
Misc bug fixes

erademacher:
- Fixed an initialization bug for datasources. It was attempting to use resource types and didn't match the Configurable interface.
- Fixed a bug where users were unable to delete allowlist and SQL user resources if the cluster had disappeared.


jenngeorge (approved):
LGTM!

erademacher:
TFTR!

PR #74, opened 2023-01-25
[CC-9116] Cluster resource allows network visibility to be toggled after creation

erademacher:
Started making use of some of the extended resource functionality
the framework provides, namely ConfigValidators and ModifyPlan.

ConfigValidators let you validate a config
by itself. You can do stuff like prevent conflicting attributes
(dedicated and serverless in this case).

ModifyPlan seems to be intended to handle things like default values,
but I'm using it to prevent users from making certain changes to
clusters. All of the apply-time validation baked into Update now
lives in this method, as well as validation for the scenario in
the commit title.

Updated the framework to v1.1.1 (hooray stability!) to make use of
ConfigValidators. No changes were necessary. The changelog suggests
that they only removed some deprecated code with the 1.0 update.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/74)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,8 +5,8 @@ go 1.18
 require (
 	github.com/golang/mock v1.6.0
 	github.com/hashicorp/terraform-plugin-docs v0.13.0
-	github.com/hashicorp/terraform-plugin-framework v0.17.0
-	github.com/hashicorp/terraform-plugin-go v0.14.2
+	github.com/hashicorp/terraform-plugin-framework v1.1.1
+	github.com/hashicorp/terraform-plugin-go v0.14.3
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.24.1
 	github.com/stretchr/testify v1.7.2
 )
@@ -31,7 +31,7 @@ require (
 	github.com/hashicorp/go-cty v1.4.1-0.20200414143053-d3edf31b6320 // indirect
 	github.com/hashicorp/go-hclog v1.2.1 // indirect
 	github.com/hashicorp/go-multierror v1.1.1 // indirect
-	github.com/hashicorp/go-plugin v1.4.6 // indirect
+	github.com/hashicorp/go-plugin v1.4.8 // indirect
 	github.com/hashicorp/go-uuid v1.0.3 // indirect
 	github.com/hashicorp/go-version v1.6.0 // indirect
 	github.com/hashicorp/hc-install v0.4.0 // indirect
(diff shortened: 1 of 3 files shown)


jenngeorge (approved):
:lgtm: !

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/74#-:-NMjdfhjDpieRFticSCr:b593w5u)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @erademacher, @fantapop, and @ryanluu12345)

___
*[`-- commits` line 5 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/74#-NMjIsVV4Af7rsr9wW1_):*
Exciting!
___
*[`internal/provider/cluster_resource.go` line 363 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/74#-NMj_VlVBIqQrCFs6Sns:-NMj_VlW4aCeMF5uD8n9:b-kgeivc) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/8f7880ce1eeb2223d76ec7527074e806c37b0950/internal/provider/cluster_resource.go#L363)):*
> ```Go
> 
> // ModifyPlan is used to make sure the user isn't trying to modify an
> // immutable attribute. We can't use `RequiresReplace` because that
> ```

Checking understanding:  Setting `RequiresReplace` for each immutable attribute will show the user in the plan that the resource will be replaced, but still let the user proceed with the update (and maybe unintentionally destroy the cluster and all its data). Adding errors in `ModifyPlan` + checking for errors in `Update` before updating the resource prevents this.

erademacher (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/74#-:-NMk1cXp24NHBAkZfMvt:b-814g8k)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @fantapop, @jenngeorge, and @ryanluu12345)

___
*[`internal/provider/cluster_resource.go` line 363 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/74#-NMj_VlVBIqQrCFs6Sns:-NMk1Ojp29pct2YTEpoQ:b5zz9tl) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/8f7880ce1eeb2223d76ec7527074e806c37b0950/internal/provider/cluster_resource.go#L363)):*
<details><summary><i>Previously, jenngeorge (Jennifer Georgevich) wrote…</i></summary><blockquote>

Checking understanding:  Setting `RequiresReplace` for each immutable attribute will show the user in the plan that the resource will be replaced, but still let the user proceed with the update (and maybe unintentionally destroy the cluster and all its data). Adding errors in `ModifyPlan` + checking for errors in `Update` before updating the resource prevents this.
</blockquote></details>

Yep, you've got it. The thinking is that the "will be replaced" warning can get buried in a bunch of other stuff and it's too easy to accidentally delete a cluster and lose all of its data, so we throw an error prompting them to explicitly `terraform destroy` it instead.

PR #75, opened 2023-01-25
Typo in region name

Sheldons92:
us-east1 is not valid, should be us-east-1


erademacher:
Not a typo actually. That example file specifies GCP as the cloud provider, and GCP doesn't use a second dash like AWS does in their region names.

PR #76, opened 2023-01-30
Allowlist resource matches wrong entry on Read

erademacher:
Fixed a logic error where we were checking if the IP address _OR_ the CIDR mask matched instead of the address _AND_mask. Added a regression test.


catj-cockroach (approved):
Looks good to me!

erademacher:
TFTR!

PR #79, opened 2023-02-08
private_networking_visibility description change - remove PrivateLink as the flag is not required 

marksoper:
@alistair-roacher noticed that the docs mentioned PrivateLink requiring private networking when it is not required.


abhinavg6 (approved):
LGTM

erademacher (approved):
LGTM, but please give this a more descriptive commit message so we can tell at a glance what it does from the commit list.

erademacher:
Almost forgot, just updating the .go file won't change the docs. You also need to run `go generate` at the project base, then add the generated files.

PR #81, opened 2023-02-22
Add logexportconfig resource

jenngeorge:
Add a new LogExportConfig resource, which manages
the log export configuration for a cluster.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/81)


erademacher (changes requested):
This looks great! So just to make sure I understand, `EnableLogExport` is declarative, right? So Create and Update are basically the same operation?

> Reviewed 3 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/81#-:-NOvcYFq-Aqfofi65CdN:b2clee3)* status: 3 of 5 files reviewed, 3 unresolved discussions (waiting on @jenngeorge)

___
*[`internal/provider/log_export_config_resource.go` line 81 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/81#-NOv2wpuClxD85FfWx8O:-NOv2wpuClxD85FfWx8P:bo8uk2n) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/d424f4016c325a420929b91db5912cf64a2e4825/internal/provider/log_export_config_resource.go#L81)):*
> ```Go
> 				},
> 				"min_level": schema.StringAttribute{
> 					Required:            true,
> ```

This probably shouldn't actually be required if it's allowed to be empty, like in the example.
___
*[`internal/provider/log_export_config_resource.go` line 181 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/81#-NOvbRAZDDpwsJ1NW9JQ:-NOvbRAZDDpwsJ1NW9JR:brd731x) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/d424f4016c325a420929b91db5912cf64a2e4825/internal/provider/log_export_config_resource.go#L181)):*
> ```Go
> 		resp.Diagnostics.AddError(
> 			"Error enabling LogExport",
> 			fmt.Sprintf("Could not enable LogExport: %v", formatAPIErrorMessage(err)),
> ```

nit: Mixing "LogExport" and "log export". Should probably standardize on "log export".
___
*[`internal/provider/log_export_config_resource.go` line 210 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/81#-NOvaYKiCS8uMFA-kXIu:-NOvaYKiCS8uMFA-kXIv:b-p5mj8t) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/d424f4016c325a420929b91db5912cf64a2e4825/internal/provider/log_export_config_resource.go#L210)):*
> ```Go
> 	minLevel := group.MinLevel.ValueString()
> 	clientGroup := client.LogExportGroup{
> 		Channels: &channels,
> ```

The fact that the client object here takes a bunch of pointers suggests that none of these fields are marked as required in the OpenAPI spec. Should they be? If so, we should probably make an API update before merging this change.

jenngeorge:
Thanks @erademacher ! Reviewable is not letting me publish comments at the moment, so responding here:

> This looks great! So just to make sure I understand, `EnableLogExport` is declarative, right? So Create and Update are basically the same operation?
> 
Right! This doc mentions that https://www.cockroachlabs.com/docs/cockroachcloud/export-logs.html?filters=gcp-log-export#update-an-existing-log-export-configuration, but I'll also update the [API reference](https://www.cockroachlabs.com/docs/api/cloud/v1#post-/api/v1/clusters/-cluster_id-/logexport) from "Create a Log Export configuration for a cluster" to "Create or update the Log Export configuration for a cluster".

> > ```go
> > 				},
> > 				"min_level": schema.StringAttribute{
> > 					Required:            true,
> > ```
> 
> This probably shouldn't actually be required if it's allowed to be empty, like in the example.
> 

Good catch! I'll ask about this while making the API update.

> 
> > ```go
> > 		resp.Diagnostics.AddError(
> > 			"Error enabling LogExport",
> > 			fmt.Sprintf("Could not enable LogExport: %v", formatAPIErrorMessage(err)),
> > ```
> 
> nit: Mixing "LogExport" and "log export". Should probably standardize on "log export".

Good catch, fixed!

> 
> > ```go
> > 	minLevel := group.MinLevel.ValueString()
> > 	clientGroup := client.LogExportGroup{
> > 		Channels: &channels,
> > ```
> 
> The fact that the client object here takes a bunch of pointers suggests that none of these fields are marked as required in the OpenAPI spec. Should they be? If so, we should probably make an API update before merging this change.

Yeah, it looks like they should probably be marked required. I'll check and make an API update!

go.mod
```diff
@@ -74,3 +74,7 @@ require (
 	google.golang.org/protobuf v1.28.1 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace (
```
jenngeorge, line 78: Removing this replacement after the sdk is updated.

jenngeorge (commented):
Here's a [crl google doc with manual testing examples](https://docs.google.com/document/d/1Ch4dB0gSn3NB1ViJnwSgYVwYTu87hf_uKsyU8KQ1K7c/edit?usp=sharing) . It includes the AWS and GCP workflow examples, plus various create / update / delete attempts that build off the examples.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/81#-:-NQclE-AB1SGZDRaa99_:bwui5a)* status: 0 of 10 files reviewed, 3 unresolved discussions (waiting on @erademacher)

erademacher (approved):
:lgtm: 

> Reviewed 3 of 7 files at r3, 3 of 6 files at r4, 4 of 4 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/81#-:-NQgX1Hq-ezS_2D9Q3FA:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jenngeorge)

___
*[`internal/provider/log_export_config_resource.go` line 212 at r6](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/81#-NQgVGmr6uN1fcO99Nnp:-NQgVGms9gjbmFbkoj1R:bjgwpwx) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/65e2ffa3d24625fb460e4903971e3e2d051e0750/internal/provider/log_export_config_resource.go#L212)):*
> ```Go
> 	var groups []LogExportGroup
> 	if len(spec.GetGroups()) != 0 {
> 		groups = make([]LogExportGroup, 0, len(spec.GetGroups()))
> ```

super nit: There probably aren't enough items in these slices for it to make a big difference, but if no items are getting filtered out, it's generally more efficient to set the length to the right size (instead of the capacity) and use the index instead of `append`.

jenngeorge (commented):
TFTR! Recent pushes include cloud provider compatibility checks that were present during manual testing but I had failed to push, rebasing, and indexing instead of appending.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/81#-:-NQka7Tx75F7S4b4FK8F:bjkkb3k)* status: 4 of 10 files reviewed, 1 unresolved discussion (waiting on @erademacher)

___
*[`internal/provider/log_export_config_resource.go` line 212 at r6](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/81#-NQgVGmr6uN1fcO99Nnp:-NQka4WSD9URGS1TplWl:b8gj3mm) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/65e2ffa3d24625fb460e4903971e3e2d051e0750/internal/provider/log_export_config_resource.go#L212)):*

super nit: There probably aren't enough items in these slices for it to make a big difference, but if no items are getting filtered out, it's generally more efficient to set the length to the right size (instead of the capacity) and use the index instead of `append`.
</blockquote></details>

Good catch, fixed!

PR #82, opened 2023-03-02
[CC-9265]Add database resource

jenngeorge:
Add a database resource, which manages a SQL
database for a cluster.


jenngeorge:
> 
> > ```go
> > 				Required: true,
> > 			},
> > 			"id": schema.StringAttribute{
> > ```
> 
> Can you add a description to this field that says what the ID format is? Users will need to know for import.
> 
Added! Also added descriptions to the computed ID fields for the allow list and sql user resources.

> 
> > ```go
> > 	}
> > 
> > 	// Since the state may have come from an import, we need to retrieve
> > ```
> 
> Imports aren't the only reason we need to read the actual state. We need to see if anything changed outside of Terraform, like if the database or underlying cluster have been deleted. I see you account for that below, which is great.
> 
Updated the comment!

> _[`internal/provider/database_resource.go` line 294 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/82#-NPYXocP3PQYzb0m4s2B:-NPYX8wk9nIdImEs8s9L:bbxrs7d) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/6d84b5db9b509dc6a1d7b2058eb511c5692f65ef/internal/provider/database_resource.go#L294)):_
> 
> > ```go
> > }
> > 
> > func loadDatabaseToTerraformState(databaseObj *client.ApiDatabase, state, plan *Database) {
> > ```
> 
> Do you just need `plan` here for the cluster ID since it isn't contained in the database model? How does that work in the Update case? Where is the plan's ID getting set, and are you sure it's going to reflect the updated name? Because you could get ID collision if the ID isn't updated and the user creates a new database with the old name. I wonder if it'd be better to just pass in the cluster ID instead of the plan and calculate the database ID here.

Good idea, updated to pass in the cluster ID. I also tested the following scenarios in staging (took [notes here](https://docs.google.com/document/d/1-I-UNz_8F2yndtLvJAR8k0xZo3NcYnuD8bEEAX0qSOY/edit?usp=sharing)):
- Attempt to update a terraformed database name to that of an existing non-terraformed database.
- Create a new non-terraformed database with a terraformed database’s old name, then import the new database to terraform.
- “name-b” and “example-database” are terraformed. “name-a” is not. Delete “name-b” outside of terraform, then try updating “name-b” to “name-a” via terraform.

erademacher (approved):
:lgtm: I appreciate the thorough testing. Nice work!

> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/82#-:-NPrxMPa0e-kgBq_3BTl:b-5oidhe)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jenngeorge)

___
*[`docs/resources/allow_list.md` line 32 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/82#-NPrwXLy3q9EekFrY5dZ:-NPrwXLz3XKPBjrNYHRY:bn1qrsz) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/50442fda1e2c8c02be015cc1e79a63bd68b68543/docs/resources/allow_list.md#L32)):*
> ```Markdown
> ### Read-Only
> 
> - `id` (String) A unique identifier with format '<cluster ID>:<CIDR IP>/<CIDR mask>'
> ```

Thanks for fixing these ones too! Didn't realize they weren't documented either :facepalm:

PR #89, opened 2023-03-12
[CC-8987] Add Client CA Cert resource

jason-crl:
Add a new TF resource to manage custom Client CAs.

NOTE: Since the corresponding new CC/console API endpoints are currently still marked `INTERNAL`, I bumped these endpoints to `PREVIEW` on my local console, rebuilt `cockroach-cloud-sdk-go` pointing to my local console, and am temp using the rebuilt sdk so I can write against the new endpoints.  Will only merge this after the new endpoints are actually bumped to `PREVIEW`.

Have a couple questions, will drop them below.  I am new to TF, so please review aggressively 😆


internal/provider/client_ca_cert_resource.go
```diff
@@ -0,0 +1,281 @@
...
+
+	"github.com/cockroachdb/cockroach-cloud-sdk-go/pkg/client"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	sdk_resource "github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+const (
+	clientCACertEndpointTimeout = time.Minute * 10 // TODO: what's a reasonable value here?
```
jason-crl, line 33: @pseudomuto what's a reasonable timeout for a cert to be loaded onto the cluster?

internal/provider/client_ca_cert_resource.go
```diff
@@ -0,0 +1,281 @@
...
+
+func (r *clientCACertResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	if r.provider == nil || !r.provider.configured {
+		addConfigureProviderErr(&resp.Diagnostics)
+		return
+	}
+
+	// Retrieve values from plan
+	var plan ClientCACertResourceModel
+	diags := req.Config.Get(ctx, &plan) // TODO: why is tfsdk.Config used here instead of tfsdk.Plan
```
jason-crl, line 87: @erademacher the TF docs/examples use `Plan.Get` here, but I noticed we use `Config.Get` instead?

internal/provider/client_ca_cert_resource.go
```diff
@@ -0,0 +1,281 @@
...
+
+func NewClientCACertResource() resource.Resource {
+	return &clientCACertResource{}
+}
+
+func (r *clientCACertResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Manages client CA certs.",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
```
jason-crl, line 48: I'm using the `id` here as the `cluster_id`, is that ok?  The idea is that the cluster_id always need to be provided, so the `id` isn't a computed, "known after apply" field (since the client CA cert feature doesn't really have a concept of it's own id)

main.go
```diff
@@ -29,7 +29,7 @@ import (
 
 // If you do not have terraform installed, you can remove the formatting command, but its suggested to
 // ensure the documentation is formatted properly.
-//go:generate terraform fmt -recursive ./examples/
+////go:generate terraform fmt -recursive ./examples/
```
jason-crl, line 32: will uncomment this before merge.

erademacher (changes requested):
Totally OK to use the cluster ID as this resource's ID.

Are you going to add an ImportState method? Are users going to want to be able to import existing certs into TF?

> Reviewed 7 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89#-:-NQR_D25CJSaRqU9mIhS:bj25q8v)* status: 7 of 9 files reviewed, 7 unresolved discussions (waiting on @jason-crl and @jenngeorge)

___
*[`examples/workflows/cockroach_client_ca_cert/main.tf` line 64 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89#-NQROTFt5GRSIt837tAb:-NQROTFt5GRSIt837tAc:bdo5qie) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e51c888c2eaa9cdbb2f862b69913e59f0fd750b4/examples/workflows/cockroach_client_ca_cert/main.tf#L64)):*
> ```HCL
> resource "cockroach_client_ca_cert" "example" {
>   id = cockroach_cluster.example.id
>   x509_pem_cert = "-----BEGIN CERTIFICATE-----\nMIIDmDCCAoCgAwIBAgIUL1h31paxTeFQkqm+BYdJaHoM81MwDQYJKoZIhvcNAQEL\nBQAwUjENMAsGA1UEAwwEdGVzdDELMAkGA1UEBhMCVVMxEzARBgNVBAgMCldhc2hp\nbmd0b24xEDAOBgNVBAcMB1NlYXR0bGUxDTALBgNVBAoMBHRlc3QwHhcNMjMwMjEw\nMDAxMDEzWhcNNDgwMjA0MDAxMDEzWjBSMQ0wCwYDVQQDDAR0ZXN0MQswCQYDVQQG\nEwJVUzETMBEGA1UECAwKV2FzaGluZ3RvbjEQMA4GA1UEBwwHU2VhdHRsZTENMAsG\nA1UECgwEdGVzdDCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBALs/XgOi\nuw2LBz3Yze07G1cNRKF8tjwI8AFjhKtrDCxBZkGPsecpn5MqHMKC+HlQHL3sVoiM\nNtIzOOV83cAODFqAecep1mtKq32LcynOUEp+Fv5ZoIt+tE13QzGl1g80t4s+IZ4V\nAKWjzDxe9LLiLsOBbU1/yg0ovVxIPXHYdOaM2HcaCpNd4f1W++9cwzFimNximq8v\n6vgDKWGOEHOVWlUbC8dcqGGQJwCTedGw7N9itkEdnMcB/KMnNxPvcbCuPfzqAMVu\nHRGAnoz4R8F+a/A1vyENVSSrax9m6xaXVtsEz9n6YdAu79KlaPdy3urDMeRBGyRP\npD/CS1yixF/wV+kCAwEAAaNmMGQwHQYDVR0OBBYEFIfK+PrY03qGYNm1QiTrUXyE\nr8w0MB8GA1UdIwQYMBaAFIfK+PrY03qGYNm1QiTrUXyEr8w0MA4GA1UdDwEB/wQE\nAwIBBjASBgNVHRMBAf8ECDAGAQH/AgEBMA0GCSqGSIb3DQEBCwUAA4IBAQBGTOyj\nuvXghvL43GWln/HrgqBCmNEkMM8xP9aubsb296sTyy4Yx6+fCDqjWLPkwg8+/4l6\nJOiUzdsVoL6skUWl6cTTI5DqQaaIfhHb0xTavDFK+fYLcPh3peif5R8DDcqTI4OU\nqQAC2bOE7SGkBQjKgdxkL42f8FpSeULTT2k4/ggdMFWhJXBvWivvTaVtEPtOabkM\nva+1dDx2QXJZhVxiUFRfZ2Vrior3Nj0AWKun5cAR/w9W3UqKDtxCcio83nUzlakO\nC0PPVuJUhaT9wIM1r6DR27VBTgQVW279Qr8AmxRZqAxG+leJUEgmvQnEkP0wORwj\nNNLEtNwKBek9MCug\n-----END CERTIFICATE-----"
> ```

It would actually be cool to store this cert as a file in the same directory and use Terraform's `file` function, i.e. `x509_pem_cert = file("cert.pem")`
___
*[`internal/provider/client_ca_cert_resource.go` line 33 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89#-NQRRgXD4XjoBqvRH2t-:-NQRPOQo3-ulFwwlMMDh:b-2s58cj) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e51c888c2eaa9cdbb2f862b69913e59f0fd750b4/internal/provider/client_ca_cert_resource.go#L33)):*
> ```Go
> 
> const (
> 	clientCACertEndpointTimeout = time.Minute * 10 // TODO: what's a reasonable value here?
> ```

Most timeouts in this repo are "it should never take this long" guesses. I wouldn't spend too much time worrying about how to fine tune it unless it's off by an order of magnitude.
___
*[`internal/provider/client_ca_cert_resource.go` line 87 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89#-NQRSM0D5yK7ifL3PqlZ:-NQRSM0D5yK7ifL3Pql_:b6x8i4p) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e51c888c2eaa9cdbb2f862b69913e59f0fd750b4/internal/provider/client_ca_cert_resource.go#L87)):*
> ```Go
> 	// Retrieve values from plan
> 	var plan ClientCACertResourceModel
> 	diags := req.Config.Get(ctx, &plan) // TODO: why is tfsdk.Config used here instead of tfsdk.Plan
> ```

In practice, there isn't much difference between the config and the plan. The config will generally have null values where the plan would have unknowns, and nulls are a little bit easier to work with. You could use either one, though.
___
*[`internal/provider/client_ca_cert_resource.go` line 117 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89#-NQRUMXy4UcVdyNHjNXl:-NQRUMXy4UcVdyNHjNXm:b-5kg5yq) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e51c888c2eaa9cdbb2f862b69913e59f0fd750b4/internal/provider/client_ca_cert_resource.go#L117)):*
> ```Go
> 		resp.Diagnostics.AddError(
> 			"Error updating Client CA Cert",
> 			fmt.Sprintf("Error updating Client CA Cert: %s", formatAPIErrorMessage(err)),
> ```

There's no need to copy the summary message. You can just use `formatAPIErrorMessage(err)` for the description. I know this pattern exists elsewhere in the provider. I need to do some cleanup.
___
*[`internal/provider/client_ca_cert_resource.go` line 163 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89#-NQRW0Kd8sw1fJOAVsxE:-NQRW0Kd8sw1fJOAVsxF:bug51uv) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e51c888c2eaa9cdbb2f862b69913e59f0fd750b4/internal/provider/client_ca_cert_resource.go#L163)):*
> ```Go
> 			resp.Diagnostics.AddError(
> 				"Error getting Client CA Cert info",
> 				fmt.Sprintf("Error getting Client CA Cert info: %s", formatAPIErrorMessage(err)))
> ```

No need to repeat the summary.
___
*[`internal/provider/client_ca_cert_resource.go` line 206 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89#-NQRW_zpAs3j2cmwY-eK:-NQRW_zpAs3j2cmwY-eL:b7q46gk) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e51c888c2eaa9cdbb2f862b69913e59f0fd750b4/internal/provider/client_ca_cert_resource.go#L206)):*
> ```Go
> 			resp.Diagnostics.AddError(
> 				"Error updating Client CA Cert",
> 				fmt.Sprintf("Error updating Client CA Cert: %s", formatAPIErrorMessage(err)),
> ```

Omit summary
___
*[`internal/provider/client_ca_cert_resource_test.go` line 32 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89#-NQRXzJJ2JixkZ2g03Aw:-NQRXzJJ2JixkZ2g03Ax:bwqj9xa) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e51c888c2eaa9cdbb2f862b69913e59f0fd750b4/internal/provider/client_ca_cert_resource_test.go#L32)):*
> ```Go
> 
> func TestIntegrationClientCACertResource(t *testing.T) {
> 	clusterName := fmt.Sprintf("tftest-cmek-%s", GenerateRandomString(4))
> ```

Update the name, please, but remember to keep it under 20 total chars.
___
*[`internal/provider/client_ca_cert_resource_test.go` line 33 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89#-NQRY3tf-IJrM__TCB1p:-NQRY3tf-IJrM__TCB1q:bd7y5em) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e51c888c2eaa9cdbb2f862b69913e59f0fd750b4/internal/provider/client_ca_cert_resource_test.go#L33)):*
> ```Go
> func TestIntegrationClientCACertResource(t *testing.T) {
> 	clusterName := fmt.Sprintf("tftest-cmek-%s", GenerateRandomString(4))
> 	clusterId := "cluster-id"
> ```

You're going to have to change this to `uuid.Nil` or something now that I've checked in format validation.
___
*[`internal/provider/client_ca_cert_resource_test.go` line 126 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89#-NQRZWgTDVlz807w57WJ:-NQRZWgTDVlz807w57WK:bmqe5f5) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e51c888c2eaa9cdbb2f862b69913e59f0fd750b4/internal/provider/client_ca_cert_resource_test.go#L126)):*
> ```Go
> }
> 
> func testClientCACertResource(t *testing.T, clusterName string, clusterId string, cert1 string, cert2 string, useMock bool) {
> ```

I know you used the cmek test as a template, but most other tests also include an acceptance test that create real resources in prod and are run before we release a new provider version. Would that be feasible for this resource? Could we use a self-signed test cert or something?

internal/provider/client_ca_cert_resource.go
jason-crl: going to leave it as 10 mins for now

jason-crl:
* Added ImportState
* Moved example workflow cert to file
* Removed repeated summary messages
* Updated test cluster name, good catch
* Acceptance test coming soon

erademacher (commented):
:lgtm_strong: 

Any idea when these endpoints are going to be ready for preview visibility?

> Reviewed 5 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89#-:-NQkEfkWA8jcU6sHlE9a:b-mzu5ho)* status: 7 of 10 files reviewed, 5 unresolved discussions (waiting on @jason-crl and @jenngeorge)

main.go
jason-crl: uncommented.

jason-crl:
@erademacher 
* endpoints are in prod! currently `PREVIEW` visibility, and just merged a change to bump them to GA.  
* added acceptance tests, looking pretty good, except:

one hiccup, the post-test destroy fails.  appears TF is trying to delete the cluster while it's still locked from the previous cert UPDATE job.  I tried adding a sleep function, but don't think that quite works.  Any trick here to wait a bit before TF attempts the cluster deletion?

```
=== RUN   TestAccClientCACertResource
=== PAUSE TestAccClientCACertResource
=== CONT  TestAccClientCACertResource
    testing_new.go:84: Error running post-test destroy, there may be dangling resources: exit status 1

        Error: Error deleting cluster

        Could not delete cluster: internal call failed: cannot delete locked cluster
--- FAIL: TestAccClientCACertResource (1534.42s)
```

erademacher (changes requested):
> Reviewed 1 of 8 files at r2, 4 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89)* status: 9 of 10 files reviewed, 5 unresolved discussions (waiting on @jason-crl and @jenngeorge)

___
*[`internal/provider/client_ca_cert_resource.go` line 219 at r4](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89#-NREMtEED6_WVYSOmsnq:-NREMtEED6_WVYSOmsnr:b8zu64m) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/858d78e9e17ed7331ab71ce3ac82b510894e2007/internal/provider/client_ca_cert_resource.go#L219)):*
> ```Go
> 	}
> 
> 	_, httpResp, err := r.provider.service.DeleteClientCACert(ctx, state.ID.ValueString())
> ```

If the delete operation locks the cluster, you need to poll until it's done, just like you do for create and update. That's probably why your acceptance test is failing.

jason-crl:
> > Reviewed 1 of 8 files at r2, 4 of 5 files at r3, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89)_ status: 9 of 10 files reviewed, 5 unresolved discussions (waiting on @jason-crl and @jenngeorge)
> 
> _[`internal/provider/client_ca_cert_resource.go` line 219 at r4](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89#-NREMtEED6_WVYSOmsnq:-NREMtEED6_WVYSOmsnr:b8zu64m) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/858d78e9e17ed7331ab71ce3ac82b510894e2007/internal/provider/client_ca_cert_resource.go#L219)):_
> 
> > ```go
> > 	}
> > 
> > 	_, httpResp, err := r.provider.service.DeleteClientCACert(ctx, state.ID.ValueString())
> > ```
> 
> If the delete operation locks the cluster, you need to poll until it's done, just like you do for create and update. That's probably why your acceptance test is failing.

* Added polling to the delete operation.
* Updated tests.

erademacher (approved):
:lgtm_strong: 

> Reviewed 1 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/89#-:-NRFPPns7-cja9xzu_px:b-uwkowd)* status: 7 of 10 files reviewed, 4 unresolved discussions (waiting on @jenngeorge)

PR #94, opened 2023-03-21
Add Terraform Provider to Grant roles

kpatron-cockroachlabs:
The added provider takes in a user id and a set of roles and then grants those roles to that user.

Note that for CC, persons, servicea accounts and groups are all valid users. That is why we have to use user ids for now until a better addressing scheme is decided on for all user types.

DO NOT MERGE UNTIL go-sdk updates

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94)


jenngeorge (changes requested):
Looking good overall! 
Could you add an example for the new resource to https://github.com/cockroachdb/terraform-provider-cockroach/tree/main/examples/resources ?
Also `go generate` will generate resource docs.

> Reviewed 3 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94#-:-NR5ukBr71lIE-rpC2ZB:b-1ddmug)* status: 3 of 7 files reviewed, 4 unresolved discussions (waiting on @kpatron-cockroachlabs)

___
*[`internal/provider/role_resource.go` line 46 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94#-NR5W-H1B8FHrcwFQOuO:-NR5W-H1B8FHrcwFQOuP:b9jzuxi) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/b2e5b214dfebfd800a28142f47f6b05e64b5ff40/internal/provider/role_resource.go#L46)):*
> ```Go
> 					stringplanmodifier.UseStateForUnknown(),
> 				},
> 				MarkdownDescription: "Always matches the user ID. Required by Terraform.",
> ```

If "id" is the user ID, then "user_id" can be removed from the schema. We do this elsewhere when the "id" is a cluster ID like 
```
	"id": schema.StringAttribute{
		Required:            true,
		MarkdownDescription: "Cluster ID",
		PlanModifiers: []planmodifier.String{
			stringplanmodifier.RequiresReplace(),
		},
	},
```
___
*[`internal/provider/role_resource.go` line 74 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94#-NR5eoma-Lc8QcKU-JNo:-NR5eoma-Lc8QcKU-JNp:b-m6qip) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/b2e5b214dfebfd800a28142f47f6b05e64b5ff40/internal/provider/role_resource.go#L74)):*
> ```Go
> 	_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse,
> ) {
> 	resp.TypeName = req.ProviderTypeName + "_role"
> ```

What do you think about something like "_user_role_grants"  ? I'm wondering if we'll want to add a different "_role" resource in the future.
___
*[`internal/provider/role_resource.go` line 162 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94#-NR5oxpR1IMS9AzhR-bR:-NR5oxpR1IMS9AzhR-bS:bo9na8k) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/b2e5b214dfebfd800a28142f47f6b05e64b5ff40/internal/provider/role_resource.go#L162)):*
> ```Go
> 	// the actual user list and make sure this one is in there.
> 	options := &client.ListRoleGrantsOptions{}
> 	apiResp, _, err := r.provider.service.ListRoleGrants(ctx, options)
> ```

Can `GetAllRolesForUser` be used here instead? The `List` endpoints are used for sql user and database resources because they don't have `Get` endpoints.
I think this method might be more similar to the [LogExportConfigResource's Read](https://github.com/cockroachdb/terraform-provider-cockroach/blob/cef53381f5b63c5985adb097a338e47323208a9f/internal/provider/log_export_config_resource.go#L294-L327).
___
*[`internal/provider/role_resource_test.go` line 32 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94#-NR5wQc12QEJqw8HBsp8:-NR5wQc12QEJqw8HBsp9:b-o79e2y) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/b2e5b214dfebfd800a28142f47f6b05e64b5ff40/internal/provider/role_resource_test.go#L32)):*
> ```Go
> 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
> )
> ```

If we can get the ID of a test user in the acceptance test org, I think we should be able to include a `TestAcc...` that sets roles for that user.

erademacher (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94)* status: 2 of 9 files reviewed, 4 unresolved discussions (waiting on @jenngeorge)

___
*[`internal/provider/role_resource.go` line 46 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94#-NR5W-H1B8FHrcwFQOuO:-NR9tsSFBdz6EUEMan5V:blrjc4n) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/b2e5b214dfebfd800a28142f47f6b05e64b5ff40/internal/provider/role_resource.go#L46)):*
<details><summary><i>Previously, kpatron-cockroachlabs (Kyle) wrote…</i></summary><blockquote>

I'm happy to do it either way. I was imitating https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/internal/provider/private_endpoint_services_resource.go#L47-L59 since it seemed like the closest analog.

I think cluster is not a great analog for this situation because for the cluster object the ID field is the cluster_id. For roles, the ID field is not the role id, it's a user_id. This might produce some semantic confusion "what's the ID of a role"? Which of the following are you thinking the model would look like?

```
type RoleGrant struct {
	ID     types.String `tfsdk:"id"`
	Roles  []Role       `tfsdk:"roles"`
}
```

```
type RoleGrant struct {
	ID     types.String `tfsdk:"user_id"`
	Roles  []Role       `tfsdk:"roles"`
}
```
</blockquote></details>

I'm actually inclined to agree with Kyle here and leave it as-is, assuming a 1-1 relationship between roles and users. A user can only have one role, right? The ID does need to be a global identifier.

kpatron-cockroachlabs (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94)* status: 2 of 9 files reviewed, 4 unresolved discussions (waiting on @erademacher and @jenngeorge)

___
*[`internal/provider/role_resource.go` line 46 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94#-NR5W-H1B8FHrcwFQOuO:-NR9w_hL3y0dk2283Cxp:b-jw7ghf) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/b2e5b214dfebfd800a28142f47f6b05e64b5ff40/internal/provider/role_resource.go#L46)):*

I'm actually inclined to agree with Kyle here and leave it as-is, assuming a 1-1 relationship between roles and users. A user can only have one role, right? The ID does need to be a global identifier.
</blockquote></details>

There is only one block per resource. The resource contains the set of roles. So no it's not 1-1 with roles but I think it is 1-1 with what you mean. For example 

```
resource "cockroach_user_role_grants" "cockroach" {
  user_id = var.user_id
  roles = [
    {
      role_name     = "ORG_ROLE_ORG_ADMIN",
      resource_type = "ORGANIZATION",
      resource_id   = ""
    },
    {
      role_name     = "ORG_ROLE_CLUSTER_ADMIN",
      resource_type = "ORGANIZATION",
      resource_id   = ""
    },
    {
      role_name     = "ORG_ROLE_ORG_MEMBER",
      resource_type = "ORGANIZATION",
      resource_id   = ""
    },
  ]
}
```

kpatron-cockroachlabs (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94)* status: 2 of 9 files reviewed, 4 unresolved discussions (waiting on @erademacher and @jenngeorge)

___
*[`internal/provider/role_resource.go` line 46 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94#-NR5W-H1B8FHrcwFQOuO:-NR9yE8bBp5EVb96e4r5:ba37cxq) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/b2e5b214dfebfd800a28142f47f6b05e64b5ff40/internal/provider/role_resource.go#L46)):*
<details><summary><i>Previously, kpatron-cockroachlabs (Kyle) wrote…</i></summary><blockquote>

There is only one block per resource. The resource contains the set of roles. So no it's not 1-1 with roles but I think it is 1-1 with what you mean. For example 

```
resource "cockroach_user_role_grants" "cockroach" {
  user_id = var.user_id
  roles = [
    {
      role_name     = "ORG_ROLE_ORG_ADMIN",
      resource_type = "ORGANIZATION",
      resource_id   = ""
    },
    {
      role_name     = "ORG_ROLE_CLUSTER_ADMIN",
      resource_type = "ORGANIZATION",
      resource_id   = ""
    },
    {
      role_name     = "ORG_ROLE_ORG_MEMBER",
      resource_type = "ORGANIZATION",
      resource_id   = ""
    },
  ]
}
```
</blockquote></details>

* one resource per user id.

erademacher (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94)* status: 2 of 9 files reviewed, 4 unresolved discussions (waiting on @jenngeorge)

___
*[`internal/provider/role_resource.go` line 46 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94#-NR5W-H1B8FHrcwFQOuO:-NRAX2e05MkXqDzcRGfc:bsnezkc) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/b2e5b214dfebfd800a28142f47f6b05e64b5ff40/internal/provider/role_resource.go#L46)):*
<details><summary><i>Previously, kpatron-cockroachlabs (Kyle) wrote…</i></summary><blockquote>

* one resource per user id.
</blockquote></details>

Gotcha. Sorry, I should've checked the schema before commenting here. The precedent in big-name Terraform providers in cases like this seems to be to call it either "<parent resource>_id" or "<parent resource>", so I still vote leaving it as `user_id`.

erademacher (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94)* status: 2 of 9 files reviewed, 4 unresolved discussions (waiting on @jenngeorge)

___
*[`internal/provider/role_resource.go` line 46 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94#-NR5W-H1B8FHrcwFQOuO:-NRAYmTUCq2gYuQFomRA:b-lzbskb) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/b2e5b214dfebfd800a28142f47f6b05e64b5ff40/internal/provider/role_resource.go#L46)):*

Gotcha. Sorry, I should've checked the schema before commenting here. The precedent in big-name Terraform providers in cases like this seems to be to call it either "<parent resource>_id" or "<parent resource>", so I still vote leaving it as `user_id`.
</blockquote></details>

Reviewable borked the formatting of that last comment. My point is, the precedent in the aws provider is to name it after the parent resource like you're doing here.

jenngeorge (commented):
LGTM after removing the replacement cockroach-cloud-sdk-go!

> Reviewed 2 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94#-:-NRZx82Z96LBxSGT8ExV:b73fon9)* status: 4 of 9 files reviewed, 3 unresolved discussions (waiting on @erademacher and @kpatron-cockroachlabs)

___
*[`internal/provider/role_resource.go` line 46 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94#-NR5W-H1B8FHrcwFQOuO:-NRZtQDz7rYrC7oF36eP:b-tnlotq) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/b2e5b214dfebfd800a28142f47f6b05e64b5ff40/internal/provider/role_resource.go#L46)):*

Reviewable borked the formatting of that last comment. My point is, the precedent in the aws provider is to name it after the parent resource like you're doing here.
</blockquote></details>

Ok `user_id` sgtm too!
___
*[`internal/provider/role_resource_test.go` line 32 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94#-NR5wQc12QEJqw8HBsp8:-NRZvLvm2h7PqqYI1LIK:bqvre75) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/b2e5b214dfebfd800a28142f47f6b05e64b5ff40/internal/provider/role_resource_test.go#L32)):*
<details><summary><i>Previously, kpatron-cockroachlabs (Kyle) wrote…</i></summary><blockquote>

What's the test org? I don't have access. Also are they different on production and staging?
</blockquote></details>

Actually I think we could skip this for now  (cc @erademacher ). I only see the prod test org "Cloud API Terraform Testing", and adding pre-existing test resources feels like it could be a separate task.

kpatron-cockroachlabs:
I have update go-sdk

jenngeorge (approved):
:lgtm: !

> Reviewed 2 of 4 files at r2, 1 of 2411 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/94#-:-NRjNg4P9H8qKyawij7x:b593w5u)* status: 5 of 2416 files reviewed, 2 unresolved discussions (waiting on @erademacher and @kpatron-cockroachlabs)

PR #95, opened 2023-03-22
Remove deleted clusters from state

erademacher:
Previously, if a cluster was deleted outside of Terraform, whether by a user action or a sweeper, attempting to delete it in Terraform would result in a "cannot delete locked cluster" error. This is only an issue for the cluster resource because GET /cluster doesn't return a 404 like most other endpoints and users can still retrieve details about deleted clusters. We just need to check the status.


jenngeorge (approved):
LGTM!

PR #96, opened 2023-03-22
Remove ExternalId condition from AWS CMEK role policy

erademacher:
Apparently, this condition is no longer advised and can cause backups to fail. It was removed from the docs shortly after this example was added.


adwittumuluri (approved):
apologies for not catching this earlier

erademacher:
All good. TFTR!

PR #102, opened 2023-04-11
Add person user data source, example role grant workflow

jenngeorge:
Add a person user data source that gets a person's user id from their email address. 
This user id can be used for the cockroach_user_role_grants resource.

Add a cockroach_user_role_grants example workflow
that includes a person user data source.


examples/data-sources/cockroach_person_user/cockroach_person_user.tf
```diff
@@ -0,0 +1,7 @@
+variable "email_address" {
+  type = string
+}
+
+data "cockroach_person_user" "cockroach" {
```
pawalt, line 5: This `cockroach_person_user` name might be a little confusing since it's unclear whether you're talking about a user in the database or the cloud. Not sure what the right naming might look like, though.

pawalt (commented):
Thanks for getting this in so quickly! All looks reasonable to me, but I'll leave the Go review to @erademacher and @kpatron-cockroachlabs since I haven't worked on this code before.

kpatron-cockroachlabs (approved):
LGTM

examples/data-sources/cockroach_person_user/cockroach_person_user.tf
jenngeorge: Yeah, I'm not sure what better naming would look like either. If we hear customers are getting confused, we can always add more detail to to data source's description.

jenngeorge:
TFTRs!

PR #105, opened 2023-04-14
Wait for lock to release on cluster update and delete

erademacher:
Previously, we assumed that all update operations were initiated by the user through Terraform, but it turns out that there are a number of locking jobs that are run automatically, like minor version upgrades. These are usually fairly short, but can still cause Terraform operations to fail.

This commit adds a wait to cluster update and delete actions if the cluster is locked. Since Terraform always performs a Read before Update or Delete, it's safe to check the cluster status in the Terraform state.


jenngeorge (approved):
LGTM!

PR #108, opened 2023-04-27
Add support for resource limits

andy-kimball:
The Cockroach Cloud API now supports a way to limit the number of request units and storage MiB that a cluster can use. This works alongside an existing limit on the $ spend a cluster can incur. A caller can choose to either specify resource limits or a spend limit, but not both. If neither is specified, then the cluster will not have any limits enforced.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/108)

### DIFF
--- a/internal/provider/cluster_resource.go
+++ b/internal/provider/cluster_resource.go
@@ -124,11 +124,27 @@ func (r *clusterResource) Schema(
 				},
 				Attributes: map[string]schema.Attribute{
 					"spend_limit": schema.Int64Attribute{
-						Required: true,
-						PlanModifiers: []planmodifier.Int64{
-							int64planmodifier.UseStateForUnknown(),
+						Optional:            true,
+						MarkdownDescription: "Spend limit in US cents.",
+					},
+					"usage_limits": schema.SingleNestedAttribute{
+						Optional: true,
+						Attributes: map[string]schema.Attribute{
+							"request_unit_limit": schema.Int64Attribute{
+								Required: true,
+								PlanModifiers: []planmodifier.Int64{
+									int64planmodifier.UseStateForUnknown(),
+								},
+								MarkdownDescription: "Maximum number of request units that the cluster can consume during the month.",
+							},
+							"storage_mib_limit": schema.Int64Attribute{
+								Required: true,
+								PlanModifiers: []planmodifier.Int64{
+									int64planmodifier.UseStateForUnknown(),
+								},
+								MarkdownDescription: "Maximum amount of storage (in MiB) that the cluster can have at any time during the month.",
+							},
 						},
-						MarkdownDescription: "Spend limit in US cents",
 					},
 					"routing_id": schema.StringAttribute{
 						Computed: true,
@@ -231,6 +247,10 @@ func (r *clusterResource) ConfigValidators(_ context.Context) []resource.ConfigV
 			path.MatchRoot("dedicated").AtName("num_virtual_cpus"),
 			path.MatchRoot("dedicated").AtName("machine_type"),
 		),
+		resourcevalidator.Conflicting(
+			path.MatchRoot("serverless").AtName("spend_limit"),
+			path.MatchRoot("serverless").AtName("usage_limits"),
+		),
 	}
 }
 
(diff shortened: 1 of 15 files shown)


internal/provider/cluster_resource.go
```diff
@@ -124,11 +124,27 @@ func (r *clusterResource) Schema(
 				},
 				Attributes: map[string]schema.Attribute{
 					"spend_limit": schema.Int64Attribute{
-						Required: true,
-						PlanModifiers: []planmodifier.Int64{
-							int64planmodifier.UseStateForUnknown(),
+						Optional:            true,
+						MarkdownDescription: "Spend limit in US cents.",
```
andy-kimball, line 128: Note that I'm not using plan modifiers. The reason is that _either_ spend limit or usage limits must be specified, but not both. `UseStateForUnknown` caused awkwardness when trying to set resource limits when a spend limit was already in place. The updated config would only specify resource limits, but Terraform would fill in the spend limit based on the previous state. This would then cause a server-side error, b/c both spend limit and resource limits were specified. Let me know if dropping `UseStateForUnknown` isn't the best way of avoiding this problem.

erademacher (approved):
:lgtm: 
I appreciate the test refactor.

> Reviewed 14 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/108#-:-NU8BW4iE2aSVmXSZnn-:b-xtm35p)* status: 14 of 15 files reviewed, 1 unresolved discussion (waiting on @andy-kimball and @jenngeorge)

___
*[`Makefile` line 15 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/108#gh-1179797605:-NU7rgFc9t-SHeEAJWHr:b-9755fn) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e5aff9149688746388aa7e3ba9f733844e22a6f6/Makefile#L15)):*
<details><summary><i>Previously, andy-kimball (Andy Kimball) wrote…</i></summary><blockquote>

Adding this just as a way of helping other developers realize what they need to do to generate updated docs, since I had to hunt to figure it out.
</blockquote></details>

Thank you!
___
*[`internal/provider/cluster_resource.go` line 128 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/108#gh-1179799168:-NU7s6962uZ1n0P4mUaQ:b5lgheo) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e5aff9149688746388aa7e3ba9f733844e22a6f6/internal/provider/cluster_resource.go#L128)):*
<details><summary><i>Previously, andy-kimball (Andy Kimball) wrote…</i></summary><blockquote>

Note that I'm not using plan modifiers. The reason is that _either_ spend limit or usage limits must be specified, but not both. `UseStateForUnknown` caused awkwardness when trying to set resource limits when a spend limit was already in place. The updated config would only specify resource limits, but Terraform would fill in the spend limit based on the previous state. This would then cause a server-side error, b/c both spend limit and resource limits were specified. Let me know if dropping `UseStateForUnknown` isn't the best way of avoiding this problem.
</blockquote></details>

At some point in the future, I plan to add a custom plan modifier to cover cases like this, but yeah, that's totally fine for now.
___
*[`internal/provider/database_resource.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/108#-NU8BE_s6YrjIBLJzsl7:-NU8BE_s6YrjIBLJzsl8:b2amkg0) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e5aff9149688746388aa7e3ba9f733844e22a6f6/internal/provider/database_resource.go#L28)):*
> ```Go
> 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
> 	"github.com/hashicorp/terraform-plugin-framework/types"
> 	"net/http"
> ```

nit: Not sure why these stdlib package imports got moved.

jenngeorge (approved):

PR #109, opened 2023-04-28
Support explicit preview CRDB versions

erademacher:
Previously, we only allowed the magic string "preview" to create new clusters with a preview crdb build. Now, the API supports cluster creation with the major version of the requested preview. That means the Terraform cluster won't get into a bad state once the version moves out of preview. This change attempts to match the version format of the state to the version format of the config when using preview versions.


jenngeorge (approved):
LGTM!

erademacher:
TFTR!

PR #118, opened 2023-05-22
Do not allow dedicated regions to be marked primary

### LINKED ISSUE #113: Provider produced inconsistent result after apply
Received an error after applying, although the cluster itself was actually created. 
```
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to module.cockroachdb_dedicated.cockroach_cluster.cluster, provider "provider[\"registry.terraform.io/cockroachdb/cockroach\"]" produced an
│ unexpected new value: .regions[0].primary: was cty.True, but now cty.False.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

Running a plan after this forced a recreation of the cluster
```

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # module.cockroachdb_dedicated.cockroach_cluster.cluster is tainted, so must be replaced
```

Terraform code
```terraform
resource "cockroach_cluster" "cluster" {
  cloud_provider = "AWS"
  name           = "test"
  regions = [
    {
      name       = "ap-southeast-1"
      node_count = 3
      primary = true
    }
  ]

  dedicated = {
    num_virtual_cpus           = 2
    storage_gib                = 100
    private_network_visibility = true
  }

}
```

Terraform versions
```
Terraform v1.3.7
on darwin_amd64
+ provider registry.terraform.io/cockroachdb/cockroach v0.4.7
+ provider registry.terraform.io/hashicorp/aws v4.66.1
```

pjtatlow:
If a dedicated cluster has a primary region, the plan that's created doesn't match the state of the cluster. This change creates an error if a dedicated cluster's region is marked as primary, and adds some documentation to specify that only serverless regions should me marked primary.

Fixes #113

### DIFF
--- a/internal/provider/cluster_resource.go
+++ b/internal/provider/cluster_resource.go
@@ -72,8 +72,9 @@ var regionSchema = schema.NestedAttributeObject{
 			},
 		},
 		"primary": schema.BoolAttribute{
-			Optional: true,
-			Computed: true,
+			Optional:    true,
+			Computed:    true,
+			Description: "Set to true to mark this region as the primary for a Serverless cluster. Exactly one region must be primary. Dedicated clusters expect to have no primary region.",
 			PlanModifiers: []planmodifier.Bool{
 				boolplanmodifier.UseStateForUnknown(),
 			},
@@ -252,6 +253,10 @@ func (r *clusterResource) ConfigValidators(_ context.Context) []resource.ConfigV
 		resourcevalidator.Conflicting(
 			path.MatchRoot("serverless").AtName("spend_limit"),
 			path.MatchRoot("serverless").AtName("usage_limits"),
+		),
+		resourcevalidator.Conflicting(
+			path.MatchRoot("regions").AtAnyListIndex().AtName("primary"),
+			path.MatchRoot("dedicated"),
 		),
 	}
 }
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -47,7 +47,7 @@ Required:
 Optional:
 
 - `node_count` (Number)
-- `primary` (Boolean)
+- `primary` (Boolean) Set to true to mark this region as the primary for a Serverless cluster. Exactly one region must be primary. Dedicated clusters expect to have no primary region.
 
 Read-Only:
 
--- a/docs/resources/cmek.md
+++ b/docs/resources/cmek.md
@@ -65,7 +65,7 @@ Required:
 Optional:
 
 - `node_count` (Number)
-- `primary` (Boolean)
+- `primary` (Boolean) Set to true to mark this region as the primary for a Serverless cluster. Exactly one region must be primary. Dedicated clusters expect to have no primary region.
 
 Read-Only:
 


internal/provider/cluster_resource.go
```diff
@@ -315,6 +316,10 @@ func (r *clusterResource) Create(
 		if plan.Regions != nil {
 			regionNodes := make(map[string]int32, len(plan.Regions))
 			for _, region := range plan.Regions {
+				if region.Primary.ValueBool() {
```
erademacher, line 319: You can actually do this validation at plan time in `ConfigValidators` instead. I think something like this should work:
```
resourcevalidator.Conflicting(
	path.MatchRoot("regions").AtAnyListIndex().AtName("primary"),
	path.MatchRoot("dedicated"),
),
```

erademacher (changes requested):

internal/provider/cluster_resource.go
pjtatlow: Done!

erademacher (approved):
Beautiful. Thanks PJ!

jaylim-crl:
I believe this PR breaks serverless cluster creation:

```bash
$ terraform plan
╷
│ Error: Invalid Attribute Combination
│
│   with cockroach_cluster.serverless,
│   on x.tf line 12, in resource "cockroach_cluster" "serverless":
│   12: resource "cockroach_cluster" "serverless" {
│
│ These attributes cannot be configured together: [regions[*].primary,dedicated]
╵
```

```tf
resource "cockroach_cluster" "serverless" {
  name           = "foo-123"
  cloud_provider = "AWS"
  serverless = {
      spend_limit = 0
  }
  regions = [
    {
      name = "us-east-1"
      primary = false
    },
    {
      name  = "eu-central-1"
      primary = true
    },
  ]
}
```

I suspect the conflict validator doesn't account for null values. Setting dedicated to `null` when specifying a primary region should be allowed.

jaylim-crl:
It appears that the following should work:

```go
resourcevalidator.Conflicting(
	path.MatchRoot("regions").AtAnyListIndex().AtName("primary"),
	path.MatchRoot("dedicated"),
),
```

Hm, not sure why that doesn't work as intended.

jaylim-crl:
I see. This worked:

```tf
resource "cockroach_cluster" "serverless" {
  name           = "foo-123"
  cloud_provider = "AWS"
  serverless = {
      spend_limit = 0
  }
  regions = [
    {
      name = "us-east-1"
      primary = false
    },
    {
      name  = "eu-central-1"
    },
  ]
}
```

But not:

```tf
resource "cockroach_cluster" "serverless" {
  name           = "foo-123"
  cloud_provider = "AWS"
  serverless = {
      spend_limit = 0
  }
  regions = [
    {
      name = "us-east-1"
      primary = false
    },
    {
      name  = "eu-central-1"
      primary = true
    },
  ]
}
```

I think if we tried to set `primary` on more than one regions, it would error out with the conflict error.

PR #120, opened 2023-06-05
Fix absent pagination when retrieving existing SQL users

samdotci:
At present, the provider does not use pagination to retrieve the list of existing SQL users when verifying that a user in the state actually exists. As such, the provider pulls only the first page of SQL users

Subsequently, a cluster with a large number of SQL users will end up with warnings during each Terraform run of `This cluster doesn't have a SQL user named... Removing from state.`. Attempting to apply will end up with the provider erroring as it attempts to create new users that conflict with existing ones.


erademacher (approved):
Thanks for your contribution, Sam! Everything looks great.

This is also likely a problem for other resources that use a List* API, so I'll create an internal ticket to audit and address those. I appreciate you bringing this to our attention!

Issue #121, opened 2023-06-08
Provider fails to implement pagination for >100 SQL users, causing user creation conflicts


samdotci:
## Issue Description:
The Terraform provider incorrectly handles pagination when verifying SQL users in the state. When more than 100 SQL users are present, it only retrieves the first 100 due to the default page size. This causes warnings during each Terraform run for a SQL cluster with more than 100 users and errors when applying changes due to conflicts with existing, unfetched users.

The warning during the plan appears as below:
> _This cluster doesn't have a SQL user named... Removing from state.._

## Expected Behavior

The provider should correctly implement pagination to verify all SQL users when their count exceeds 100.

## Steps to Reproduce:

1. Write Terraform code that would provision >100 SQL users for a cluster
2. Execute a `terraform apply`
3. Observe the success of the apply (assuming there were no existing users in the state)
4. Execute a `terraform apply` again
5. Observe the warnings indicating that some SQL users are missing, and that they will need to be recreated
6. Try to apply changes and observe the resultant errors due to user creation conflicts.

## Proposed Solution

I have created a [PR](https://github.com/cockroachdb/terraform-provider-cockroach/pull/120) to resolve this issue.

erademacher:
Fixed in v0.6.0.

PR #122, opened 2023-06-09
Use pagination with List* API endpoints for Read operations

erademacher:
[issue #121](https://github.com/cockroachdb/terraform-provider-cockroach/issues/121) revealed a problem with resources where we need to use a List endpoint to read the resource state. Namely, there is a default page size and we aren't using pagination. That results in state getting deleted because the provider thinks the resource has disappeared.

The customer who discovered the issue already fixed it for SQL users. This commit fixes it for databases, IP allowlist entries, and user role grants. It also bumps the default page size for SQL users.

Also fixes the cluster resource acceptance tests, since the supported versions have changed.


pawalt (approved):
LGTM pending acceptance testing

erademacher:
TFTR!

PR #123, opened 2023-06-09
Use retryablehttp HTTP client

erademacher:
Use the drop-in [retryablehttp](https://pkg.go.dev/github.com/hashicorp/go-retryablehttp) package as the HTTP client in the API client config, with default retry settings. This gives us automatic retries with exponential backoff. We just pass in the global logger (so that retries and their errors are logged) and specify that it should pass through the final error to the caller instead of a generic message about exceeding the number of retries.

The default number of retries is 4. Initial backoff is 1 second and doubles with each attempt, up to 30 seconds max.


fantapop (approved):
nice. Looks good!

erademacher:
TFTR!

PR #125, opened 2023-06-13
Add resource for cluster maintenance windows

dasrirez:
This patch adds a new resource to the cockroach terraform provider for managing cluster maintenance windows.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/125)


dasrirez:
This addition requires an SDK update to 1.2.0. I've included the update in this PR, but can split that into a separate PR if preferred.

erademacher (changes requested):
> Reviewed 19 of 20 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/125)* status: all files reviewed, 5 unresolved discussions (waiting on @dasrirez and @fantapop)

___
*[`examples/resources/cockroach_maintenance_window/cockroach_maintenance_window.tf` line 7 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/125#-NY04beL1qeufyTg7uG-:-NY04beL1qeufyTg7uG0:bb945i5) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/cbece8c36150b7108f7555956bc99538969b2a4f/examples/resources/cockroach_maintenance_window/cockroach_maintenance_window.tf#L7)):*
> ```HCL
> resource "cockroach_maintenance_window" "example" {
>   id              = var.cluster_id
>   offset_duration = "172800s"
> ```

nit: Can you turn theses durations into variables with these values as defaults? Maybe also add some descriptive comments, i.e. "2 days" and "6 hours".
___
*[`internal/provider/maintenance_window.go` line 38 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/125#-NY057OQ3rrzKZ6OS2rn:-NY057OQ3rrzKZ6OS2ro:bt7412w) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/cbece8c36150b7108f7555956bc99538969b2a4f/internal/provider/maintenance_window.go#L38)):*
> ```Go
> 		MarkdownDescription: "Cluster ID",
> 	},
> 	"offset_duration": schema.StringAttribute{
> ```

If you're planning on eventually allowing users to pass in other time formats, using strings here is fine. If these are always going to be a value in seconds, they should be ints instead of strings. Then users can do math to compute them. You'd probably want to also tack on a "_seconds" to their names.
___
*[`internal/provider/maintenance_window.go` line 118 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/125#-NY05qHA2DinFu1owx-O:-NY05qHA2DinFu1owx-P:b5h8ufq) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/cbece8c36150b7108f7555956bc99538969b2a4f/internal/provider/maintenance_window.go#L118)):*
> ```Go
> 			resp.Diagnostics.AddWarning(
> 				"Maintenance window not found",
> 				fmt.Sprintf("Maintenance window for cluster ID %s not found.", clusterID))
> ```

nit: It's helpful to tell the user when we remove a resource from the state.
___
*[`internal/provider/maintenance_window.go` line 171 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/125#-NY07VkN0t7D80gQOcP2:-NY07VkN0t7D80gQOcP3:b-and86j) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/cbece8c36150b7108f7555956bc99538969b2a4f/internal/provider/maintenance_window.go#L171)):*
> ```Go
> 		return
> 	}
> 	_, _, err := r.provider.service.DeleteMaintenanceWindow(ctx, state.ID.ValueString())
> ```

Will this return an error if the window has already been deleted? If so, you should check the HTTP status and skip the error if it's a 404. Otherwise users won't be able to delete the resource.
___
*[`internal/provider/maintenance_window_test.go` line 35 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/125#-NY080B34MP1-CmBjWwC:-NY080B34MP1-CmBjWwD:b-r9ifgn) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/cbece8c36150b7108f7555956bc99538969b2a4f/internal/provider/maintenance_window_test.go#L35)):*
> ```Go
> // real cluster. It will be skipped if TF_ACC isn't set.
> func TestAccMaintenanceWindowResource(t *testing.T) {
> 	t.Skip("Skipping until we can either integrate the AWS provider " +
> ```

Is this relevant? Why do you need to use the AWS provider?

erademacher (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/125)* status: 17 of 20 files reviewed, 2 unresolved discussions (waiting on @dasrirez and @fantapop)

___
*[`internal/provider/maintenance_window.go` line 38 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/125#-NY057OQ3rrzKZ6OS2rn:-NYOeY2O25Ka4m0kYyuX:b-rivgn5) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/cbece8c36150b7108f7555956bc99538969b2a4f/internal/provider/maintenance_window.go#L38)):*
<details><summary><i>Previously, dasrirez (Darius Asri Rezaei) wrote…</i></summary><blockquote>

The format that the API accepts is currently 1:1 with how the value of `google.protobuf.Duration` is json serialized. This is currently a float formatted as `<seconds>.<nanos>s`. Its unlikely, but there could a point where the format changes, in which case it would be easier to adjust the format of the string at the client end as opposed to the format of the resource from the provider end. That said I don't feel too strongly about keeping as a string, and am happy to change it if we still want to.
</blockquote></details>

Yeah, let's change it to an int if we're locked into the format. Then users can do math without having to worry about string formatting. I don't think changing the name is necessary, though, since "seconds" is in the description.
___
*[`internal/provider/maintenance_window_test.go` line 35 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/125#-NY080B34MP1-CmBjWwC:-NYOfD41DQHZIpVDyCh5:b22qnac) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/cbece8c36150b7108f7555956bc99538969b2a4f/internal/provider/maintenance_window_test.go#L35)):*
<details><summary><i>Previously, dasrirez (Darius Asri Rezaei) wrote…</i></summary><blockquote>

Nope, you're right there is no infra specific logic involved in maintenance windows. This has been removed!
</blockquote></details>

I wasn't saying you should remove it :) Just don't skip it. It should run fine, right? Let me know if I need to flip a feature flag in the test org.

dasrirez (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/125)* status: 15 of 20 files reviewed, 2 unresolved discussions (waiting on @erademacher and @fantapop)

___
*[`internal/provider/maintenance_window.go` line 38 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/125#-NY057OQ3rrzKZ6OS2rn:-NYTW1Cv3po8avoDfOs0:byku48u) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/cbece8c36150b7108f7555956bc99538969b2a4f/internal/provider/maintenance_window.go#L38)):*

Yeah, let's change it to an int if we're locked into the format. Then users can do math without having to worry about string formatting. I don't think changing the name is necessary, though, since "seconds" is in the description.
</blockquote></details>

SGTM, I've changed the durations from strings to numbers.
___
*[`internal/provider/maintenance_window_test.go` line 35 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/125#-NY080B34MP1-CmBjWwC:-NYTZqWE2rIE1sme_2Kv:b-wg625n) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/cbece8c36150b7108f7555956bc99538969b2a4f/internal/provider/maintenance_window_test.go#L35)):*

I wasn't saying you should remove it :) Just don't skip it. It should run fine, right? Let me know if I need to flip a feature flag in the test org.
</blockquote></details>

Oh I see, yes it should run fine since the FF was enabled on all orgs recently. I've added the test back without the skip.

erademacher (approved):
:lgtm_strong: 

> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/125#-:-NYTbSnB6gUXnn-FCdHL:b-uwkowd)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @fantapop)

PR #133, opened 2023-07-12
internal/provider: add network allowlist support for serverless clusters

jaylim-crl:
Previously, we disallowed network allowlist usage for serverless clusters. Now
that CockroachCloud supports this feature for all serverless clusters, we will
remove that guard, and this commit does that.

### DIFF
--- a/internal/provider/networking_resource.go
+++ b/internal/provider/networking_resource.go
@@ -130,23 +130,6 @@ func (r *allowListResource) Create(
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


jaylim-crl:
@pjtatlow I've decided to split it out into its own PR instead of combining it with the Private Endpoints work.

pjtatlow (approved):
LGTM! and thank you for adding the tests for me!

erademacher (approved):

jaylim-crl:
TFTR!

PR #134, opened 2023-07-12
internal/provider: fix acceptance tests for IP allowlist

jaylim-crl:
Previously, the tests accidentally swapped both the `sql` and `ui` fields. The
dedicated tests were not affected since both values were true. For the
serverless ones, the values were different, and the test was failing. This
commit fixes that.


erademacher (approved):

pjtatlow (approved):
LGTM!

PR #136, opened 2023-07-13
[CC-25142] Cluster tests for import and datasource

erademacher:
- Changed cluster datasource to use loadClusterToTerraformState to share code with the resource
- Added import steps to dedicated cluster tests
- Added datasources to configs for all cluster tests
- Added a test to make sure cluster resource and datasource stay in sync

I'm planning on a followup PR to add similar tests for other resources/datasources.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - New cockroach_version_deferral resource allows users to defer automated minor version
   upgrades by a fix 60-day interval.
+- Allowlists and AWS PrivateLink services are now supported for serverless clusters. PrivateLink for serverless clusters
+  is in [limited access](https://www.cockroachlabs.com/docs/v23.1/cockroachdb-feature-availability).
+
+### Fixed
+
+- Fixed an issue where the provider could crash when importing a serverless cluster.
+- Fixed an issue where usage limits weren't reported properly in cockroach_cluster data sources.
 
 ## [0.6.0] - 2023-06-21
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - New cockroach_version_deferral resource allows users to defer automated minor version
   upgrades by a fix 60-day interval.
+- Allowlists and AWS PrivateLink services are now supported for serverless clusters.
```
jaylim-crl, line 14: One thing to note is that AWS PrivateLink is guarded behind a feature flag, so it will only work on some orgs/clusters. Not sure if we should clarify it there.

internal/provider/cockroach_cluster_data_source.go
```diff
@@ -207,40 +203,7 @@ func (d *clusterDataSource) Read(
...
-		reg := Region{
-			Name:      types.StringValue(r.Name),
-			SqlDns:    types.StringValue(r.SqlDns),
-			UiDns:     types.StringValue(r.UiDns),
-			NodeCount: types.Int64Value(int64(r.NodeCount)),
-			Primary:   types.BoolValue(r.GetPrimary()),
-		}
-		cluster.Regions = append(cluster.Regions, reg)
-	}
+	loadClusterToTerraformState(cockroachCluster, &cluster, nil)
```
jaylim-crl, line 206: nit: maybe a comment stating that passing in `nil` as plan is intentional?

jaylim-crl (approved):
LGTM.

CHANGELOG.md
erademacher: Good callout. Added that info and linked to the feature availability doc.

internal/provider/cockroach_cluster_data_source.go
erademacher: Done.

erademacher:
TFTR!

PR #139, opened 2023-07-13
Test fixes

erademacher:
- Updated cluster state import logic to check the plan for which limit type to set
- Updated tests that use machine_type to use num_virtual_cpus instead, for futureproofing

Running tests manually now. Will make sure they're passing before merging.

### DIFF
--- a/internal/provider/cluster_resource.go
+++ b/internal/provider/cluster_resource.go
@@ -794,15 +794,22 @@ func loadClusterToTerraformState(
 			RoutingId: types.StringValue(clusterObj.Config.Serverless.RoutingId),
 		}
 
-		// Set either the spend limit or usage limits, depending on the cluster
-		// object requested. Both options are returned by the API.
-		if clusterObj.Config.Serverless.UsageLimits != nil {
+		// The whole plan or just its serverless config can be nil.
+		var planConfig *ServerlessClusterConfig
+		if plan != nil {
+			planConfig = plan.ServerlessConfig
+		}
+
+		// Set either the spend limit or usage limits, depending on the plan.
+		// Both options are returned by the API. If there's no plan (import or
+		// data source), we default to usage limits.
+		if clusterObj.Config.Serverless.UsageLimits != nil && (planConfig == nil || planConfig.UsageLimits != nil) {
 			usageLimits := clusterObj.Config.Serverless.UsageLimits
 			serverlessConfig.UsageLimits = &UsageLimits{
 				RequestUnitLimit: types.Int64Value(usageLimits.RequestUnitLimit),
 				StorageMibLimit:  types.Int64Value(usageLimits.StorageMibLimit),
 			}
-		} else if clusterObj.Config.Serverless.SpendLimit != nil {
+		} else if clusterObj.Config.Serverless.SpendLimit != nil && planConfig != nil && !planConfig.SpendLimit.IsNull() {
 			serverlessConfig.SpendLimit = types.Int64Value(int64(clusterObj.Config.Serverless.GetSpendLimit()))
 		}
 		state.ServerlessConfig = serverlessConfig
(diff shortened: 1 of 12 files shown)


internal/provider/networking_resource_test.go
```diff
@@ -315,7 +315,7 @@ resource "cockroach_cluster" "dedicated" {
     cloud_provider = "AWS"
     dedicated = {
         storage_gib = 15
-        machine_type = "m5.large"
+        num_virtual_cpus = 2
```
jaylim-crl, line 318: Some of the states like this one needs to be updated as well: https://github.com/cockroachdb/terraform-provider-cockroach/blob/725942c94aa64b9a252a7fb5e69a03cf70c15b42/internal/provider/networking_resource_test.go#L130-L135

jaylim-crl (approved):

internal/provider/cluster_resource.go
```diff
@@ -794,15 +794,22 @@ func loadClusterToTerraformState(
 			RoutingId: types.StringValue(clusterObj.Config.Serverless.RoutingId),
 		}
 
-		// Set either the spend limit or usage limits, depending on the cluster
-		// object requested. Both options are returned by the API.
-		if clusterObj.Config.Serverless.UsageLimits != nil {
+		// The whole plan or just its serverless config can be nil.
+		var planConfig *ServerlessClusterConfig
```
andy-kimball, line 798: Consider doing something like this to avoid the complicated-looking conditional logic:
```
forceSpendLimit := false
if plan != nil && plan.ServerlessConfig != nil {
  forceSpendLimit = !plan.ServerlessConfig.SpendLimit.IsNull()
}
...
if !forceSpendLimit && clusterObj.Config.Serverless.UsageLimits != nil {
} else if clusterObj.Config.Serverless.SpendLimit != nil {
}
```

internal/provider/cluster_resource_test.go
```diff
@@ -370,8 +387,8 @@ func serverlessClusterWithNoLimits(clusterName string) resource.TestStep {
 			resource.TestCheckNoResourceAttr(resourceName, "serverless.spend_limit"),
 			resource.TestCheckNoResourceAttr(resourceName, "serverless.usage_limits"),
 			resource.TestCheckResourceAttr(dataSourceName, "name", clusterName),
-			resource.TestCheckNoResourceAttr(dataSourceName, "serverless.spend_limit"),
```
andy-kimball, line 373: Why not assert the `spend_limit` field doesn't exist? Is the idea that we'll be getting rid of that field?

andy-kimball (approved):
LGTM, with a couple suggestions.

internal/provider/cluster_resource.go
erademacher: Unfortunately, that doesn't work in the case where the plan exists but doesn't contain either field.  The API response will say the spend limit is zero, but the state needs to be null to match the plan. I'm sure there's still a way to simplify this, but my brain is in vacation mode :)

internal/provider/cluster_resource_test.go
erademacher: Yeah, my thought was that it's just one less thing we'll have to change, but that's kind of a lame reason. I added the check back.

erademacher:
TFTRs!

internal/provider/cluster_resource.go
andy-kimball: Yes, you're right, I forgot that the API returns `0/nil` rather than `nil/nil` for this case. I'd probably modify my code above to be like this:
```
forceSpendLimit := false
if plan != nil && plan.ServerlessConfig != nil {
  forceSpendLimit = !plan.ServerlessConfig.SpendLimit.IsNull()
}
...
if !forceSpendLimit && clusterObj.Config.Serverless.UsageLimits != nil {
} else {
  spendLimit := clusterObj.Config.Serverless.SpendLimit
  if spendLimit != nil && *spendLimit != 0 {
  }
}
```

PR #140, opened 2023-07-17
[CC-25137] support modifying serverless cluster regions

pjtatlow:
This change updates `cockroach-cloud-sdk-go` to support updating serverless regions.


internal/provider/cluster_resource.go
```diff
@@ -602,6 +602,24 @@ func (r *clusterResource) Update(
...
+		if plan.Regions != nil {
+			var regions []string
+			var primaryRegion string
+			for _, region := range plan.Regions {
+				if region.Primary.ValueBool() {
+					primaryRegion = region.Name.ValueString()
+				}
+				regions = append(regions, region.Name.ValueString())
+			}
+			if len(regions) > 1 && primaryRegion == "" {
```
jaylim-crl, line 614: Should it be an error if `len(regions) == 0` as well, or is that not an issue?

jaylim-crl (approved):
LGTM.

internal/provider/cluster_resource.go
pjtatlow: I don't think it's an issue, it will just fail later I think. Dedicated isn't doing that at the moment, so I'm probably going to leave it as is.

PR #142, opened 2023-07-19
internal/provider: address a test TODO for private endpoint services

jaylim-crl:
**NOTE: Only merge this end of the week.**

----

Previously, we asserted that the private endpoint services returned by the
server span all regions of the host (for a serverless cluster), and not just
the regions of the serverless cluster. This was a bug in CockroachCloud itself,
and has already been addressed there. This commit updates that assertion to
the number of regions on the serverless cluster, addressing a TODO.


pjtatlow (approved):
LGTM!

jaylim-crl:
TFTR!

PR #143, opened 2023-07-20
internal/provider: expose the internal_dns field within the region model

jaylim-crl:
Previously, there was no way to obtain the internal_dns field via the
terraform provider; the CC API does return the internal_dns field. This
commit updates the provider to include that field so that users could use it
for VPC Peering or AWS PrivateLink.

### DIFF
--- a/internal/provider/models.go
+++ b/internal/provider/models.go
@@ -29,11 +29,12 @@ const (
 )
 
 type Region struct {
-	Name      types.String `tfsdk:"name"`
-	SqlDns    types.String `tfsdk:"sql_dns"`
-	UiDns     types.String `tfsdk:"ui_dns"`
-	NodeCount types.Int64  `tfsdk:"node_count"`
-	Primary   types.Bool   `tfsdk:"primary"`
+	Name        types.String `tfsdk:"name"`
+	SqlDns      types.String `tfsdk:"sql_dns"`
+	UiDns       types.String `tfsdk:"ui_dns"`
+	InternalDns types.String `tfsdk:"internal_dns"`
+	NodeCount   types.Int64  `tfsdk:"node_count"`
+	Primary     types.Bool   `tfsdk:"primary"`
 }
 
 type DedicatedClusterConfig struct {
--- a/docs/data-sources/cluster.md
+++ b/docs/data-sources/cluster.md
@@ -49,6 +49,7 @@ Read-Only:
 
 Read-Only:
 
+- `internal_dns` (String)
 - `name` (String)
 - `node_count` (Number)
 - `primary` (Boolean)
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -51,6 +51,7 @@ Optional:
 
 Read-Only:
 
+- `internal_dns` (String)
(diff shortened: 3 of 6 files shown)


internal/provider/models.go
```diff
@@ -29,11 +29,12 @@ const (
...
 type Region struct {
-	Name      types.String `tfsdk:"name"`
-	SqlDns    types.String `tfsdk:"sql_dns"`
-	UiDns     types.String `tfsdk:"ui_dns"`
-	NodeCount types.Int64  `tfsdk:"node_count"`
-	Primary   types.Bool   `tfsdk:"primary"`
+	Name        types.String `tfsdk:"name"`
+	SqlDns      types.String `tfsdk:"sql_dns"`
+	UiDns       types.String `tfsdk:"ui_dns"`
+	InternalDns types.String `tfsdk:"internal_dns"`
```
jaylim-crl, line 35: I've manually tested this, and the provider behaves as expected: 

```json
    {
      "mode": "managed",
      "type": "cockroach_cluster",
      "name": "example",
      "provider": "provider[\"registry.terraform.io/cockroachdb/cockroach\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "account_id": "",
            "cloud_provider": "AWS",
            "cockroach_version": "v23.1",
            "creator_id": "a66837d8-1a2d-4290-afbf-e893a26037e4",
            "dedicated": null,
            "id": "10fb98e2-18a4-476a-8a15-c236d3a2502e",
            "name": "<redacted>",
            "operation_status": "UNSPECIFIED",
            "plan": "SERVERLESS",
            "regions": [
              {
                "internal_dns": "<redacted>-36.internal-75rf.aws-ap-south-1.<redacted>",
                "name": "ap-south-1",
                "node_count": 0,
                "primary": true,
                "sql_dns": "<redacted-cluster-name>.aws-ap-south-1.<redacted>",
                "ui_dns": ""
              },
              {
                "internal_dns": "<redacted>-36.internal-75rf.aws-us-east-1.<redacted>",
                "name": "us-east-1",
                "node_count": 0,
                "primary": false,
                "sql_dns": "<redacted-cluster-name>.aws-us-east-1.<redacted>",
                "ui_dns": ""
              }
            ],
            "serverless": {
              "routing_id": "<redacted>-36",
              "spend_limit": 0,
              "usage_limits": null
            },
            "state": "CREATED",
            "upgrade_status": "FINALIZED"
          },
          "sensitive_attributes": []
        }
      ]
    },
```

I see a few current issues:
1. The tests only cover GCP clusters, and the InternalDNS field will only show up for supported AWS Serverless clusters. I think implementing some sort of data driven testing that covers both AWS and GCP will be beneficial for the long term.
2. The `sql_dns` field returned by the CC API is incorrect. We will need to fix this in CC's codebase. This isn't an issue today because the connect modal only displays the global DNS.
3. We don't return the global `sql_dns` field for the cluster, which is a bit unfortunate.

pjtatlow (approved):
LGTM!

jaylim-crl:
TFTR!

PR #147, opened 2023-07-27
[CC-25232] Use plan instead of config

erademacher:
For most resources, this is a simple find & replace.

For the database resource, I had it read the database from the create response and load it into the state instead of copying the plan (which can contain unknowns).

Private endpoint services is a weird case. Apparently, Terraform doesn't like it when Unknown values are loaded into non-Terraform types. Thankfully, this is only a problem for fully computed attributes. Required and optional attributes are loaded as nil. To resolve this, though, I had to change the data type for `services` to `types.List` and pass in the exact schema type to its constructor. I pulled the schema object out of the Schema method so I could extract the attribute type without dupicating code.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)

### DIFF
--- a/internal/provider/private_endpoint_services_resource.go
+++ b/internal/provider/private_endpoint_services_resource.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/cockroachdb/cockroach-cloud-sdk-go/pkg/client"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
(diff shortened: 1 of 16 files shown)


internal/provider/private_endpoint_services_resource.go
```diff
@@ -237,8 +242,17 @@ func loadEndpointServicesIntoTerraformState(
...
 			azs[j] = types.StringValue(az)
 		}
-		state.Services[i].Aws.AvailabilityZoneIds = azs
+		services[i].Aws.AvailabilityZoneIds = azs
 	}
+	var diags diag.Diagnostics
+	state.Services, diags = types.ListValueFrom(
+		ctx,
+		// Yes, it really does apparently need to be this complicated.
+		// https://github.com/hashicorp/terraform-plugin-framework/issues/713
```
jenngeorge, line 251: thanks for this link!

jenngeorge (approved):

erademacher:
TFTR!

PR #153, opened 2023-08-08
[DO NOT MERGE] adding api_oidc_config support

kpatron-cockroachlabs:
This change adds a new API OIDC Config resource type.

**Commit checklist**
- [ ] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)


kpatron-cockroachlabs:
This requires a release of go sdk and for conflicts with the newer go-sdk to be resolved before this can merge

erademacher (changes requested):
> Reviewed 6 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153)* status: 6 of 17 files reviewed, 5 unresolved discussions (waiting on @jenngeorge and @kpatron-cockroachlabs)

___
*[`internal/provider/api_oidc_config.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-Nc8o2ylAJJwJupyLhBX:-Nc8o2ylAJJwJupyLhBY:bovggz0) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e89e55b4036a3e78946abf69276f81d2f6e67e61/internal/provider/api_oidc_config.go#L47)):*
> ```Go
> 					stringplanmodifier.UseStateForUnknown(),
> 				},
> 				MarkdownDescription: "ID of the API OIDC Configuration. Required by Terraform.",
> ```

Is this an actual UUID that a user would have on hand for `terraform import`? If it is, the "Required by Terraform" wording isn't really necessary. If this is a synthetic ID, please describe how to build it so users can import their existing configs.
___
*[`internal/provider/api_oidc_config.go` line 63 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-Nc8MSJbCsEyWLIwj2Xg:-Nc8MSJbCsEyWLIwj2Xh:b-f8gnwv) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e89e55b4036a3e78946abf69276f81d2f6e67e61/internal/provider/api_oidc_config.go#L63)):*
> ```Go
> 			"claim": schema.StringAttribute{
> 				Required:    true,
> 				Description: "The JWT claim that should be used as the user identifier. Defaults to the subject if an empty string is provided.",
> ```

Does this actually need to be required? Could the user omit it instead of passing in an empty string?
___
*[`internal/provider/api_oidc_config.go` line 67 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-Nc8mqYc7rOXJTXFsJkw:-Nc8mqYc7rOXJTXFsJkx:b-svh3ql) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e89e55b4036a3e78946abf69276f81d2f6e67e61/internal/provider/api_oidc_config.go#L67)):*
> ```Go
> 			"identity_map": schema.StringAttribute{
> 				Required:    true,
> 				Description: "The mapping rules to convert token user identifiers into a new form.",
> ```

I wasn't able to find anything on the map syntax here through a google search. Do values for this field follow some sort of standard OIDC syntax? If so, it'd be helpful to link to a doc or something. If this is our proprietary syntax, it would definitely be good to break this out into an object with all the map components in a separate field.
___
*[`internal/provider/api_oidc_config.go` line 150 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-Nc8okGWDdGvv8a0R5ez:-Nc8okGWDdGvv8a0R5f-:b-5wgk7i) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e89e55b4036a3e78946abf69276f81d2f6e67e61/internal/provider/api_oidc_config.go#L150)):*
> ```Go
> 			resp.Diagnostics.AddWarning(
> 				"API OIDC Config not found",
> 				"API OIDC Config not found. Maintenance window will be removed from state.")
> ```

s/Maintenance window/API OIDC Config
___
*[`internal/provider/api_oidc_config_test.go` line 35 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-Nc8pBfv3G653bqB6vY0:-Nc8pBfv3G653bqB6vY1:b-z7sbdh) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e89e55b4036a3e78946abf69276f81d2f6e67e61/internal/provider/api_oidc_config_test.go#L35)):*
> ```Go
> // TestIntegrationApiOIdcConfigResource attempts to create, check, and destroy
> // an API OIDC Config, but uses a mocked API service.
> func TestIntegrationApiOIdcConfigResource(t *testing.T) {
> ```

Could you also add an acceptance test shell with `t.Skip()` and a comment about why it can't be implemented?

kpatron-cockroachlabs (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153)* status: 6 of 17 files reviewed, 5 unresolved discussions (waiting on @erademacher and @jenngeorge)

___
*[`internal/provider/api_oidc_config.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-Nc8o2ylAJJwJupyLhBX:-Nc8y2jl4L4LyQzDaxgN:bl2xkur) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e89e55b4036a3e78946abf69276f81d2f6e67e61/internal/provider/api_oidc_config.go#L47)):*

Is this an actual UUID that a user would have on hand for `terraform import`? If it is, the "Required by Terraform" wording isn't really necessary. If this is a synthetic ID, please describe how to build it so users can import their existing configs.
</blockquote></details>

This is a UUID the backend creates. I just copied the wording from somewhere else. Happy to delete that bit.
___
*[`internal/provider/api_oidc_config.go` line 63 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-Nc8MSJbCsEyWLIwj2Xg:-Nc8yBI58oXbo6bkFkBr:b-3ff5jo) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e89e55b4036a3e78946abf69276f81d2f6e67e61/internal/provider/api_oidc_config.go#L63)):*

Does this actually need to be required? Could the user omit it instead of passing in an empty string?
</blockquote></details>

I was struggling with our API. Basically you can omit it just fine but it confuses TF because when you get the field from the API it returns an empty string which is "different" from being omitted. I tried making it optional with a default value of empty string but the references I found to set a default empty string didn't work here for some reason (stringdefault was missing if I remember correctly).
___
*[`internal/provider/api_oidc_config.go` line 67 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-Nc8mqYc7rOXJTXFsJkw:-Nc8yba5436rB5vxmdqm:bsm2bk) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e89e55b4036a3e78946abf69276f81d2f6e67e61/internal/provider/api_oidc_config.go#L67)):*

I wasn't able to find anything on the map syntax here through a google search. Do values for this field follow some sort of standard OIDC syntax? If so, it'd be helpful to link to a doc or something. If this is our proprietary syntax, it would definitely be good to break this out into an object with all the map components in a separate field.
</blockquote></details>

It's https://www.postgresql.org/docs/current/auth-username-maps.html without the first column. We unfortunately, don't document our use of this well from CRDB which is why I didn't link to any docs yet.
___
*[`internal/provider/api_oidc_config.go` line 150 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-Nc8okGWDdGvv8a0R5ez:-Nc8yuoPBPM9FUN4cguj:b-bqivcu) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e89e55b4036a3e78946abf69276f81d2f6e67e61/internal/provider/api_oidc_config.go#L150)):*

s/Maintenance window/API OIDC Config
</blockquote></details>

Will fix.

jenngeorge (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153)* status: 5 of 17 files reviewed, 9 unresolved discussions (waiting on @erademacher and @kpatron-cockroachlabs)

___
*[`examples/resources/cockroach_api_oidc_config/cockroach_api_oidc_config.tf` line 4 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-NcUS8fPA5HHOWwNJr57:-NcUS8fPA5HHOWwNJr58:b-c91w78) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/08d862054a2188be7c8c1831a65d2de17fe4ddc6/examples/resources/cockroach_api_oidc_config/cockroach_api_oidc_config.tf#L4)):*
> ```HCL
>   issuer       = "https://accounts.google.com"
>   audience     = "test_audience"
>   jwks         = "{\"keys\":[{\"alg\":\"RS256\",\"e\":\"AQAB\",\"kid\":\"test_kid1\",\"kty\":\"RSA\",\"n\":\"09lq1lCEuteonwDJOhGTDak11ThplZuC9JEWQNdBnBSQwlkJQIE7A7nTBO0xTibcsh2HwYkC-N_Gs1jP4iwN3dRqnu5FwG2ct5mY8KLwJiHzToFC0MKenSFQCy0FviNtOnpiObcUlDvR2NDeNtMl_6SPzcQEt7GUTBBYZgoAxPmOgevki6ZNO6Y86xFqx3y6v8EPwW010AiC60r4AHGCTBhYF4uqmq5JH2UU4dDh9Udc-9LZxlSqPwJvnKDG2GjcnD8TsU3wjfEM_nRmx3dnXsrZUXYfNGtdv5dlHywf5AhkJmTavqcsJkgrNA-PNBghFMcCR816_kCIkCYWLWC5vQ\"}]}"
> ```

I prefer using [jsonencode](https://developer.hashicorp.com/terraform/language/functions/jsonencode) for JSON string values because (imo) it's easier to type and read.
Here is an example from the aws provider: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy.html 
___
*[`internal/provider/api_oidc_config.go` line 64 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-NcUMQlh0wJxZOAd_K2x:-NcUMQlh0wJxZOAd_K2y:bhnnh35) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/08d862054a2188be7c8c1831a65d2de17fe4ddc6/internal/provider/api_oidc_config.go#L64)):*
> ```Go
> 				Optional:    true,
> 				Computed:    true,
> 				Description: "The JWT claim that should be used as the user identifier. Defaults to the subject if an empty string is provided.",
> ```

nit: What do you think about updating the second sentence of the description to just "Defaults to the subject."?
___
*[`internal/provider/api_oidc_config_test.go` line 36 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-NcUO3aX-bJ_zTn6DqO0:-NcUO3aX-bJ_zTn6DqO1:bikdpkr) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/08d862054a2188be7c8c1831a65d2de17fe4ddc6/internal/provider/api_oidc_config_test.go#L36)):*
> ```Go
> // a real API OIDC Config. It will be skipped if TF_ACC isn't set.
> // In order to work the ApiOidcEnabled Feature Flag must be enabled and
> // the test org must have Org SSO enabled (no need for SAML/OIDC).
> ```

@erademacher is it ok for us to turn on SSO and that feature flag for the test org?
___
*[`internal/provider/api_oidc_config_test.go` line 50 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-NcURCJbE-8bLcKizYje:-NcURCJbE-8bLcKizYjf:b-fmmhyb) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/08d862054a2188be7c8c1831a65d2de17fe4ddc6/internal/provider/api_oidc_config_test.go#L50)):*
> ```Go
> // TestIntegrationApiOIdcConfigResource attempts to create, check, and destroy
> // an API OIDC Config, but uses a mocked API service.
> func TestIntegrationApiOIdcConfigResource(t *testing.T) {
> ```

supernit: The `I` in `OIdc` is uppercase for `TestAccApiOIdcConfigResource` & `TestIntegrationApiOIdcConfigResource` but lowercase elsewhere.

kpatron-cockroachlabs (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153)* status: 5 of 17 files reviewed, 8 unresolved discussions (waiting on @erademacher and @jenngeorge)

___
*[`examples/resources/cockroach_api_oidc_config/cockroach_api_oidc_config.tf` line 4 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-NcUS8fPA5HHOWwNJr57:-NcXVSNs-S1Rz7dGWUZZ:bdpe7qt) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/08d862054a2188be7c8c1831a65d2de17fe4ddc6/examples/resources/cockroach_api_oidc_config/cockroach_api_oidc_config.tf#L4)):*
<details><summary><i>Previously, jenngeorge (Jenn Georgevich) wrote…</i></summary><blockquote>

I prefer using [jsonencode](https://developer.hashicorp.com/terraform/language/functions/jsonencode) for JSON string values because (imo) it's easier to type and read.
Here is an example from the aws provider: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy.html
</blockquote></details>

It's an interesting tradeoff. I agree that would make it easier to read but the usual method users will have for getting the JWKS is by pulling it from a website like https://www.googleapis.com/oauth2/v3/certs. In those cases it might be easier for the author to just copy and past rather than converting it to TF format.
___
*[`internal/provider/api_oidc_config.go` line 64 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-NcUMQlh0wJxZOAd_K2x:-NcXV_kj-cBCZBcHIV_E:b-896fix) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/08d862054a2188be7c8c1831a65d2de17fe4ddc6/internal/provider/api_oidc_config.go#L64)):*
<details><summary><i>Previously, jenngeorge (Jenn Georgevich) wrote…</i></summary><blockquote>

nit: What do you think about updating the second sentence of the description to just "Defaults to the subject."?
</blockquote></details>

Done.
___
*[`internal/provider/api_oidc_config_test.go` line 50 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-NcURCJbE-8bLcKizYje:-NcXVheF4pgnadNy904S:b6zrrkh) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/08d862054a2188be7c8c1831a65d2de17fe4ddc6/internal/provider/api_oidc_config_test.go#L50)):*
<details><summary><i>Previously, jenngeorge (Jenn Georgevich) wrote…</i></summary><blockquote>

supernit: The `I` in `OIdc` is uppercase for `TestAccApiOIdcConfigResource` & `TestIntegrationApiOIdcConfigResource` but lowercase elsewhere.
</blockquote></details>

Good catch.

erademacher (changes requested):
Sorry for the delay. Looks great so far. Almost there. Just need to make a final decision on the structure of `identity_map` and rerun codegen.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-:-NccEEeRA28zzt9svFg-:bcacrya)* status: 5 of 17 files reviewed, 5 unresolved discussions (waiting on @jenngeorge and @kpatron-cockroachlabs)

___
*[`internal/provider/api_oidc_config.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-Nc8o2ylAJJwJupyLhBX:-NccCa2D8AwHkXSu7rvr:b8s920v) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e89e55b4036a3e78946abf69276f81d2f6e67e61/internal/provider/api_oidc_config.go#L47)):*
<details><summary><i>Previously, kpatron-cockroachlabs (Kyle) wrote…</i></summary><blockquote>

This is a UUID the backend creates. I just copied the wording from somewhere else. Happy to delete that bit.
</blockquote></details>

Thanks. Don't forget to rerun `make generate`.
___
*[`internal/provider/api_oidc_config.go` line 67 at r1](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-Nc8mqYc7rOXJTXFsJkw:-NccCn4QE__HufTLU-LI:b-sozawy) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/e89e55b4036a3e78946abf69276f81d2f6e67e61/internal/provider/api_oidc_config.go#L67)):*
<details><summary><i>Previously, kpatron-cockroachlabs (Kyle) wrote…</i></summary><blockquote>

It's https://www.postgresql.org/docs/current/auth-username-maps.html without the first column. We unfortunately, don't document our use of this well from CRDB which is why I didn't link to any docs yet.
</blockquote></details>

I guess my real question is "how do we make this as easy to use as possible?" Is this likely something users are going to pass in as a reference to a field from another resource? Or paste it in from an external source? If so, string is great. If it's something the user is probably going to need to populate manually, or get all three components from separate resources, then I think you should separate this out into its own object with `map_name`, `system_username`, and `database_username` fields.

If you still feel like a string is the best way to go, please add a link to that Postgres doc in the `MarkdownDescription` field.
___
*[`internal/provider/api_oidc_config_test.go` line 36 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-NcUO3aX-bJ_zTn6DqO0:-NccDrxdCoYTvkhL_UMd:b-a7dcwa) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/08d862054a2188be7c8c1831a65d2de17fe4ddc6/internal/provider/api_oidc_config_test.go#L36)):*
<details><summary><i>Previously, jenngeorge (Jenn Georgevich) wrote…</i></summary><blockquote>

@erademacher is it ok for us to turn on SSO and that feature flag for the test org?
</blockquote></details>

Yeah, go for it. This org is only used for these tests, so we have free reign over feature flags.

jenngeorge (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153)* status: 5 of 17 files reviewed, 4 unresolved discussions (waiting on @erademacher and @kpatron-cockroachlabs)

___
*[`examples/resources/cockroach_api_oidc_config/cockroach_api_oidc_config.tf` line 4 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-NcUS8fPA5HHOWwNJr57:-NccNa1g9Gs6PxZcsrJL:bbc7imm) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/08d862054a2188be7c8c1831a65d2de17fe4ddc6/examples/resources/cockroach_api_oidc_config/cockroach_api_oidc_config.tf#L4)):*
<details><summary><i>Previously, kpatron-cockroachlabs (Kyle) wrote…</i></summary><blockquote>

It's an interesting tradeoff. I agree that would make it easier to read but the usual method users will have for getting the JWKS is by pulling it from a website like https://www.googleapis.com/oauth2/v3/certs. In those cases it might be easier for the author to just copy and past rather than converting it to TF format.
</blockquote></details>

Ok thanks, the copy&pasted string sgtm!
___
*[`internal/provider/api_oidc_config_test.go` line 36 at r2](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-NcUO3aX-bJ_zTn6DqO0:-NccQmoXAFXojVG71B7L:buhlh2w) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/08d862054a2188be7c8c1831a65d2de17fe4ddc6/internal/provider/api_oidc_config_test.go#L36)):*

Yeah, go for it. This org is only used for these tests, so we have free reign over feature flags.
</blockquote></details>

Thanks, done! Org SSO is enabled (org members received a re-invite email) and ApiOidcEnabled is enabled.

jenngeorge (approved):
Thanks for the `identity_map` structure change! :lgtm: after the one nit.

> Reviewed 8 of 18 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-:-Ndk1pjdCncs_aKEaNwK:bix2ac7)* status: 8 of 19 files reviewed, 3 unresolved discussions (waiting on @erademacher and @kpatron-cockroachlabs)

___
*[`internal/provider/api_oidc_config.go` line 71 at r4](https://reviewable.io/reviews/cockroachdb/terraform-provider-cockroach/153#-Ndk-t-JBH6TuUNspRc8:-Ndk-t-JBH6TuUNspRc9:bs1i85j) ([raw file](https://github.com/cockroachdb/terraform-provider-cockroach/blob/1f06bf52e7598d2c87bfceff0437674af5f075fc/internal/provider/api_oidc_config.go#L71)):*
> ```Go
> 						"token_identity": schema.StringAttribute{
> 							Required:    true,
> 							Description: "The token value that needs to be mapped",
> ```

nit: Period at end of description (then rerun `make generate` to update api_oidc_config.md ).

Issue #152, opened 2023-08-08
Error when scaling down cluster


uLan08:
Trying to scale down a cluster from a 4vCPU node to 2vCPU node but encountered an inconsistent apply error.

Terraform plan
```shell
  # module.crdb_foo.cockroach_cluster.cluster will be updated in-place
  ~ resource "cockroach_cluster" "cluster" {
      ~ dedicated         = {
          ~ disk_iops                  = 1500 -> (known after apply)
          ~ machine_type               = "m6i.xlarge" -> (known after apply)
          ~ memory_gib                 = 16 -> (known after apply)
          ~ num_virtual_cpus           = 4 -> 2
            # (2 unchanged attributes hidden)
        }
        id                = "<REDACTED>"
        name              = "foo-cluster"
      ~ operation_status  = "UNSPECIFIED" -> (known after apply)
      ~ regions           = [
          ~ {
                name       = "<REDACTED>"
              ~ sql_dns    = "<REDACTED>" -> (known after apply)
              ~ ui_dns     = "<REDACTED>" -> (known after apply)
                # (2 unchanged attributes hidden)
            },
        ]
      ~ state             = "CREATED" -> (known after apply)
      ~ upgrade_status    = "FINALIZED" -> (known after apply)
        # (5 unchanged attributes hidden)
    }
```

Error
```shell
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to module.crdb_foo.cockroach_cluster.cluster, provider "provider[\"registry.terraform.io/cockroachdb/cockroach\"]" produced an unexpected new value: .dedicated.num_virtual_cpus: was cty.NumberIntVal(2), but now cty.NumberIntVal(4).
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

Versions
```
▶ terraform version
Terraform v1.3.7
on darwin_amd64
+ provider registry.terraform.io/cockroachdb/cockroach v0.4.7
```

Had to manually apply the changes via the CockroachDB console.

erademacher:
Hi Patrick. Thanks for reporting this! I'm able to reproduce locally and should be able to get a fix into the next release.

erademacher:
This should be fixed in v1.0.0. Please let me know if the issue persists.

PR #154, opened 2023-08-08
Check both IsNull and IsUnknown in plan

erademacher:
Previously, we used the config instead of the plan for create and update operations and were able to simply check for null values. Now we use the plan, but that means we need to check if values are either null or unknown.

This change adds a utility method that checks both. Replaced all references to both states, since in all existing cases, we want to treat them the same.

Refactored the cluster resource integration tests to allow any number of GetCluster calls and return the latest iteration of the cluster after UpdateCluster mutates it.

This fixes issue #152.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/internal/provider/cluster_resource_test.go
+++ b/internal/provider/cluster_resource_test.go
@@ -618,13 +618,19 @@ func TestIntegrationDedicatedClusterResource(t *testing.T) {
 	finalizedCluster := upgradingCluster
 	finalizedCluster.UpgradeStatus = client.CLUSTERUPGRADESTATUSTYPE_FINALIZED
 
+	scaledCluster := finalizedCluster
+	scaledCluster.Config.Dedicated = &client.DedicatedHardwareConfig{}
+	*scaledCluster.Config.Dedicated = *finalizedCluster.Config.Dedicated
+	scaledCluster.Config.Dedicated.NumVirtualCpus = 4
+
+	httpOk := &http.Response{Status: http.StatusText(http.StatusOK)}
+
 	// Creation
 
 	s.EXPECT().CreateCluster(gomock.Any(), gomock.Any()).
 		Return(&cluster, nil, nil)
 	s.EXPECT().GetCluster(gomock.Any(), clusterID).
-		Return(&cluster, &http.Response{Status: http.StatusText(http.StatusOK)}, nil).
-		Times(8)
+		Return(&cluster, httpOk, nil).AnyTimes()
 
 	// Upgrade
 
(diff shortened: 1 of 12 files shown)


internal/provider/utils.go
```diff
@@ -134,3 +134,13 @@ func formatEnumMarkdownList[T ~string](allowedValues []T) (mdList string) {
...
 	return mdList
 }
+
+type Knowable interface {
+	IsUnknown() bool
+	IsNull() bool
+}
+
+// IsKnown is a shortcut that checks in a value is neither null nor unknown.
+func IsKnown[T Knowable](t T) bool {
```
sjbarag, line 144: nit: I'm curious if this needs the generic `[T Knowable]` or if `func IsKnown(v Knowable) bool` would suffice 🤷

internal/provider/utils.go
```diff
@@ -134,3 +134,13 @@ func formatEnumMarkdownList[T ~string](allowedValues []T) (mdList string) {
 	}
 	return mdList
 }
+
+type Knowable interface {
+	IsUnknown() bool
+	IsNull() bool
+}
+
+// IsKnown is a shortcut that checks in a value is neither null nor unknown.
```
sjbarag, line 143: super duper typo nit:
```suggestion
// IsKnown is a shortcut that checks if a value is neither null nor unknown.
```

internal/provider/cluster_resource_test.go
```diff
@@ -618,13 +618,19 @@ func TestIntegrationDedicatedClusterResource(t *testing.T) {
 	finalizedCluster := upgradingCluster
 	finalizedCluster.UpgradeStatus = client.CLUSTERUPGRADESTATUSTYPE_FINALIZED
 
+	scaledCluster := finalizedCluster
+	scaledCluster.Config.Dedicated = &client.DedicatedHardwareConfig{}
+	*scaledCluster.Config.Dedicated = *finalizedCluster.Config.Dedicated
```
sjbarag, line 623: Something about these three lines feels funky, but I think I'm just having a mental block. If you get a moment, I'd really appreciate a small comment explaining what this does :)

internal/provider/cluster_resource_test.go
```diff
@@ -639,66 +645,93 @@ func TestIntegrationDedicatedClusterResource(t *testing.T) {
...
-	s.EXPECT().GetCluster(gomock.Any(), clusterID).
-		Return(&pendingCluster, &http.Response{Status: http.StatusText(http.StatusOK)}, nil)
+		DoAndReturn(
+			func(context.Context, string, *client.UpdateClusterSpecification,
+			) (*client.Cluster, *http.Response, error) {
+				cluster = upgradingCluster
+				return &cluster, httpOk, nil
+			},
+		)
+
```
sjbarag, line 655: Ooooh very clean!

internal/provider/cluster_resource_test.go
```diff
@@ -639,66 +645,93 @@ func TestIntegrationDedicatedClusterResource(t *testing.T) {
 		},
 	}, nil, nil)
 	s.EXPECT().UpdateCluster(gomock.Any(), clusterID, &client.UpdateClusterSpecification{UpgradeStatus: &upgradingCluster.UpgradeStatus}).
-		Return(&upgradingCluster, &http.Response{Status: http.StatusText(http.StatusOK)}, nil)
-	s.EXPECT().GetCluster(gomock.Any(), clusterID).
-		Return(&pendingCluster, &http.Response{Status: http.StatusText(http.StatusOK)}, nil)
+		DoAndReturn(
+			func(context.Context, string, *client.UpdateClusterSpecification,
+			) (*client.Cluster, *http.Response, error) {
```
sjbarag, line 650: looks like gofmt is being a silly goose

sjbarag (approved):
Looks good! Thanks for the quick fix on this, @erademacher 😄 

I've got a pair of pretty small questions, but nothing's blocking IMO!

internal/provider/utils.go
erademacher: Nope, that's not necessary. I think I was originally planning on using a concrete base type but switched to an interface and forgot to adapt.

internal/provider/utils.go
erademacher: Fixed!

internal/provider/cluster_resource_test.go
erademacher: Removed this whole section in favor of updating `cluster` directly. But in case you're curious, when copying `finalizedCluster`, it copied `Config` by reference, so I had to create a new config and copy the value over to avoid mutating `finalizedCluster`.

internal/provider/cluster_resource_test.go
erademacher: Fixed

erademacher:
TFTR!

PR #155, opened 2023-08-08
Fix TestAccMultiRegionServerlessClusterResource

erademacher:
Previously, the primary_region field used a `UseStateForUnknown` plan modifier, but that wasn't actually doing anything because we were checking the config instead of the plan. Now that we're checking the plan instead, the modifier is retaining the attribute's previous value when unknown (removed from the config). That was causing the previous primary region to still be marked as primary, and there was no validation to ensure that only one primary region is specified.

This change removes the plan modifier and adds validation to protect against multiple primaries.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)


pjtatlow (approved):
LGTM!

erademacher:
TFTR!

PR #158, opened 2023-08-16
[CC-25438] Update `LogExport` to include `omitted_channels`

Santamaura:
This change updates `LogExport` to include `omitted_channels`

### DIFF
--- a/internal/provider/log_export_config_resource.go
+++ b/internal/provider/log_export_config_resource.go
@@ -106,6 +106,11 @@ var logExportAttributes = map[string]schema.Attribute{
 		Computed:    true,
 		Description: "Indicates when the log export configuration was last updated.",
 	},
+	"omitted_channels": schema.ListAttribute{
+		Optional:    true,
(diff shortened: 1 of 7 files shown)


internal/provider/log_export_config_resource.go
```diff
@@ -320,6 +325,13 @@ func loadLogExportIntoTerraformState(
 		apiRegion = types.StringValue(spec.GetRegion())
 	}
 
+	var omittedChannels []types.String
+	if len(spec.GetOmittedChannels()) != 0 {
```
sjbarag, line 329: nit: looks like the `len() != 0` check is unnecessary!

sjbarag (approved):
Looks good to my eyes! I'll defer to @erademacher if he gets time :)

internal/provider/log_export_config_resource.go
Santamaura: You are correct, removed it.

PR #162, opened 2023-08-30
Convert most tests to use GCP instead of AWS

erademacher:
Also moved an AWS test from ap-southeast-1 to us-east-1.


jaylim-crl (approved):

erademacher:
TFTR

PR #164, opened 2023-09-13
.github/workflows: split release into pre-release-ci and release (manual)

jaylim-crl:
Previously, we run the release workflow automatically whenever a v* tag
gets pushed. The existing workflow does not work because the tests can be
flaky. When that happens, there's no way to release a new TF version since
goreleaser will only run once all the acceptance tests pass.

Until the flakiness goes away, we will split the release workflow into
two: pre-release-ci (that runs automatically like before), and release (which
has to be triggered manually after all the acceptance tests are triaged).

**Commit checklist**
- [X] Changelog -- irrelevant
- [X] Doc gen (`make generate`) -- irrelevant
- [X] Integration test(s) -- irrelevant
- [X] Acceptance test(s) -- irrelevant
- [X] Example(s) -- irrelevant

### DIFF
--- a/.github/workflows/pre-release-ci.yaml
+++ b/.github/workflows/pre-release-ci.yaml
@@ -1,0 +1,25 @@
+name: pre-release-ci
+on:
+  push:
+    tags:
+      - 'v*'
+jobs:
+  acceptance-test:
+    name: acceptance-test
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


erademacher (approved):

jaylim-crl:
TFTR!

PR #166, opened 2023-09-15
Revert "Replace list attribute with sets (#163)"

erademacher:
This reverts commit 551369a73d92e53298fcec2c0922d19c45398b63.


erademacher:
Using sets only partially solves the problem of regions being destroyed and recreated, and it causes other problems as well, namely that Terraform can't figure out how to correctly update regions when the primary is changed.

kpatron-cockroachlabs (approved):
Thank you

erademacher:
TFTR!

PR #167, opened 2023-09-19
Remove cockroach_version from the dedicated example

fantapop:
cockroach_version requires a feature flag to be set even if the version is set to the latest version.  Customers who use this example nearly always run into this issue. Removing this block since its not core to the example.

For now, I've left out any docs changes.

**Commit checklist**
- [ ] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)


erademacher (approved):
Thanks for doing this!

PR #168, opened 2023-10-09
defer to base retry check when response is nil

fantapop:
We use a custom retryable http client to retry GET requests. Since it is intended to only retry GET requests we check that the request was a GET at the beginning of the method.  The only way to access the request at this point in the code is through the response object, however, there are some cases where this  object is nil.  This was resulting in a nil pointer exception.

After this change, any errors that are seen without a set Response object will fall through to the default handler.

Fixes a Nil pointer exception reported here:
https://cockroachlabs.slack.com/archives/C01J7MGJ3S7/p1696613858625329?thread_ts=1696613850.476459&cid=C01J7MGJ3S7

**Commit checklist**
- ✔️  Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed a nil pointer exception which could occur while retrying a certain
+  class of api failures.
+
 ## [1.3.0] - 2023-09-14
 
 ### Added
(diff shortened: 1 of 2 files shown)


jaylim-crl (approved):
LGTM. I wonder how did we end up with a nil object in the first place 🤔

fantapop:
Yeah, I'm not sure.   I looked through the httpclient.Do code and found that there are a few places that can return a nil response.  It could be that this change is just kicking the can up the road but we'll see.

here is one example:
  https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/net/http/client.go;l=674

PR #174, opened 2023-11-30
log-export: make group specific redact noncomputed

arjunmahishi:
Currently, the "redact" field in the individual log groups is declared as a "computed" field. When a field is marked as "computed", user removing the field/commenting it out is sometimes not recognised as a "diff" in terraform state. This makes it look like the "redact" field is not working

**Commit checklist**
- [x] Changelog
- [x] Integration test(s) -- Tested manually on staging
- [ ] ~Acceptance test(s)~ No new test was added
- [ ] ~Example(s)~ No change in usage. This is just a bugfix.

---

### Manual test

Applied this configuration of the `log_exporter`
Screenshot 2023-11-30 at 4 26 00 PM

#### Before

Does not show a diff when `redact` is commented out in the second log group
Screenshot 2023-11-30 at 4 29 36 PM

#### After the change

Shows the expected diff
Screenshot 2023-11-30 at 4 31 03 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed a nil pointer exception which could occur while retrying a certain
   class of api failures.
+- Fixed a bug where the `log_export_config` would sometimes fail to detect changes to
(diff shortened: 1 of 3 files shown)


fantapop (approved):
This seems reasonable to me.

PR #178, opened 2024-02-14
fix error when primary attribute is specified in multiple regions

andy-kimball:
Previously, the TF provider raised an error when the primary attribute was specified in multiple regions, e.g.:

  regions = [
    {
      name    = "eu-west-1"
      primary = true
    },
    {
      name    = "eu-west2"
      primary = false
    },
    {
      name    = "ap-south-1"
      primary = false
    },
  ]

This PR fixes that by adding some custom validation logic rather than using the built-in config validators.

Fixes: CC-27160

**Commit checklist**
- [X] Changelog
- [X] Integration test(s)
- [X] Acceptance test(s)
- [X] Example(s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed a nil pointer exception which could occur while retrying a certain
   class of api failures.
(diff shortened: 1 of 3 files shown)


rgcase (approved):

andy-kimball:
TFTR

PR #181, opened 2024-03-03
Include examples in the generated docs

fantapop:
Previously, examples were not included in the documentation.  This was because the naming scheme for the example files did not match was was expected by the tf provider framework.  The files are now renamed to the correct convention and the docs regenerated.

**Commit checklist**
- [x] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)


rgcase (approved):
Nice, good find.

PR #183, opened 2024-03-10
Use new private endpoint connections API

carloruiz:
Previously, the private_endpoint_connection resource only supported AWS
    private link connections. This commit updates the provider code to use
    the new CC API Private Endpoint Connections methods. These cloud-neutral
    methods will work for all cloud providers and cluster types, except
    Serverless clusters on Azure as that configuration is not yet
    supported.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/internal/provider/private_endpoint_connection_resource.go
+++ b/internal/provider/private_endpoint_connection_resource.go
@@ -48,7 +48,7 @@ func (r *privateEndpointConnectionResource) Schema(
 	_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse,
 ) {
 	resp.Schema = schema.Schema{
-		MarkdownDescription: "AWS PrivateLink Endpoint Connection.",
+		MarkdownDescription: "Private Endpoint Connection.",
 		Attributes: map[string]schema.Attribute{
 			"id": schema.StringAttribute{
 				Computed:            true,
@@ -76,14 +76,14 @@ func (r *privateEndpointConnectionResource) Schema(
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplace(),
 				},
-				Description: "Client side ID of the PrivateLink connection.",
+				Description: "Client side ID of the Private Endpoint Connection.",
 			},
 			"service_id": schema.StringAttribute{
 				Computed: true,
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.UseStateForUnknown(),
 				},
-				Description: "Server side ID of the PrivateLink connection.",
+				Description: "Server side ID of the Private Endpoint Connection.",
 			},
 			"cluster_id": schema.StringAttribute{
 				Required: true,
@@ -129,7 +129,8 @@ func (r *privateEndpointConnectionResource) Create(
 		return
 	}
 
-	cluster, _, err := r.provider.service.GetCluster(ctx, plan.ClusterID.ValueString())
+	svc := r.provider.service
+	cluster, _, err := svc.GetCluster(ctx, plan.ClusterID.ValueString())
 	if err != nil {
 		resp.Diagnostics.AddError(
 			"Error getting cluster",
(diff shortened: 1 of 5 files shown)


fantapop (commented):
These changes look good to me.  I'm slightly confused about the state of the example here:
https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/examples/resources/cockroach_aws_private_endpoint_connection/cockroach_aws_private_endpoint_connection.tf

I actually have an [outstanding PR](https://github.com/cockroachdb/terraform-provider-cockroach/pull/181) to rename it anyways but perhaps the contents should be update as part of this to reflect the correct usage.   

Along similar lines, I'm wondering if there is any necessary migration as part of this or period where we need to keep a deprecated resource name around?

carloruiz:
> perhaps the contents should be update as part of this to reflect the correct usage.

ack. will update

> Along similar lines, I'm wondering if there is any necessary migration as part of this or period where we need to keep a deprecated resource name around?

Fortunately, we're not deprecating any TF fields as part of this change.

carloruiz:
Acceptance test: had do manually change some code to make this work. There's no straight-forward way to automatically test Private endpoint connection creation bc the client needs to create the endpoint out of band before calling the API. I created the endpoint manually and hardcoded the endpoint id into the test.
```
COCKROACH_API_KEY=<KEY> COCKROACH_SERVER=https://management-staging.crdb.io TF_ACC=1 make test
go test ./... -run TestAccServerlessPrivateEndpointConnectionResource -v  -timeout 20m
?   	github.com/cockroachdb/terraform-provider-cockroach	[no test files]
?   	github.com/cockroachdb/terraform-provider-cockroach/mock	[no test files]
=== RUN   TestAccServerlessPrivateEndpointConnectionResource
=== PAUSE TestAccServerlessPrivateEndpointConnectionResource
=== CONT  TestAccServerlessPrivateEndpointConnectionResource
--- PASS: TestAccServerlessPrivateEndpointConnectionResource (68.70s)
PASS
ok  	github.com/cockroachdb/terraform-provider-cockroach/internal/provider	69.241s
```

carloruiz:
Actually I was able to add the test as a fixture so it runs with the acceptance tests.

internal/provider/private_endpoint_connection_resource.go
```diff
@@ -299,10 +290,11 @@ func waitForEndpointConnectionCreatedFunc(
...
 			if connection.GetEndpointId() == endpointID {
 				switch status := connection.GetStatus(); status {
-				case client.AWSENDPOINTCONNECTIONSTATUSTYPE_AVAILABLE:
+				case client.PRIVATEENDPOINTCONNECTIONSTATUS_AVAILABLE:
 					return nil
-				case client.AWSENDPOINTCONNECTIONSTATUSTYPE_PENDING,
-					client.AWSENDPOINTCONNECTIONSTATUSTYPE_PENDING_ACCEPTANCE:
+				case client.PRIVATEENDPOINTCONNECTIONSTATUS_PENDING,
+					client.PRIVATEENDPOINTCONNECTIONSTATUS_PENDING_ACCEPTANCE,
+					client.PRIVATEENDPOINTCONNECTIONSTATUS_REJECTED:
```
fantapop, line 297: the include of `REJECTED` inside the retryable error case seems wrong.  If this is actually the right thing, can you add a comment here explaining why?

internal/provider/private_endpoint_connection_resource_test.go
```diff
@@ -38,17 +38,18 @@ func TestAccDedicatedPrivateEndpointConnectionResource(t *testing.T) {
 }
 
 func TestAccServerlessPrivateEndpointConnectionResource(t *testing.T) {
-	t.Skip("Skipping until we can either integrate the AWS provider " +
-		"or import a permanent test fixture.")
+	// This test relies on a pre-created private endpoint created via the AWS
```
fantapop, line 41: This feels a little fragile to me.  Do we have anything protecting it from being removed?  Many teams won't have the perms to fix it if it breaks.

internal/provider/private_endpoint_connection_resource_test.go
```diff
@@ -238,7 +231,7 @@ resource "cockroach_private_endpoint_services" "services" {
 
 resource "cockroach_private_endpoint_connection" "connection" {
     cluster_id = cockroach_cluster.dedicated.id
-    endpoint_id = "endpoint-id"
+    endpoint_id = "vpce-0011573c1fa6afa3d"
```
fantapop, line 234: rather than hard coding this id everywhere, is it possible to use a datasource to fetch the endpoint using something else like name?

fantapop (commented):
Hey, I had a few concerns.  Left some comments in there.

internal/provider/private_endpoint_connection_resource_test.go
carloruiz: >  is it possible to use a datasource to fetch the endpoint using something else like name?

To fetch from AWS? We'd need AWS creds and we'd need to figure out a way to share the AWS API across CC eng. Also, the endpoint is only uniquely identified by ID, so we'd still need to store the endpointId somewhere.

I can use a const if that makes it cleaner.

internal/provider/private_endpoint_connection_resource_test.go
carloruiz: I can't think of a way to make it "unfragile", except actually creating an AWS service account, an API key for the account, adding AWS specific logic to create and delete the private endpoint connection on each test run, running the tests with the AWS API key. We'd also need to figure out a way to store AWS API Key so all engineers have access? Maybe vault, but then the TF tests would have a vault dep. I don't think it's a viable approach atm, too much hassle.

I know it's janky but I think the approach in the PR could get us working acceptance tests for a year. It will break eventually and it be annoying to fix but IMO it's worth having the coverage.

internal/provider/private_endpoint_connection_resource.go
carloruiz: Yea seems odd but it's actually correct. Will add a comment.

internal/provider/private_endpoint_connection_resource_test.go
carloruiz: Hmm actually I'm going to leave this test as skipped since the endpointId only works for staging. Any acceptance test ran against any other env will fail.

fantapop: Is there a way to make it conditional on the env?  Or maybe we can add instructions for running the test in staging?

fantapop (approved):
LGTM

PR #185, opened 2024-03-13
migrate testing framework to terraform-plugin-testing

fantapop:
Previously, the terraform provider utilized the sdkv2 framework. In order to stay up to date with testing framework improvements we upgrade to the new terraform-plugin-testing framework following these instructions:

https://developer.hashicorp.com/terraform/plugin/testing/migrating

The change that instigated this upgrade is that the id field is no longer required in the new framework.  During development of the role_grant_resource an ID was not used because the ccapi does not provide an ID for it.  Rather than work around this using a synthetic ID, we upgrade.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)


fantapop:
@carloruiz / @andy-kimball  as recent contributors to the terraform provider, I've added you two here as reviewers.

fantapop:
The changes are separated into 2 commits, one with all the vendor changes.   Don't let the 1600+ file counts scare you since its mostly vendor.

andy-kimball (approved):
LGTM

PR #188, opened 2024-03-14
[CC-27365] add user_role_grant resource

fantapop:
Previously, the provider supported user role management but only in an authoritative way.  Using the user_role_grants resource would overwrite all roles for a user.  This caused churn for terraform users trying to manage user roles across multiple projects or via the console UI.  We now add a resource, user_role_grant which allows management of a single user role grant.  The existing user_role_grants resource is left to maintain the previous functionality for those that need it.

As part of this changes, I've also renamed the role_resource files which contain the user_role_grants resource after the full name of the role since that seems to be the convention.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)


internal/provider/models.go
```diff
@@ -268,12 +268,17 @@ type Role struct {
...
 
-type RoleGrant struct {
+type UserRoleGrants struct {
 	ID     types.String `tfsdk:"id"`
 	UserId types.String `tfsdk:"user_id"`
 	Roles  []Role       `tfsdk:"roles"`
 }
 
+type UserRoleGrant struct {
+	UserID types.String `tfsdk:"user_id"`
```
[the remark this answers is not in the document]
fantapop: I experimented with this and it doesn't look like we need it.  I did eventually run into an issue with the testing framework where we needed an ID but this was avoided by migrating to the new framework.   See here:
https://developer.hashicorp.com/terraform/plugin/framework/acctests#no-id-found-in-attributes

Here is the PR where I migrated the framework.
https://github.com/cockroachdb/terraform-provider-cockroach/pull/185

Also, I mentioned to you early on that we were planning to add a unique ID for this but decided against it because it didn't fit into our current public api definitions at all.

docs/resources/user_role_grant.md
```diff
@@ -0,0 +1,78 @@
...
+### Nested Schema for `role`
+
+Required:
+
+- `resource_type` (String) Type of resource. Allowed values are:
+  * ORGANIZATION
+  * CLUSTER
+  * FOLDER
+- `role_name` (String) Name of the role to grant. Allowed values are:
+  * DEVELOPER
```
pawalt, line 51: Developer/Admin are not supported

docs/resources/user_role_grant.md
```diff
@@ -0,0 +1,78 @@
...
+  * CLUSTER_ADMIN
+  * CLUSTER_OPERATOR_WRITER
+  * CLUSTER_DEVELOPER
+  * CLUSTER_CREATOR
+  * FOLDER_ADMIN
+  * FOLDER_MOVER
+
+Optional:
+
+- `resource_id` (String) ID of the resource. Omit if resource_type is 'ORGANIZATION'.
```
pawalt, line 65: Required if resource is cluster/folder is probably useful info.

internal/provider/user_role_grant_resource.go
```diff
@@ -0,0 +1,272 @@
...
+
+	_, err := fromRoleToBuiltInRole(plan.Role)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Invalid role format",
+			fmt.Sprintf("Could not parse role: %s", formatAPIErrorMessage(err)),
+		)
+		return
+	}
+	_, _, err = r.provider.service.AddUserToRole(
```
pawalt, line 110: Do we need to do any logic here to check if this assignment is necessary? Or should we just always do it because it's idempotent?

internal/provider/user_role_grant_resource.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+const resourceTypeOrg = string(client.RESOURCETYPETYPE_ORGANIZATION)
+
+func parseRoleGrantResourceID(
+	id string,
+) (userID, roleName, resourceType, resourceID string, err error) {
+	idParts := strings.Split(id, ",")
+	if len(idParts) == 4 {
+		if idParts[2] == resourceTypeOrg && idParts[3] != "" {
```
pawalt, line 251: Maybe a `uuid.Parse` would be helpful here as a sanity check

docs/resources/user_role_grant.md
fantapop: yes, good point.  Let me know what you think of this:

     ID of the resource. Required if the resource_type is 'FOLDER' or 'CLUSTER'. It should be omitted otherwise.

Additionally, I believe we'll have some docs review before landing this.

internal/provider/user_role_grant_resource.go
fantapop: In general I'm not concerned with idempotency as terraform sprinkles read in various places to determine if something is necessary. Drift can obviously happen and whether it happens during or close to a terraform run is just luck.  It would be detected during the next run and updated.   For this particular case however I can't recall if I've tested what happens when a role grant already exists.  It *should* fail in this case but if our API is idempotent, it may think that it has been able to "adopt" this resource without using import.   I'll do some testing around this.  Maybe I can add an integration test.

docs/resources/user_role_grant.md
fantapop: These are coming through the cloud sdk which gets them from the open api spec which still reflects this. I'll get a PR out right away to remove.  Maybe we can still cherry pick them for this weeks release and I can sneak the change in here.

fantapop: https://github.com/cockroachlabs/managed-service/pull/16711

internal/provider/user_role_grant_resource.go
fantapop: I've updated that method and tests to do checking on the uuids.

docs/resources/user_role_grant.md
pawalt: I like that wording

pawalt (approved):
LGTM pending docs review

internal/provider/user_role_grant_resource.go
fantapop: So I think this is actually a problem.  If terraform tries to create a resource which already exists and doesn't fail, it means two projects that are accidentally using the same resource will both think they own it.

fantapop: It could also lead to issues in the same project where a project thinks it has 2 managed resources but only one on the backend.

marksoper:
@mikeCRL @codingconcepts @fantapop @pawalt @shannonbradshaw 
Hi Mike, Rob-
I got word from @shannonbradshaw that you may be able to help with docs for this. A good first step would be to have a look at this PR to see if there are concerns and understand the docs implications.
As you probably know docs are critical for Terraform, as so much Terraform usage begins with `examples`

internal/provider/user_role_grant_resource.go
fantapop: I pushed an update to check for the resource before attempting to "Create" it.  I've also updated the tests to exercise a failure.

pawalt (approved):
Idempotency changes LGTM

fantapop:
I'm waiting on the new SDK tag to be available before merging this down.  We can add documentation revisions to a later PR if any manifest.  Just let me know.

PR #192, opened 2024-03-21
Use GCP in private_endpoint_services_resource test for dedicated

fantapop:
There are less limitations on GCP hosts so we prefer to use those for spinning up dedicated tests where possible.  This updates the the dedicated test in the private_endpoint_services_resource test to use GCP.

**Commit checklist**
- [ ] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/internal/provider/private_endpoint_services_resource_test.go
+++ b/internal/provider/private_endpoint_services_resource_test.go
@@ -35,7 +35,7 @@ import (
 // skipped if TF_ACC isn't set.
 func TestAccDedicatedPrivateEndpointServicesResource(t *testing.T) {
 	t.Parallel()
-	clusterName := fmt.Sprintf("endpoint-services-%s", GenerateRandomString(2))
+	clusterName := fmt.Sprintf("tftest-endpt-svc-%s", GenerateRandomString(2))
 	testPrivateEndpointServicesResource(t, clusterName, false /* useMock */, false /* isServerless */)
 }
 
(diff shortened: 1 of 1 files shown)


internal/provider/private_endpoint_services_resource_test.go
```diff
@@ -68,18 +68,18 @@ func TestIntegrationPrivateEndpointServicesResource(t *testing.T) {
 			client.Cluster{
 				Name:          clusterName,
 				Id:            uuid.Nil.String(),
-				CloudProvider: "AWS",
+				CloudProvider: "GCP",
 				Config: client.ClusterConfig{
 					Dedicated: &client.DedicatedHardwareConfig{
 						StorageGib:     15,
-						MachineType:    "m5.large",
+						MachineType:    "not verified",
```
carloruiz, line 75: I'm not sure about this value. You could set it to `n2-standard-2`. Maybe leaving it unset also works.

carloruiz (approved):
Looks good to me

internal/provider/private_endpoint_services_resource_test.go
fantapop: I updated it because I noticed that all the values used for GCP in terraform are using aws names.  We can go through and start validating them at some point if we want.

fantapop:
The tests passed in staging.  I'm going to go ahead and merge this down and tag for the release.

PR #195, opened 2024-03-22
Fantapop improve examples

fantapop:
**Commit checklist**
- [x] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [x] Example(s)


examples/resources/cockroach_serverless_cluster/resource.tf
```diff
@@ -1,12 +0,0 @@
-resource "cockroach_cluster" "cockroach" {
```
carloruiz, line 1: Why is this removed?

carloruiz (approved):
nice

examples/resources/cockroach_serverless_cluster/resource.tf
fantapop: The cockroach_serverless_cluster and cockroach_dedicated_cluster examples were merged into one.  This is because the example directory name must match the resource in order for it to be included in the docs.  

See here: https://github.com/cockroachdb/terraform-provider-cockroach/pull/195/files#diff-9fbf857764cbe159f98d70fb75ba7f82ede6c04edd00ff216538db4e64ac1d6e

And the generated docs here:
https://github.com/cockroachdb/terraform-provider-cockroach/pull/195/files#diff-ea87985827a5e33692d9699416d9144971b191e9688a542125e8f31b5133adbd

PR #196, opened 2024-03-22
ensure all test cluster names have tftest prefix

fantapop:
Prefixing the cluster names with "tftest" makes them easier to find in case they are failed to be cleaned up. Most were already prefixed that way. This change updates the remaining few that were not.

**Commit checklist**
- [ ] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/internal/provider/allowlist_resource_test.go
+++ b/internal/provider/allowlist_resource_test.go
@@ -37,7 +37,7 @@ import (
 // isn't set.
 func TestAccDedicatedAllowlistEntryResource(t *testing.T) {
 	t.Parallel()
-	clusterName := fmt.Sprintf("tftest-networking-%s", GenerateRandomString(2))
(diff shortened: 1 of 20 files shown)


jaylim-crl (approved):
LGTM. 

One idea is to have a helper test method (e.g. `terraformTestClusterName("priv-conn")`) that all tests could use. That way, we don't need to remember that we need to add a `tftest` prefix when we add a new test.

fantapop:
I was also thinking about something like that. @jaylim-crl are you fine with a `tfTestPrefix` variable instead of a function?  I think its probably a similar amount of characters but easier to see at a glance whats going on.

jaylim-crl:
> I was also thinking about something like that. @jaylim-crl are you fine with a `tfTestPrefix` variable instead of a function? I think its probably a similar amount of characters but easier to see at a glance whats going on.

No strong preferences on that - either approaches are fine.

PR #199, opened 2024-04-11
Allow enabling log export for serverless clusters.

alyshanjahani-crl:


jaylim-crl (approved):
LGTM. 

super nit: if we could add a test, that would be great, but I don't feel strongly. This is what I did back then for IP allowlists: https://github.com/cockroachdb/terraform-provider-cockroach/pull/133/files

PR #200, opened 2024-04-12
[CC-26984] metric-export: add resource for prometheus integration

aa-joshi:
This change introduces resource MetricExportPrometheusConfig which manages prometheus metric export integration in gcp and aws cloud providers.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)


fantapop:
@aa-joshi @marksoper  considering that there is no actual configuration for enabling prometheus what you do think of having it as an additional field on the cluster resource?  It could be something like:

```
resource "cockroach_cluster" "dedicated" {
  name           = "cockroach-dedicated"
  cloud_provider = "GCP"
  prometheus_metric_export_enabled = true
  dedicated = {
  ....
  }
```

If we were planning to add configuration options to this at some point in the near future, it seems like keeping it as its own resource would be desirable but even in that case we could just have that resource just represent the configuration and keep the enabled / disabled flag on cluster.

For example,
```
resource "cockroach_cluster" "dedicated" {
  name           = "cockroach-dedicated"
  cloud_provider = "GCP"
  prometheus_metric_export_enabled = true
  dedicated = {
  ....
  }
resource "cockroach_cluster_prometheus_export_config" "dedicated" {
  cluster_id = cockroach_cluster.dedicated.id
  some_config_option = 1234
}
```

fantapop:
I guess another consideration is whether we'd expose some of the outputs as part of the resource.  I see that the api response has the following data:
```
PrometheusMetricExportInfo
{
    cluster_id*: string
    status: enum  // Allowed: NOT_DEPLOYED┃DISABLING┃ENABLING┃ENABLED┃ERROR
    targets: {   // targets is a map of ports exposing metrics to regions.
        [any-key]: string
    }
    user_message: string
}
```

aa-joshi:
> @aa-joshi @marksoper considering that there is no actual configuration for enabling prometheus what you do think of having it as an additional field on the cluster resource? It could be something like:
> 
> ```
> resource "cockroach_cluster" "dedicated" {
>   name           = "cockroach-dedicated"
>   cloud_provider = "GCP"
>   prometheus_metric_export_enabled = true
>   dedicated = {
>   ....
>   }
> ```
> 
> If we were planning to add configuration options to this at some point in the near future, it seems like keeping it as its own resource would be desirable but even in that case we could just have that resource just represent the configuration and keep the enabled / disabled flag on cluster.
> 
> For example,
> 
> ```
> resource "cockroach_cluster" "dedicated" {
>   name           = "cockroach-dedicated"
>   cloud_provider = "GCP"
>   prometheus_metric_export_enabled = true
>   dedicated = {
>   ....
>   }
> resource "cockroach_cluster_prometheus_export_config" "dedicated" {
>   cluster_id = cockroach_cluster.dedicated.id
>   some_config_option = 1234
> }
> ```

We have existing metric export integration which has their own resources:
- [cockroach_metric_export_datadog_config](https://github.com/cockroachdb/terraform-provider-cockroach/tree/9c26a157180ee4ced1171205692e02df75703e13/examples/resources/cockroach_metric_export_cloudwatch_config)
- [cockroach_metric_export_cloudwatch_config](https://github.com/cockroachdb/terraform-provider-cockroach/tree/9c26a157180ee4ced1171205692e02df75703e13/examples/resources/cockroach_metric_export_cloudwatch_config)

Hence we would like to keep it consistent with other metric export integrations. I will create a seperate backlog item to track integrations against the cluster resource.

aa-joshi:
> I guess another consideration is whether we'd expose some of the outputs as part of the resource. I see that the api response has the following data:
> 
> ```
> PrometheusMetricExportInfo
> {
>     cluster_id*: string
>     status: enum  // Allowed: NOT_DEPLOYED┃DISABLING┃ENABLING┃ENABLED┃ERROR
>     targets: {   // targets is a map of ports exposing metrics to regions.
>         [any-key]: string
>     }
>     user_message: string
> }
> ```

Thanks for suggestion! I have made required changes to keep track of `targets`.

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,390 @@
...
+func (r *metricExportPrometheusConfigResource) Configure(
+	_ context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse,
+) {
+	if req.ProviderData == nil {
+		return
+	}
+	var ok bool
+	if r.provider, ok = req.ProviderData.(*provider); !ok {
+		resp.Diagnostics.AddError("Internal provider error",
+			fmt.Sprintf("Error in Configure: expected %T but got %T", provider{}, req.ProviderData))
```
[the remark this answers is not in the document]
aa-joshi: This is addressed.

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,405 @@
...
+
+var metricExportPrometheusConfigAttributes = map[string]schema.Attribute{
+	"id": schema.StringAttribute{
+		Required:            true,
+		MarkdownDescription: "Cluster ID.",
+		PlanModifiers: []planmodifier.String{
+			stringplanmodifier.RequiresReplace(),
+		},
+	},
+	"enabled": schema.BoolAttribute{
```
aa-joshi, line 27: We have introduced `enabled` field to identify changes against the resources. Since other fields are with `Computed: true`, tf will not be consider as changes unless and until there is change in resource in `terraform plan`.

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,405 @@
...
+		Default:     booldefault.StaticBool(true),
+		Optional:    true,
+		Computed:    true,
```
fantapop, line 30: Is having both of these allowed? If you're not allowed to set the field, I think it should be `Computed: true` only.  Also, I'm a little confused about the description here.  Can we remove the additional bit about refraining to set the value if we just make it Computed?

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,405 @@
...
+		Description: "Encodes the possible states that a metric export configuration can be in as it is created, deployed, and disabled.",
+	},
+	"user_message": schema.StringAttribute{
+		Computed:    true,
+		Description: "Elaborates on the metric export status and hints at how to fix issues that may have occurred during asynchronous operations.",
+	},
+	"targets": schema.MapAttribute{
+		Computed:    true,
+		ElementType: types.StringType,
+		Description: "Represents the endpoint details to scrape metrics.",
```
fantapop, line 44: This description seems a bit vague to me. Actually, looking at the api documentation, I'm still unclear what this is.  In the api we have this:

https://github.com/cockroachlabs/managed-service/blob/41b3b2a651844d2445e2ed914c6ae0ee1e708dff/console/apipb/api.proto#L1604-L1606

>   // targets is a map of ports exposing metrics to regions.
  map<string, string> targets = 4;

So based on the types, it seems like we might have a list of host:port -> region mappings but I'm not sure. @abarganier would you be a good resource to ask about this?  Is there documentation on this somewhere?  Ideally we can improve the api docs here with something clearer.  An example would help a lot.

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,405 @@
...
+	},
+	"enabled": schema.BoolAttribute{
+		Default:     booldefault.StaticBool(true),
+		Optional:    true,
+		Computed:    true,
+		Description: "Represents prometheus integration status. Please refrain setting any value as it is used for state management.",
+	},
+	"status": schema.StringAttribute{
+		Computed:    true,
+		Description: "Encodes the possible states that a metric export configuration can be in as it is created, deployed, and disabled.",
```
fantapop, line 35: We may want to list out the actual states here.   How about this wording instead:

```
Description: "The current state of the metric export configuration.  Possible values are: " + strings.Join(client.AllowedMetricExportStatusTypeEnumValues, ", "),
```

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,405 @@
...
+	},
+	"user_message": schema.StringAttribute{
```
fantapop, line 37: I don't really love the name `user_message` for this.  I don't think we need to change the api but we could choose a better name for the terraform resource field.  Maybe `status_detail` would be a better name here. @abarganier can you help with this one too?

internal/provider/metric_export_prometheus_config_resource.go
fantapop: I was just reading about this.  I see its required because there is a default value.  But I think in this case you should still just use computed with no default since this is synthesized from the status attribute which is also computed.   Or am I not understanding this correctly.  Is this supposed to be a like an on / off switch?

aa-joshi: We have removed this now. We introduced it to compute diff on the resource. tf plan will show diff in fields if and only if there is change in input fields. Since enable prometheus export doesn't have body, we had added this field to track the integrations status.
Scenario:
- create resource (enable prometheus metric export) through tf.
- delete prometheus metric export integration through API.
- create resource (enable prometheus metric export) through tf.
Earlier this was showing no changes. Now, we are removing the resource if integration is deactivated through API.

internal/provider/metric_export_prometheus_config_resource.go
aa-joshi: targets would contain the region -> scrape endpoint mapping like below:
`"targets": {
        "australia-southeast2": "http://localhost:7900/api/v1/clusters/c33bdfbe-b107-4d59-b110-ed435ace547a/metricexport/prometheus/australia-southeast2/scrape"
    }`
I have updated the description to improve readability and it's usage. I hope it is clear now.

internal/provider/metric_export_prometheus_config_resource.go
aa-joshi: This is addressed.

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,405 @@
...
+func waitForPrometheusMetricExportReadyFunc(
+	ctx context.Context,
+	clusterID string,
+	cl client.Service,
+	prometheusMetricExportInfo *client.PrometheusMetricExportInfo,
+) retry.RetryFunc {
+	return func() *retry.RetryError {
+		apiObj, httpResp, err := cl.GetPrometheusMetricExportInfo(ctx, clusterID)
+		if err != nil {
+			if httpResp != nil && httpResp.StatusCode < http.StatusInternalServerError {
```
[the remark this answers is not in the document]
aa-joshi: Thanks for pointing it out! We also need to handle retries based on `Unavailable`. I had mistakenly kept it as `Internal Server Error`.

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,405 @@
...
+
+			if cluster.State == client.CLUSTERSTATETYPE_DELETED {
+				resp.Diagnostics.AddWarning(
+					"Cluster deleted",
+					fmt.Sprintf("The Prometheus metric export config's cluster with ID %s was deleted. Removing from state.",
+						clusterID,
+					))
+				resp.State.RemoveResource(ctx)
+				return
+			}
```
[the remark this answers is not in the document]
aa-joshi: https://github.com/cockroachdb/terraform-provider-cockroach/pull/200#discussion_r1576248029

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,405 @@
...
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	clusterID := state.ID.ValueString()
+	cluster := &client.Cluster{}
+	retryFunc := func() retry.RetryFunc {
+		return func() *retry.RetryError {
+			_, httpResp, err := r.provider.service.DeletePrometheusMetricExport(ctx, clusterID)
```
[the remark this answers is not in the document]
aa-joshi: Thanks for highlighting this! We are now polling till `GetPrometheusMetricExport` gives us stable state.

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,405 @@
...
+						if clusterErr != nil {
+							return retry.NonRetryableError(
+								fmt.Errorf("Error checking cluster availability: %s", clusterErr.Error()))
+						}
+						return retry.RetryableError(fmt.Errorf("cluster was not ready - trying again"))
+					}
+				}
+				return retry.NonRetryableError(
+					fmt.Errorf("could not delete Prometheus metric export config: %v", apiErrMsg))
+			}
```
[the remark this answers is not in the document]
aa-joshi: I have updated error case scenarios. This handles errors which are generated in prepare request in `cockroach-cloud-sdk-go`

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,405 @@
...
+		return
+	}
+
+	clusterID := plan.ID.ValueString()
+	cluster := &client.Cluster{}
+	apiObj := &client.PrometheusMetricExportInfo{}
+	err := retry.RetryContext(ctx, clusterUpdateTimeout, retryEnablePrometheusMetricExport(
+		ctx, r.provider.service, clusterID, cluster,
+		apiObj,
+	))
```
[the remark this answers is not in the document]
aa-joshi: Make sense! I have updated it as no op.

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,405 @@
...
+) {
+	state.ID = types.StringValue(apiObj.GetClusterId())
+	state.Status = types.StringValue(string(apiObj.GetStatus()))
+	state.UserMessage = types.StringValue(apiObj.GetUserMessage())
+	state.Enabled = types.BoolValue(apiObj.GetStatus() == client.METRICEXPORTSTATUSTYPE_ENABLED)
+	state.Targets, _ = types.MapValueFrom(ctx, types.StringType, apiObj.GetTargets())
+
+}
+
+func waitForPrometheusMetricExportReadyFunc(
```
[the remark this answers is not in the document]
aa-joshi: Thanks for suggestion! I have updated the name.

internal/provider/metric_export_prometheus_config_resource.go
abarganier: @aa-joshi has it right - this is returned to tell the user which scrape URLs are exposed per-region. These are the URLs that users plug into their Prometheus scrape configs to start scraping metrics.

arjunmahishi:
Went over the changes in a pair programming session (learnt a lot about terraform lifecycle today!). This LGTM.
Will wait for @fantapop's ✅

internal/provider/metric_export_prometheus_config_resource.go
fantapop: cool.  I'll put up a PR to get this example into the ccapi docs.  I like the new description.  I think the more typical way to access the targets would be to create an Output Value for them. Although I'm unclear on the ramifications of this being a map.  I found this: https://stackoverflow.com/a/55006832

internal/provider/metric_export_prometheus_config_resource.go
fantapop: Nice, thank you!

internal/provider/metric_export_prometheus_config_resource_test.go
```diff
@@ -0,0 +1,213 @@
...
+	"net/http"
+	"os"
+	"testing"
+)
+
+// TestAccMetricExportPrometheusConfigResource attempts to create, check, and destroy
+// a real cluster. It will be skipped if TF_ACC isn't set.
+func TestAccMetricExportPrometheusConfigResource(t *testing.T) {
+	t.Skip("Skipping until we can either integrate the AWS provider " +
+		"or import a permanent test fixture.")
```
fantapop, line 22: Does this rely on an AWS resource?  It seems like it wouldn't....

internal/provider/metric_export_prometheus_config_resource_test.go
```diff
@@ -0,0 +1,213 @@
...
+	t.Parallel()
+	clusterName := fmt.Sprintf("%s-prometheus-%s", tfTestPrefix, GenerateRandomString(4))
```
fantapop, line 24: I've been trying to keep these names short so an org that runs the test doesn't need to enable the feature flag to allow long names.  

the test prefix is 6
-prometheus- is 12
and random is 4

The name limit is 20. 

How about using `"%s-prom-exp-%s"` instead?

internal/provider/metric_export_prometheus_config_resource_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+		return fmt.Errorf("metric export Prometheus config with site %s does not exist", rs.Primary.Attributes["site"])
+	}
+}
+
+func getTestMetricExportPrometheusConfigResourceCreateConfig(name string) string {
+	return fmt.Sprintf(`
+resource "cockroach_cluster" "test" {
+  name           = "%s"
+  cloud_provider = "AWS"
```
fantapop, line 177: Will this work with serverless clusters?   That would be ideal because they're faster to spin up for the test.  If not that's okay.  But we should switch to using GCP so we don't use up an AWS account during the test.

internal/provider/metric_export_prometheus_config_resource_test.go
```diff
@@ -0,0 +1,213 @@
...
+			return fmt.Errorf("not found: %s", clusterResourceName)
+		}
+
+		clusterID := clusterRs.Primary.Attributes["id"]
+		log.Printf("[DEBUG] clusterID: %s, name %s", clusterRs.Primary.Attributes["id"], clusterRs.Primary.Attributes["name"])
+
+		_, _, err := p.service.GetPrometheusMetricExportInfo(context.TODO(), clusterID)
+		if err == nil {
+			return nil
+		}
```
fantapop, line 167: Is this enough to show that its there?  I thought it could return NOT_DEPLOYED or even DEPLOYING, etc...

fantapop (commented):
This is a looking a lot cleaner.  Thanks for your work here.  I added some followup comments and some that I missed earlier.

internal/provider/metric_export_prometheus_config_resource_test.go
```diff
@@ -0,0 +1,213 @@
...
+		if !ok {
+			return fmt.Errorf("not found: %s", resourceName)
+		}
+		clusterRs, ok := s.RootModule().Resources[clusterResourceName]
+		if !ok {
+			return fmt.Errorf("not found: %s", clusterResourceName)
+		}
+
+		clusterID := clusterRs.Primary.Attributes["id"]
+		log.Printf("[DEBUG] clusterID: %s, name %s", clusterRs.Primary.Attributes["id"], clusterRs.Primary.Attributes["name"])
```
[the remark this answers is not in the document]
arjunmahishi: Removed

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,377 @@
...
+						fmt.Sprintf("Could not delete Prometheus metric export: %s", formatAPIErrorMessage(err)),
+					)
+				}
+			}
+
+			return nil
+		}
+	}
+
+	err := retry.RetryContext(ctx, clusterUpdateTimeout, retryFunc())
```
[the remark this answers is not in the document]
arjunmahishi: Done. Created new constants. Looks like the other metric export integrations are also using this cluster timeout. Will fix that in an other PR.

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,377 @@
...
+	var state ClusterPrometheusMetricExportConfig
+	diags := req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	clusterID := state.ID.ValueString()
+	cluster := &client.Cluster{}
+	retryFunc := func() retry.RetryFunc {
```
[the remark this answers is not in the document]
arjunmahishi: Makes sense! Done.

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,377 @@
...
+						strings.Contains(apiErrMsg, "lock") {
+						// Wait for cluster to be ready.
+						clusterErr := retry.RetryContext(ctx, clusterUpdateTimeout,
+							waitForClusterReadyFunc(ctx, clusterID, r.provider.service, cluster))
+						if clusterErr != nil {
+							return retry.NonRetryableError(
+								fmt.Errorf("Error checking cluster availability: %s", clusterErr.Error()))
+						}
+						return retry.RetryableError(fmt.Errorf("cluster was not ready - trying again"))
+					}
```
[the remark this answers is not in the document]
arjunmahishi: Agreed. This can probably be moved to `utils.go` as a function. Will take that up in another PR as that would require testing other resources too.

internal/provider/metric_export_prometheus_config_resource.go
```diff
@@ -0,0 +1,377 @@
...
+		apiObj, httpResp, err := cl.GetPrometheusMetricExportInfo(ctx, clusterID)
+		if err != nil {
+			if httpResp != nil && httpResp.StatusCode < http.StatusServiceUnavailable {
+				return retry.NonRetryableError(fmt.Errorf(
+					"error getting Prometheus metric export info: %s", formatAPIErrorMessage(err)))
+			} else {
+				return retry.RetryableError(fmt.Errorf(
+					"encountered a server error while reading Prometheus metric export status - trying again"))
+			}
+		}
```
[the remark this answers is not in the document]
arjunmahishi: Made a small change - `< http.StatusInternalServerError`

---

> How do we know this error is retryable

I think the assumption here is that all internal servers are temporary and retryable. Because this is a GET request with no URL params.

---

> I'm also curious what happens if the retry func times out after falling into this block over and over?

I think it reaches here and shows up as a request timeout: https://github.com/cockroachdb/terraform-provider-cockroach/blob/ef9d625265a359472c218ebe977ad943d8358927/vendor/github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry/state.go#L260-L262

---

> In this block we've never actually logged the or returned the actual error.

It would be helpful to log this error condition. I've added a log line that denotes that there is an internal server error so the request will be retried.

internal/provider/metric_export_prometheus_config_resource_test.go
arjunmahishi: I think this block was copied from other metric export integrations. This is suppose to spin up an actual AWS based CRDB cluster and run the tests on that. To test the interaction with the cloud providers.  But this seems to be `t.Skip()`d in all other integrations. So, it's better to just remove this block until we actually plan to run this test in the future.

internal/provider/metric_export_prometheus_config_resource_test.go
arjunmahishi: This block has been removed. Refer to the reply on the previous comment.

internal/provider/metric_export_prometheus_config_resource_test.go
arjunmahishi: Yes. This should work with serverless as well. But we are not spinning up any actual resources in the tests. Refer to above replies. 

Will keep this in mind. Thanks!

arjunmahishi:
@fantapop Can you please give it one final look?
Thanks!

internal/provider/metric_export_prometheus_config_resource_test.go
fantapop: In general I prefer to keep the skip blocks so that when the tests run, we can see which tests are skipped.  But considering that the retry logic in this resource is fairly complex and that it does not rely on something in the AWS provider, we should add an acceptance test for it.

If you look at another resource where this is implemented you can see that the addition of the acceptance test is much simpler than the integration test since no mocking needs to occur.   Usually it looks something like [this](https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/internal/provider/maintenance_window_test.go#L34-L38).  Let me know if you need further help adding this.

arjunmahishi: We have added this back and removed `t.Skip()`

internal/provider/metric_export_prometheus_config_resource_test.go
arjunmahishi: We have added a check to make sure that the status is `ENABLED` in API response as well.

arjunmahishi:
@fantapop we have made the requested changes

vendor/modules.txt
```diff
@@ -205,6 +205,7 @@ github.com/hashicorp/terraform-plugin-framework/provider/schema
 github.com/hashicorp/terraform-plugin-framework/providerserver
 github.com/hashicorp/terraform-plugin-framework/resource
 github.com/hashicorp/terraform-plugin-framework/resource/schema
+github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault
```
fantapop, line 208: did this end up being used?  I can't find it in the source.

fantapop (approved):
I've added some additional small nits to the review.  Please take a look at these before merging.  Also, please rebase off the latest main branch.   Thanks for everyone's work on this!

CHANGELOG.md
```diff
@@ -21,7 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   to true, attempts to delete the cluster will fail. Set to false to disable
   delete protection.
 
-## [1.5.0] - 2024-04-26
+## [1.5.0] - 2024-05-26
```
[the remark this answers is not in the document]
aa-joshi: This is updated.

vendor/modules.txt
aa-joshi: This got introduced in first iteration of development. I missed to run `go mod vendor` after clean up. Thanks for pointing it out!

internal/provider/metric_export_prometheus_config_resource_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+		return nil
+	}
+}
+
+func getTestMetricExportPrometheusConfigResourceCreateConfig(name string) string {
+	return fmt.Sprintf(`
+resource "cockroach_cluster" "test" {
+  name           = "%s"
+  cloud_provider = "AWS"
```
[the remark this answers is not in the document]
aa-joshi: We have changed it to GCP.

PR #201, opened 2024-04-12
Allow enabling log export for serverless clusters.

alyshanjahani-crl:


fantapop:
Are there any test updates we can make for this change?

alyshanjahani-crl:
> Are there any test updates we can make for this change?

I didn't take a look at the tests here, tbh i was kinda surprised we're doing this checking here in terraform provider. Because the actual console/intrusion endpoints perform all this validation I.e previously, enabling log export for serverless was not supported, console/intrusion would return an error. After i implemented it, it is supported and i was surprised that i even had to make this PR :/  

Is the validation done here so that the error message is user readable? Does marking an error from console/intrusion as a UserError not bubble up through the terraform provider well?

fantapop:
> > Are there any test updates we can make for this change?
> 
> I didn't take a look at the tests here, tbh i was kinda surprised we're doing this checking here in terraform provider. Because the actual console/intrusion endpoints perform all this validation I.e previously, enabling log export for serverless was not supported, console/intrusion would return an error. After i implemented it, it is supported and i was surprised that i even had to make this PR :/

If there are no differences between the tf resources or expected differences from the perspective of the api, I think its fine to test with either serverless or dedicated.  We do have an integration test for each resource and the current one for `log_export_config` is spinning up a dedicated AWS cluster ([here](https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/internal/provider/log_export_config_resource_test.go#L277-L291)).   I think we should swap it to use a serverless cluster instead so its quicker and doesn't use up an AWS account every time we run to the acceptance tests.

> Is the validation done here so that the error message is user readable? Does marking an error from console/intrusion as a UserError not bubble up through the terraform provider well?

Sometimes we do some validation up front to provide a quicker or better message to user.  In general I'm a fan of relying on the api to provide those errors and I think that generally works.

fantapop:
I just noticed that the acceptance tests for this are actually marked as skipped for now.  So no need to update the test for now.

fantapop (approved):

andy-kimball:
But does this mean we have not tests for this feature, if there are none for Serverless and Dedicated tests are disabled?

fantapop:
I think we're good to close this now as it was only in support of the cloud 2.0 branch which has now been merged back into the main line.   This change was added to the main line here: https://github.com/cockroachdb/terraform-provider-cockroach/pull/199 and released in v1.7.7

PR #205, opened 2024-04-26
add `delete_protection` to the Cluster resource and data source

rgcase:
Enabling delete protection on a cluster causes attempts to delete the cluster to fail. Users can set this to protect against accidental deletion of important clusters by making deletion a two-step process. To delete a protected cluster, first disable delete protection.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)


internal/provider/cluster_resource.go
```diff
@@ -404,6 +411,12 @@ func (r *clusterResource) Create(
 		clusterSpec.SetParentId(parentID)
 	}
 
+	deleteProtection := client.DELETEPROTECTIONSTATETYPE_DISABLED
+	if plan.DeleteProtection.ValueBool() {
+		deleteProtection = client.DELETEPROTECTIONSTATETYPE_ENABLED
+	}
+	clusterSpec.SetDeleteProtection(deleteProtection)
```
fantapop, line 418: are the client constants actually just bools?  Could we just:

```go
clusterSpec.SetDeleteProtection(plan.DeleteProtection.ValueBool())
```

internal/provider/cluster_resource.go
```diff
@@ -736,6 +749,17 @@ func (r *clusterResource) Update(
 		clusterReq.SetParentId(parentID)
 	}
 
+	if !(plan.DeleteProtection.IsNull() || plan.DeleteProtection.IsUnknown()) &&
```
fantapop, line 752: did you find that these checks were still necessary after adding the default?

internal/provider/cluster_resource.go
```diff
@@ -882,6 +906,13 @@ func loadClusterToTerraformState(
 		state.ParentId = types.StringValue(*clusterObj.ParentId)
 	}
 
+	if clusterObj.DeleteProtection != nil &&
```
fantapop, line 909: I forget, was DeleteProtection being a pointer just because we missed adding it to that json spec array?

internal/provider/cluster_resource.go
rgcase: No the API uses enums so I have to map from bool to the enum value.

internal/provider/cluster_resource.go
rgcase: Yes, but I think it's actually the thing we want since we don't include the UNSPECIFIED value in the JSON spec so we'd have no way to represent an unset value without this being a pointer.

internal/provider/cluster_resource.go
rgcase: Oh I forgot about that. Let me see if I can verify it.

internal/provider/cluster_resource.go
fantapop: Interesting. I feel like we'd want to default the UNSPECIFIED value to false.

rgcase: The server should never return `UNSPECIFIED`, but the client can leave it unset which will map to `UNSPECIFIED` on the server, which the server will map to `DISABLED` during cluster creation. Leaving it unset during update will map to `UNSPECIFIED` on the server which will mean "don't change the value".

fantapop: I guess i'm surprised we would allow explicitly passing a value that meant don't change it via a patch when they should already be able to exclude passing it in.

internal/provider/cluster_resource.go
rgcase: I thought about this some more and I think I actually do want this, and I also don't want a default value on the field because it causes the plan to always have a `false` value if the spec leaves the field unset. In that case I actually just want to let the server decide the state and for TF it leave it alone.

internal/provider/cluster_resource.go
rgcase: I think it's kind of a dissonance between our different JSON, protobuf, and various client and server generated code. `UNSPECIFIED` isn't included in the spec so users should never know that they can pass it, so the way they'll represent "unset" is to not set the pointer in the Go SDK or leave out the field from the JSON payload.

In the past there has been value in having a way to represent both "unset the value" and "leave the value alone" which I think can make it valuable to have the "`UNSPECIFIED'" value even if the only the server knows about it for now.

fantapop: Since we don't generally publish it as part of the spec, I guess its probably not causing any confusion.  incidentally, I do see it in a few of the enums in the api spec.  Not this particular one.  Do you have any examples of times this has provided value for the api?

vendor/modules.txt
```diff
@@ -205,13 +205,13 @@ github.com/hashicorp/terraform-plugin-framework/provider/schema
 github.com/hashicorp/terraform-plugin-framework/providerserver
 github.com/hashicorp/terraform-plugin-framework/resource
 github.com/hashicorp/terraform-plugin-framework/resource/schema
+github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault
```
fantapop, line 208: It seems like this is not up to date

internal/provider/cluster_resource.go
```diff
@@ -247,6 +247,11 @@ func (r *clusterResource) Schema(
 					validators.FolderParentID(),
 				},
 			},
+			"delete_protection": schema.BoolAttribute{
+				Computed:    true,
+				Optional:    true,
+				Description: "Set to true to enable delete protection on the cluster.",
```
fantapop, line 253: It feels like the behavior of not setting it should also be documented somewhere.  Specifically how the behavior interacts with a user who has set the value via the UI.

rgcase: I updated this description to describe the behavior. Let me know what you think.

vendor/modules.txt
rgcase: Sorry ya I had the change to remove this package again local.

internal/provider/cluster_resource.go
rgcase: Looks like this is why there are some `"UNSPECIFIED"` values in the spec, they use a different name for the default `0` value. Not ideal...
Screenshot 2024-04-29 at 1 06 14 PM

I don't know if I have an example where an enum in particular needed the "unset" vs "don't touch" property but it came up when we were implementing the API for serverless cluster usage limits and we had to distinguish between "leave the usage limits alone", "set usage limits to unlimited" and "set usage limits to 0".

In general I just find that it's good for backwards compatibility and nearly impossible to retrofit on later.

internal/provider/cluster_resource.go
fantapop: It seems truthful.  I'm still a bit concerned about the usability of this documentation but I'm good with this for now. We can revise it later after discussion with @marksoper and the docs team.  I know they're already planning a revamp of the docs. 

To be more specific, It seems like the logical next question is to ask is "well... what does the server set it to?".  If it were me, I'd rather just set it at that point than go check what the actual default is.

fantapop (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Added
+
+- Added `delete_protection` to the Cluster resource and data source. When set
```
fantapop, line 12: are you planning to tag?  You could update this now for the next version.

internal/provider/cluster_resource_test.go
```diff
@@ -685,28 +692,55 @@ func TestIntegrationDedicatedClusterResource(t *testing.T) {
...
 		Check:  resource.TestCheckResourceAttr("cockroach_cluster.dedicated", "dedicated.num_virtual_cpus", "4"),
 	}
 
-	testDedicatedClusterResource(t, clusterName, true, scaleStep)
+	failedDeleteStep := resource.TestStep{
+		Config:      " ",
+		Destroy:     true,
+		ExpectError: regexp.MustCompile(".*Cannot destroy cluster with delete protection enabled*"),
+	}
+
```
fantapop, line 737: It's kind of unfortunate this step is not getting tested via the acceptance tests.  can it be added somewhere more general?

CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.6.0] - 2024-05-29
```
fantapop, line 10: I know I asked you to add this yesterday but now since I'm planning to merge my branches down as well can you remove it?  :-P

internal/provider/cluster_resource_test.go
rgcase: Oh, ya I think I can make that work. Let me give it a try.

internal/provider/cluster_resource_test.go
```diff
@@ -757,6 +785,20 @@ func testDedicatedClusterResource(
...
 		},
+		{
+			Config: getTestDedicatedClusterResourceConfig(clusterName, latestClusterMajorVersion, false, 2, ptr(true)),
+			Check:  resource.TestCheckResourceAttr(resourceName, "delete_protection", "true"),
+		},
+		{
+			// Delete step that fails since delete protection is enabled.
+			Config:      " ",
+			Destroy:     true,
+			ExpectError: regexp.MustCompile(".*Cannot destroy cluster with delete protection enabled*"),
```
fantapop, line 796: nit: It might be nice to show that resources that are attached to cluster have not been destroyed.   Sorry I keep finding one off things.  This one is optional.

fantapop (approved):

internal/provider/cluster_resource_test.go
rgcase: Talked about this offline. There aren't already any peripheral resources for this cluster and it would be painful and make the test significantly more complicated to add them so I'll skip it for now.

PR #208, opened 2024-06-05
fix allowlist_resource churn on empty name

fantapop:
Previously, allowlist.name attribute was not marked as computed so values from the server would not set the state in case it was unused. The server side default of "" would then conflict with a nil value in the state and show that it needed to be updated on every apply. We now mark the allowlist name as computed so that it can update the state when it is not set.  Additionally we stop sending the value on updates unless it explicitly set in the resource.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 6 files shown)


rgcase (approved):
:shipit:

PR #214, opened 2024-06-07
private_endpoint_services: fix backwards compat bug

carloruiz:
Previously, when reading the `private_endpoint_services` resource for AWS the clusters, the '.aws' field remained unset, despite expectations otherwise. This commit fixes the bug to ensure that field remains set.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [ ] Acceptance test(s)
- [x] Example(s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.7.6] - 2024-06-10
+
 ## Fixed
 
 - Update docs for allowlist resource to clear up with cidr_mask is
@@ -16,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed mention of Limited Access for Azure clusters in README
 
 - Added some example values for clarity in README
+
+- Fix bug when reading `cockroach_private_endpoint_services.#.aws.service_name`.
 
 ## [1.7.5] - 2024-06-06
 
--- a/internal/provider/private_endpoint_services_resource.go
+++ b/internal/provider/private_endpoint_services_resource.go
@@ -252,13 +252,6 @@ func loadEndpointServicesIntoTerraformState(
 			EndpointServiceId: types.StringValue(service.GetEndpointServiceId()),
 		}
 
-		if services[i].CloudProvider.String() == "AWS" {
-			services[i].Aws = PrivateLinkServiceAWSDetail{
-				ServiceName: types.StringValue(service.Aws.GetServiceName()),
-				ServiceId:   types.StringValue(service.Aws.GetServiceId()),
-			}
-		}
-
 		traceAPICall("GetAvailabilityZoneIds")
 		apiAZs := service.GetAvailabilityZoneIds()
 		azs := make([]types.String, len(apiAZs))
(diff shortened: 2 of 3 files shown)


internal/provider/private_endpoint_services_resource_test.go
```diff
@@ -36,7 +36,7 @@ import (
 func TestAccDedicatedPrivateEndpointServicesResource(t *testing.T) {
 	t.Parallel()
 	clusterName := fmt.Sprintf("%s-endpt-svc-%s", tfTestPrefix, GenerateRandomString(2))
-	testPrivateEndpointServicesResource(t, clusterName, false /* useMock */, false /* isServerless */)
+	testPrivateEndpointServicesResource(t, clusterName, false /* useMock */, false /* isServerless */, client.CLOUDPROVIDERTYPE_GCP)
 }
```
fantapop, line 40: should we also run a dedicated test for AWS?

internal/provider/private_endpoint_services_resource.go
```diff
@@ -252,21 +253,23 @@ func loadEndpointServicesIntoTerraformState(
...
 		azs := make([]types.String, len(apiAZs))
 		for j, az := range apiAZs {
 			azs[j] = types.StringValue(az)
 		}
 		services[i].AvailabilityZoneIds = azs
-		services[i].Aws.AvailabilityZoneIds = azs
+
+		// Set the .AWS field for backward compatibility.
+		// Note that `services[i].CloudProvider.String()` "\"AWS\"".
+		if strings.Contains(services[i].CloudProvider.String(), "AWS") {
```
fantapop, line 266: I think you can use `ValueString()` here to get the actual value inside.
```suggestion
		if services[i].CloudProvider.ValueString() == "AWS" {
```

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added some example values for clarity in README
 
+- Fix bug when reading `cockroach_private_endpoint_services.#.aws.service_name`.
```
fantapop, line 20: Can you add a tag? I think we'll want to release this asap.  Monday probably makes sense.

internal/provider/private_endpoint_services_resource.go
carloruiz: nice.

internal/provider/private_endpoint_services_resource.go
```diff
@@ -252,21 +253,23 @@ func loadEndpointServicesIntoTerraformState(
 			EndpointServiceId: types.StringValue(service.GetEndpointServiceId()),
 		}
 
-		if services[i].CloudProvider.String() == "AWS" {
```
fantapop, line 255: Is this the root cause of the original bug?  This clause was getting skipped because .String() is used instead of .ValueString()?

carloruiz: Yes

fantapop: I think it was me that missed that during review.   The terraform api makes this an easy mistake to make.

internal/provider/private_endpoint_services_resource_test.go
carloruiz: Serverless is already covering the logic for acceptance tests here since it's the same for Serverless and Dedicated.

CHANGELOG.md
carloruiz: done.

carloruiz:
Taged this with `v1.7.6`

fantapop (approved):
LGTM!

carloruiz:
TFTR!

PR #219, opened 2024-06-18
provider: add services map to private_endpoint_services

carloruiz:
Previously, the private_endpoint_services resource contains the regional services in a list. This made it hard for users to access the service for a given region. This commit adds a service map that contains all the services but are keyed by the cloud region.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Enable log export for serverless clusters.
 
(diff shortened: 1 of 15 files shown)


fantapop (commented):
Can you update the aws_privatelink workflow to use the new map as well?  It seems like this should be the preferred usage we recommend.

fantapop (approved):
nice.  Thank you for adding this!

PR #221, opened 2024-07-02
private_endpoints_service: wait for all regions

carloruiz:
Previously, when waiting for private endpoints service creation, the resource could be considered ready if a single resource was ready and others were not yet created. This commit fixes the bug by explicitly waiting for all regional services to be AVAILABLE.

**Commit checklist**
- [x] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/internal/provider/private_endpoint_services_resource.go
+++ b/internal/provider/private_endpoint_services_resource.go
@@ -75,7 +75,8 @@ You must install/upgrade to [version 1.7.6](https://github.com/cockroachdb/terra
 			NestedObject: serviceObject,
 		},
 		"services": schema.ListNestedAttribute{
-			Computed: true,
+			Computed:            true,
+			MarkdownDescription: "A list of regional private endpoint services for the cluster",
 			PlanModifiers: []planmodifier.List{
 				listplanmodifier.UseStateForUnknown(),
 			},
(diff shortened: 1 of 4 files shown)


internal/provider/private_endpoint_services_resource.go
```diff
@@ -339,44 +344,55 @@ func (r *privateEndpointServicesResource) ImportState(
...
-		apiServices, httpResp, err := cl.ListPrivateEndpointServices(ctx, clusterID)
-		if err != nil {
-			if httpResp != nil && httpResp.StatusCode < http.StatusInternalServerError {
-				return retry.NonRetryableError(fmt.Errorf("error getting endpoint services: %s", formatAPIErrorMessage(err)))
-			} else {
-				return retry.RetryableError(fmt.Errorf("encountered a server error while reading endpoint status - trying again: %v", err))
-			}
+) (*client.PrivateEndpointServices, *retry.RetryError) {
+	// Set of all regions.
+	regions := func() map[string]bool {
```
fantapop, line 353: nit: I think renaming this could help code readability later in this function
```suggestion
	pendingClusterRegions := func() map[string]bool {
```

fantapop (approved):
I left one minor variable naming nit that I think would improve readability.  Thanks for working on this.  Nice test as well.  It was clear and thorough.

fantapop:
also It looks like the doc generation action is failing.

PR #222, opened 2024-08-01
convert from provisioned_capacity to provisioned_vcpus

andy-kimball:
Rename the `provisioned_capacity` field to `provisioned_vcpus` and update all
tests and examples to use the vCPUs unit rather than request units by dividing
by 500. This is part of the overall migration to vCPUs for Cloud 2.0.

**Commit checklist**
- [X] Changelog
- [X] Integration test(s)
- [X] Acceptance test(s)
- [X] Example(s)


andy-kimball:
Similar to the SDK PR, I will hard reset the `cloud-20` branch to `main` and then apply this sequence of commits to that branch.

IMPORTANT: All but the last commit are older commits that are being reapplied to a new base. All of the older commits have already been reviewed. While I did need to make a few tweaks to those to get past some merge conflicts, there are no material changes. I'd advise only reviewing the last commit.

andy-kimball:
Also, I won't merge this PR until all of the required API changes are in production (it seems we're still pretty far behind in our rollouts).

rgcase (approved):
I only looked at the last commit but this seems fine. Will we cut a new version of the Cloud 2.0 TF provider for this?

andy-kimball:
I will hard reset the current version of the Cloud 2.0 TF provider to this commit. Only Private Preview customers have been using this, but they will need to switch any config they have to use `provisioned_vcpus` rather than `provisioned_capacity`.

PR #224, opened 2024-08-08
update examples for Cloud 2.0

andy-kimball:
**Commit checklist**
- [X] Changelog
- [X] Integration test(s)
- [X] Acceptance test(s)
- [X] Example(s)


docs/resources/cluster.md
```diff
@@ -45,6 +46,19 @@ resource "cockroach_cluster" "serverless" {
   ]
   delete_protection = false
 }
+
+resource "cockroach_cluster" "basic" {
```
mdlinville, line 50: Is it a good practice to have more than one resource in the same template?

Separately, what's the significance of the resource name vs the cluster name? Is there anything special that happens depending on what you call the resource, or only the plan field within the resource?

What does it mean if the serverless field is empty? It seems almost like there should be instead a "dedicated" field that is only true for Advanced. WDYT?

examples/resources/cockroach_cluster/resource.tf
```diff
@@ -30,3 +31,16 @@ resource "cockroach_cluster" "serverless" {
   ]
   delete_protection = false
 }
+
+resource "cockroach_cluster" "basic" {
```
mdlinville, line 35: Here's what is "basic" and what is "cockroach-basic"?

docs/resources/cluster.md
```diff
@@ -13,12 +13,13 @@ CockroachDB Cloud cluster.
...
-resource "cockroach_cluster" "dedicated" {
-  name           = "cockroach-dedicated"
+resource "cockroach_cluster" "advanced" {
+  name           = "cockroach-advanced"
   cloud_provider = "GCP"
+  plan           = "ADVANCED"
   dedicated = {
     storage_gib  = 15
-    machine_type = "n1-standard-4"
+    machine_type = "n2-standard-4"
```
mdlinville, line 22: https://github.com/cockroachdb/terraform-provider-cockroach/pull/224/files#diff-78accda5330258f3c72bbbd93e54e91c1a00463630f0003901bbda25a7326f21R99, you have number of CPUs I think. Can it be either?

examples/workflows/cockroach_basic_cluster/main.tf
```diff
@@ -0,0 +1,87 @@
+variable "cluster_name" {
+  type     = string
+  nullable = false
+}
+
+variable "sql_user_name" {
```
mdlinville, line 6: Is it actually required to create a SQL user with Terraform? I cancel out of that all the time in the CC UI.

examples/workflows/cockroach_basic_cluster/main.tf
```diff
@@ -0,0 +1,87 @@
...
+  type     = string
+  nullable = false
+  default  = "maxroach"
+}
+
+# Remember that even variables marked sensitive will show up
+# in the Terraform state file. Always follow best practices
+# when managing sensitive info.
+# https://developer.hashicorp.com/terraform/tutorials/configuration-language/sensitive-variables#sensitive-values-in-state
+variable "sql_user_password" {
```
mdlinville, line 16: Is there a way to show the user how to hash this before sending or otherwise be safe, besides making them read the Hashicorp doc?

examples/workflows/cockroach_basic_cluster/main.tf
```diff
@@ -0,0 +1,87 @@
...
+
+terraform {
+  required_providers {
+    cockroach = {
+      source = "cockroachdb/cockroach"
+    }
+  }
+}
+provider "cockroach" {
+  # export COCKROACH_API_KEY with the cockroach cloud API Key
```
mdlinville, line 42: Will they know what to do here?

examples/workflows/cockroach_basic_cluster/main.tf
```diff
@@ -0,0 +1,87 @@
...
+}
+
+resource "cockroach_cluster" "example" {
+  name           = var.cluster_name
+  cloud_provider = var.cloud_provider
+  plan           = "BASIC"
+  serverless     = {}
```
mdlinville, line 49: If the plan is BASIC can we imply the "serverless" field? I think maybe it would be an error for them to have anything within the brackets.

examples/workflows/cockroach_basic_cluster/main.tf
```diff
@@ -0,0 +1,87 @@
...
+  regions        = [for r in var.cloud_provider_regions : { name = r }]
+}
+
+resource "cockroach_sql_user" "example" {
+  name       = var.sql_user_name
+  password   = var.sql_user_password
+  cluster_id = cockroach_cluster.example.id
+}
+
+resource "cockroach_database" "example" {
```
mdlinville, line 59: I wonder if we can give them an example of initializing `movr` or something here?

examples/workflows/cockroach_basic_cluster/main.tf
```diff
@@ -0,0 +1,87 @@
...
+  name       = "example-database"
+  cluster_id = cockroach_cluster.example.id
+}
+
+resource "cockroach_service_account" "example_scoped_sa" {
```
mdlinville, line 64: Does this field work the same across cloud providers?

examples/workflows/cockroach_basic_cluster/main.tf
```diff
@@ -0,0 +1,87 @@
...
+  name        = "example-scoped-service-account"
+  description = "A service account providing limited read access to single cluster."
+}
+
+resource "cockroach_user_role_grant" "example_limited_access_scoped_grant" {
```
mdlinville, line 69: Perhaps we can change this to something that is more widely used, like Cluster Admin? I don't think Cluster Operator is broadly used or something we recommend. If we want to show how to grant no privilege at all, we could grant ORG MEMBER. But I think that's granted by default.

mdlinville (commented):
Some initial questions, which may or may not be ridiculous, because I'm still coming up to speed on Terraform.

docs/resources/cluster.md
andy-kimball: > Is it a good practice to have more than one resource in the same template?
I don't know the convention. However, this is how it was before; I just added one additional example to the 2 that already existed in `examples/resources/cockroach_cluster/resource.tf`. This doc file is generated from that file, so I assume that if you want to show multiple examples in the `cluster.md` docs page, you need to include multiple examples in the `resource.tf` file.

> Separately, what's the significance of the resource name vs the cluster name? Is there anything special that happens depending on what you call the resource, or only the plan field within the resource?
`cockroach_cluster` is the type of the resource, `basic` is the name of this particular instance of the resource. The resource name is used if you need to reference this resource instance in another TF rule, e.g.:
```
resource "cockroach_client_ca_cert" "example" {
  id            = cockroach_cluster.basic.id
  x509_pem_cert = file("client-ca.crt")
}
```
The resource name is used by TF. The cluster name is used by CC.

> What does it mean if the serverless field is empty?
The `serverless` field needs to be defined for "Basic" and "Standard" clusters. For "Standard" clusters, it needs to contain the `usage_limits {provisioned_vcpus=X}` field. For "Basic" clusters, if no usage limits are specified, then an "Unlimited" cluster is created, where there's no limit to the number of RU's that can be consumed in a given month.

> It seems almost like there should be instead a "dedicated" field that is only true for Advanced. WDYT?
I'm not sure I understand what you're suggesting. Can you restate?

examples/resources/cockroach_cluster/resource.tf
andy-kimball: See above: "basic" is the name of the TF resource. "cockroach-basic" is the name of the CC cluster.

docs/resources/cluster.md
andy-kimball: Yes, my understanding is that it can be either. I just fixed this b/c I noticed it was using an old machine type that we're phasing out.

examples/workflows/cockroach_basic_cluster/main.tf
andy-kimball: It's probably not required, I just followed the pattern from other workflows.

examples/workflows/cockroach_basic_cluster/main.tf
andy-kimball: I don't know. Perhaps @fantapop knows.

examples/workflows/cockroach_basic_cluster/main.tf
andy-kimball: This is how the other examples work. I think it'd be a different change to improve this, as all the examples that have this comment should be updated in the same way. If you have ideas on how to reword this or to point to docs that explain this better, I think it'd be a good improvement.

examples/workflows/cockroach_basic_cluster/main.tf
andy-kimball: They can have `usage_limits` in the brackets. If they have nothing, it indicates there are no limits enforced on their Basic Serverless cluster. It would be possible to imply the "serverless" field if the plan is BASIC, but that would be a different change. @marksoper for visibility.

examples/workflows/cockroach_basic_cluster/main.tf
andy-kimball: Maybe, I think it'd be a different change, though, since it would/should cut across all our cluster types, not just Basic. @marksoper would be the person to spearhead any changes along these lines, or you could submit a PR if you have good ideas on how to improve.

examples/workflows/cockroach_basic_cluster/main.tf
andy-kimball: I don't know; this is how it was done in the other examples.

examples/workflows/cockroach_basic_cluster/main.tf
andy-kimball: This is how the other examples work: it would be a separate change to do something like that, since it's not connected with Cloud 2.0. You can file an issue and/or talk with @marksoper, since a change like this should cut across all the cluster types.

docs/resources/cluster.md
mdlinville: Do we have a preference on defining the machine type vs the number of CPUs?

examples/workflows/cockroach_basic_cluster/main.tf
mdlinville: Looking at their docs it looks too complicated for us to cover.

examples/workflows/cockroach_basic_cluster/main.tf
marksoper: @mdlinville I agree with @andy-kimball that the two concerns: 1. how we use role grants in our examples and 2. our development for Cloud 2.0 should be separated.

The role grant concern you're referring to was introduced ~4 months ago when @fantapop added support for role grants in part because of an ask from JPMC.

marksoper: I understood the use of CLUSTER_OPERATOR was meant to show the assignment of cluster management capability to a single user for that cluster alone - which was part of the original JPMC requirement.

marksoper: @mdlinville could you please clarify the concern with CLUSTER_OPERATOR. From this doc it looks like CLUSTER_OPERATOR is consistent with our current docs and is less permissive than CLUSTER_ADMIN while still giving the user cluster management capabilities.
https://www.cockroachlabs.com/docs/cockroachcloud/authorization

mdlinville: I suppose I was just remembering discussions during development of FGAC, where I was asked to use Cluster Admin rather than Cluster Operator in some contexts. I think I'm good here.

mdlinville (approved):

docs/resources/cluster.md
andy-kimball: I'm not sure, referring that to @marksoper.

PR #225, opened 2024-08-08
[CC-29279] improve user_role_grant/s testing with sa user

linhcrl:
Previously, the top level terraform user was used for testing in the user_role_grant_resource and user_role_grants_resource tests. Changes to this user are fragile and could affect other tests or require manual fixing on failure. This commit updates the user associated with those tests to be a newly created service account user. Additionally this commit also includes a new acceptance test for
user_role_grants_resource.

Checklist is unnecessary since the changes only affect tests

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -123,7 +123,9 @@ process. Unit tests are great, but rare since TF provider methods don't usually 
 
 **It's generally a good idea to run acceptance tests affected by your change manually before submitting a PR.** If
 you're using GoLand, you can run them individually, setting `TF_ACC=1` and setting `COCKROACH_SERVER` and
-`COCKROACH_API_KEY` to appropriate values.
+`COCKROACH_API_KEY` to appropriate values. Please ensure the service account associated with the provided
+`COCKROACH_API_KEY` has the necessary role/permissions (Org Admin, Billing Coordinator, Cluster Admin, Folder Admin) to
+run the tests.
 
 Getting the mock calls right for integration tests can be challenging. You may be surprised at how many read calls are
 issued at each stage. Some optional logging has been enabled which can help
--- a/internal/provider/cluster_resource_test.go
+++ b/internal/provider/cluster_resource_test.go
@@ -698,11 +698,12 @@ func TestIntegrationDedicatedClusterResource(t *testing.T) {
 		})
 
 	s.EXPECT().GetCluster(gomock.Any(), clusterID).
-		Return(&firstUpdateCluster, httpOk, nil).Times(7)
+		Return(&firstUpdateCluster, httpOk, nil).Times(6)
 
 	// Failed Delete Attempt
 
 	// Second Update
+
 	s.EXPECT().UpdateCluster(gomock.Any(), clusterID, gomock.Any()).
 		DoAndReturn(func(context.Context, string, *client.UpdateClusterSpecification,
 		) (*client.Cluster, *http.Response, error) {
--- a/internal/provider/user_role_grant_resource_test.go
+++ b/internal/provider/user_role_grant_resource_test.go
@@ -40,18 +40,16 @@ func TestAccRoleGrantResource(t *testing.T) {
 	t.Parallel()
 	clusterName := fmt.Sprintf("%s-rg-resource-%s", tfTestPrefix, GenerateRandomString(4))
 
-	// Skip the org test for the Acceptance Test because the test relies on the
-	// user having access to set an org level role. In order to avoid adding
-	// configuration, we use the service account user as the user to grant
-	// resources to.  This user doesn't seem to have the necessary permissions
-	// to perform the org grant so we skip it.
-	testRoleGrantResource(t, clusterName, true /* skipOrgTest */, false /* , false /* useMock */)
+	// The Acceptance Test relies on the user having access to set an org level role.
+	// Please ensure that the service account used satisfies this condition.
+	testRoleGrantResource(t, clusterName, false /* useMock */)
 }
 
 // TestIntegrationRoleGrantResource attempts to create, check, and destroy a
 // role grant for a user, but uses a mocked API service.
 func TestIntegrationRoleGrantResource(t *testing.T) {
 	userID := uuid.Must(uuid.NewUUID()).String()
+	creatorID := uuid.Must(uuid.NewUUID()).String()
 	clusterID := uuid.Must(uuid.NewUUID()).String()
 	clusterName := "test-cluster-name"
 	if os.Getenv(CockroachAPIKey) == "" {
@@ -63,6 +61,12 @@ func TestIntegrationRoleGrantResource(t *testing.T) {
 	defer HookGlobal(&NewService, func(c *client.Client) client.Service {
 		return s
 	})()
+
+	serviceAccount := client.ServiceAccount{
+		Id:          userID,
+		Name:        "Test cluster SA",
+		Description: "A service account used for managing access to the test cluster",
+	}
 
 	spendLimit := int32(1)
 	cluster := client.Cluster{
@@ -80,7 +84,7 @@ func TestIntegrationRoleGrantResource(t *testing.T) {
 				Name: "us-central1",
 			},
 		},
-		CreatorId: userID,
+		CreatorId: creatorID,
 	}
 
 	orgMemberRole := client.BuiltInRole{
@@ -128,6 +132,18 @@ func TestIntegrationRoleGrantResource(t *testing.T) {
 		},
 	}
 
+	PostDeleteGetResponse := client.GetAllRolesForUserResponse{
+		Roles: &[]client.BuiltInRole{
+			orgMemberRole,
+			orgAdminRole,
+			clusterOperatorWriteRole,
+		},
+	}
+
+	// Called by Service Account Create
+	s.EXPECT().CreateServiceAccount(gomock.Any(), gomock.Any()).
+		Return(&serviceAccount, nil, nil)
+
 	// Called by Cluster Create
 	s.EXPECT().CreateCluster(gomock.Any(), gomock.Any()).
 		Return(&cluster, nil, nil)
@@ -138,6 +154,7 @@ func TestIntegrationRoleGrantResource(t *testing.T) {
 	// Called by Create - Cluster role addition
 	s.EXPECT().GetAllRolesForUser(gomock.Any(), userID).
 		Return(&PreCreateGetResponse, nil, nil)
+
 	s.EXPECT().AddUserToRole(gomock.Any(), userID, string(client.RESOURCETYPETYPE_CLUSTER), clusterID, string(client.ORGANIZATIONUSERROLETYPE_CLUSTER_OPERATOR_WRITER)).
 		Return(nil, nil, nil)
 
@@ -150,19 +167,30 @@ func TestIntegrationRoleGrantResource(t *testing.T) {
 				clusterOperatorWriteRole,
 			},
 		}, nil, nil)
+
 	s.EXPECT().AddUserToRole(gomock.Any(), userID, string(client.RESOURCETYPETYPE_ORGANIZATION), "", string(client.ORGANIZATIONUSERROLETYPE_BILLING_COORDINATOR)).
 		Return(nil, nil, nil)
 
 	// Called by testRoleGrantExists below
 	s.EXPECT().GetAllRolesForUser(gomock.Any(), userID).
-		Return(&PostCreateGetResponse, nil, nil).Times(4)
+		Return(&PostCreateGetResponse, nil, nil).Times(2)
+
+	s.EXPECT().GetServiceAccount(gomock.Any(), serviceAccount.Id).
+		Return(&serviceAccount, &http.Response{Status: http.StatusText(http.StatusOK)}, nil)
 
 	s.EXPECT().GetCluster(gomock.Any(), clusterID).
 		Return(&cluster, &http.Response{Status: http.StatusText(http.StatusOK)}, nil)
+
+	s.EXPECT().GetAllRolesForUser(gomock.Any(), userID).
+		Return(&PostCreateGetResponse, nil, nil).Times(2)
 
 	// Called by pre update Read for second step
+	s.EXPECT().GetServiceAccount(gomock.Any(), serviceAccount.Id).
+		Return(&serviceAccount, &http.Response{Status: http.StatusText(http.StatusOK)}, nil)
+
 	s.EXPECT().GetCluster(gomock.Any(), clusterID).
 		Return(&cluster, &http.Response{Status: http.StatusText(http.StatusOK)}, nil)
+
 	s.EXPECT().GetAllRolesForUser(gomock.Any(), userID).
 		Return(&PostCreateGetResponse, nil, nil).Times(2)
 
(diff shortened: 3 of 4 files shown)


linhcrl:
I'm seeing this test fail. Off the top off your head, do you know what could be going on? @fantapop 


Screenshot 2024-08-08 at 2 34 32 PM

DEVELOPMENT.md
```diff
@@ -123,7 +123,8 @@ process. Unit tests are great, but rare since TF provider methods don't usually
 
 **It's generally a good idea to run acceptance tests affected by your change manually before submitting a PR.** If
 you're using GoLand, you can run them individually, setting `TF_ACC=1` and setting `COCKROACH_SERVER` and
-`COCKROACH_API_KEY` to appropriate values.
+`COCKROACH_API_KEY` to appropriate values. Please ensure the service account associated with the provided
+`COCKROACH_API_KEY` has the necessary role/permissions (e.g. ORG ADMIN) to set org-level roles.
```
linhcrl, line 127: From my understanding, your suggestion was to just ensure the service account has the necessary permissions prior to running the acceptance tests right? 

Do you know if the API key used upon release has the necessary permissions already?

internal/provider/user_role_grant_resource_test.go
```diff
@@ -40,12 +40,9 @@ func TestAccRoleGrantResource(t *testing.T) {
 	t.Parallel()
 	clusterName := fmt.Sprintf("%s-rg-resource-%s", tfTestPrefix, GenerateRandomString(4))
 
-	// Skip the org test for the Acceptance Test because the test relies on the
-	// user having access to set an org level role. In order to avoid adding
-	// configuration, we use the service account user as the user to grant
-	// resources to.  This user doesn't seem to have the necessary permissions
-	// to perform the org grant so we skip it.
```
linhcrl, line 47: I removed this logic because now we need the newly added service account user to have two roles so that we can check that on removal of one, the other one stays. So the user needs an org level and a cluster level role to test

internal/provider/user_role_grant_resource_test.go
```diff
@@ -171,61 +198,80 @@ func TestIntegrationRoleGrantResource(t *testing.T) {
...
-	if !skipOrgTest {
-		checkFuncs = append(checkFuncs, testRoleGrantExists(orgGrantResourceName, clusterResourceName, string(client.ORGANIZATIONUSERROLETYPE_BILLING_COORDINATOR), string(client.RESOURCETYPETYPE_ORGANIZATION)))
+		testRoleGrantExists(clusterGrantResourceName, string(client.ORGANIZATIONUSERROLETYPE_CLUSTER_OPERATOR_WRITER), string(client.FOLDERRESOURCETYPETYPE_CLUSTER)),
+		testRoleGrantExists(orgGrantResourceName, string(client.ORGANIZATIONUSERROLETYPE_BILLING_COORDINATOR), string(client.RESOURCETYPETYPE_ORGANIZATION)),
 	}
 
 	resource.Test(t, resource.TestCase{
-		// After removal of the test resources, verify that the user running the
-		// tests still has at least one non ORG_MEMBER role to show that this
-		// deletion did not remove the other roles like the user_role_grants
```
linhcrl, line 214: instead of using CheckDestroy, I added a third step which removes one of the resources

internal/provider/user_role_grant_resource_test.go
```diff
@@ -238,68 +284,21 @@ func testRoleGrantResource(t *testing.T, clusterName string, skipOrgTest, useMoc
...
 	})
 }
 
-// verifyRemainingRolesIntact checks to ensure our user account still has some
-// non ORG_MEMBER roles attached to it. What differentiates the user_role_grant
-// resource from the user_role_grants resource (besides an s) is that it manages
-// only a single role grant.  If more roles are removed than just this one, this
-// is an error case.  A side affect of this check is that it relies on other
-// role grants already existing for this user.
-func verifyRemainingRolesIntact(clusterResourceName string) resource.TestCheckFunc {
```
linhcrl, line 264: this is now checked by the third test step so we don't need this function

internal/provider/user_role_grant_resource_test.go
```diff
@@ -445,10 +439,10 @@ func TestParseRoleGrantResourceID(t *testing.T) {
 	}
 }
 
-func getTestRoleGrantResourceConfig(clusterName string, skipOrgTest bool) string {
+func getTestRoleGrantResourceConfig(clusterName string, removeOrgResource bool) string {
```
linhcrl, line 442: just changed the variable name to be more descriptive/accurate

internal/provider/user_role_grants_resource_test.go
```diff
@@ -140,10 +162,11 @@ func testRoleResource(t *testing.T, userId string, useMock bool) {
...
 		Steps: []resource.TestStep{
 			{
-				Config: getTestRoleFullyPermissionsedResourceConfig(userId),
+				Config: getTestRoleFullyPermissionsedResourceConfig(),
 				Check: resource.ComposeTestCheckFunc(
-					testRoleMembership(resourceNameTest, userId, "CLUSTER_ADMIN", true),
-					resource.TestCheckResourceAttr(resourceNameTest, "id", userId),
+					testRoleMembership(resourceNameTest, "CLUSTER_ADMIN", true),
+					testRoleMembership(resourceNameTest, "ORG_MEMBER", true),
+					testRoleMembership(resourceNameTest, "ORG_ADMIN", true),
```
linhcrl, line 169: checking that all roles are present

internal/provider/user_role_grants_resource_test.go
```diff
@@ -140,10 +162,11 @@ func testRoleResource(t *testing.T, userId string, useMock bool) {
 		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
 		Steps: []resource.TestStep{
 			{
-				Config: getTestRoleFullyPermissionsedResourceConfig(userId),
+				Config: getTestRoleFullyPermissionsedResourceConfig(),
 				Check: resource.ComposeTestCheckFunc(
-					testRoleMembership(resourceNameTest, userId, "CLUSTER_ADMIN", true),
-					resource.TestCheckResourceAttr(resourceNameTest, "id", userId),
```
linhcrl, line 146: I removed this check because we don't have access to the service account's ID at this point. I added line 197 to make up for this. Not sure if that suffices

fantapop:
> I'm seeing this test fail. Off the top off your head, do you know what could be going on? @fantapop
> 
> Screenshot 2024-08-08 at 2 34 32 PM

I think its saying that an update was made to a cluster and that the value that ended up in the state is inconsistent with the value that the server is returning.  Or something like that.  That field delete protection is one of the ones I was telling you about where if you don't set it explicitly its supposed to default to what the server returns.  I'd need to look closer to try and determine the cause of this particular case.  Is the test failing due to your changes here?  Or does it fail on the main branch as well?

DEVELOPMENT.md
fantapop: The api keys used for testing in staging and prod should both have the necessary roles already.  I think you need all [the default roles](https://www.cockroachlabs.com/docs/cockroachcloud/authorization#organization-user-roles).    This is one reason I was pushing for that org_owner role to be added.

linhcrl:
> 
> I think its saying that an update was made to a cluster and that the value that ended up in the state is inconsistent with the value that the server is returning. Or something like that. That field delete protection is one of the ones I was telling you about where if you don't set it explicitly its supposed to default to what the server returns. I'd need to look closer to try and determine the cause of this particular case. Is the test failing due to your changes here? Or does it fail on the main branch as well?

it seems to be failing on the master branch too

DEVELOPMENT.md
linhcrl: okay, updated the README accordingly

internal/provider/user_role_grant_resource_test.go
fantapop: how about `excludeOrgResource`?  using remove here makes it sound like we're actively removing something.

internal/provider/user_role_grant_resource_test.go
```diff
@@ -171,61 +198,80 @@ func TestIntegrationRoleGrantResource(t *testing.T) {
...
-		// deletion did not remove the other roles like the user_role_grants
-		// resource would.
-		CheckDestroy:             verifyRemainingRolesIntact(clusterResourceName),
 		IsUnitTest:               useMock,
 		PreCheck:                 func() { testAccPreCheck(t) },
 		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
 		Steps: []resource.TestStep{
 			{
-				Config: getTestRoleGrantResourceConfig(clusterName, skipOrgTest),
+				Config: getTestRoleGrantResourceConfig(clusterName, false),
```
fantapop, line 261: when passing opaque primitives, its helpful to include the name in a comment.
Screenshot 2024-08-09 at 4 10 45 PM

internal/provider/user_role_grant_resource_test.go
```diff
@@ -171,61 +198,80 @@ func TestIntegrationRoleGrantResource(t *testing.T) {
...
 				Check:  resource.ComposeTestCheckFunc(checkFuncs...),
 			},
 			{
-				Config:      getTestRoleGrantResourceConfigUpdated(clusterName, skipOrgTest),
+				Config:      getTestRoleGrantResourceConfigUpdated(clusterName),
```
fantapop, line 265: It might be nice to rename this function to make it more clear what its doing.  Something like `getTestRoleGrantResourceWithDuplicateRole`

internal/provider/user_role_grant_resource_test.go
```diff
@@ -171,61 +198,80 @@ func TestIntegrationRoleGrantResource(t *testing.T) {
...
 				ExpectError: regexp.MustCompile(`Role already exists`),
 			},
+			// After removal of the org role, verify that the associated user still
+			// has the cluster role to show that this deletion did not remove the
+			// other roles like the user_role_grants resource would.
+			{
+				Config: getTestRoleGrantResourceConfig(clusterName, true),
+				Check:  resource.ComposeTestCheckFunc(checkFuncs[0]),
```
fantapop, line 273: It would improve readability to name that checkFunc before putting it in the array.  Or you could skip the array and just have something like:
```go
checkClusterRole := testRoleGrantExists(clusterGrantResourceName, string(client.ORGANIZATIONUSERROLETYPE_CLUSTER_OPERATOR_WRITER), string(client.FOLDERRESOURCETYPETYPE_CLUSTER))

checkOrgRole := testRoleGrantExists(orgGrantResourceName, string(client.ORGANIZATIONUSERROLETYPE_BILLING_COORDINATOR), string(client.RESOURCETYPETYPE_ORGANIZATION))

// ....

Check:  resource.ComposeTestCheckFunc(checkClusterRole, checkOrgRole)
```

Also, I don't think you need to use the `ComposeTestCheckFunc` when there is only one.  So this line would become:

```go
Check:  checkClusterRole,
```

internal/provider/user_role_grant_resource_test.go
fantapop: or you could flip it and use `includeOrgResource`.  Sometimes its easier for people to think in terms of positives instead of negatives.  But up to you.

internal/provider/user_role_grant_resource_test.go
```diff
@@ -80,7 +83,7 @@ func TestIntegrationRoleGrantResource(t *testing.T) {
 				Name: "us-central1",
 			},
 		},
-		CreatorId: userID,
```
fantapop, line 83: I think technically this is not true right?  The actual creator of the cluster will still be the user running terraform and not the service account.  It might be nice to use a different random ID here just to illustrate that it doesn't matter for the test.

internal/provider/user_role_grants_resource_test.go
```diff
@@ -30,14 +31,29 @@ import (
...
 )
 
-// TestIntegrationRoleResource attempts to create, check, and destroy
+// TestAccRoleGrantsResource attempts to create, check, and destroy a
+// real role grant for a user. It will be skipped if TF_ACC isn't set.
+func TestAccRoleGrantsResource(t *testing.T) {
+	t.Parallel()
+	// The Acceptance Test relies on the user having access to set an org level role.
+	// Please ensure that the service account used satisfies this condition.
+	testRoleResource(t, false)
```
fantapop, line 40: please add the /* param name */ comment here

internal/provider/user_role_grants_resource_test.go
fantapop: I don't think its really checking the same thing but passing in the userId as done before also seems unnecessary.  The check you added seems fine to keep.

internal/provider/user_role_grants_resource_test.go
fantapop: I'm not sure. Checking all seems good to me.

fantapop (commented):
nice work on this.  I left bit of little feedback here and there but nothing major.

fantapop:
> > I think its saying that an update was made to a cluster and that the value that ended up in the state is inconsistent with the value that the server is returning. Or something like that. That field delete protection is one of the ones I was telling you about where if you don't set it explicitly its supposed to default to what the server returns. I'd need to look closer to try and determine the cause of this particular case. Is the test failing due to your changes here? Or does it fail on the main branch as well?
> 
> it seems to be failing on the master branch too

Okay cool.  I'll see if I can figure it out.

internal/provider/user_role_grant_resource_test.go
linhcrl: good point, I found the original name super confusing too

internal/provider/user_role_grant_resource_test.go
linhcrl: wow nice catch, you're totally right. I forgot that the service account only gets assigned a cluster role and not actual ownership

internal/provider/user_role_grant_resource_test.go
linhcrl: switched to `includeOrgResource` instead

internal/provider/user_role_grant_resource_test.go
```diff
@@ -64,6 +61,12 @@ func TestIntegrationRoleGrantResource(t *testing.T) {
 		return s
 	})()
 
+	sa := client.ServiceAccount{
```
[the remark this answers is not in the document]
linhcrl: renamed it to `serviceAccount` per your suggestion

internal/provider/user_role_grants_resource_test.go
linhcrl: done

internal/provider/user_role_grant_resource_test.go
fantapop: oh, that's not what I was pointing out I don't think. I was trying to suggest that the user running terraform is the one who is the creator of both the service account and the cluster in this case.

linhcrl: yes that was my interpretation of your comment. I was just trying to explain why I got the creator mixed up. Is there anything in the code that's still incorrect?

fantapop: no, it looks good now

internal/provider/cluster_resource_test.go
```diff
@@ -698,11 +698,12 @@ func TestIntegrationDedicatedClusterResource(t *testing.T) {
 		})
 
 	s.EXPECT().GetCluster(gomock.Any(), clusterID).
-		Return(&firstUpdateCluster, httpOk, nil).Times(7)
+		Return(&firstUpdateCluster, httpOk, nil).Times(6)
```
fantapop, line 701: Do you mind separating this fix out into a separate commit.  Same PR is fine.  We should document pretty heavily whats going on here in the commit message.

linhcrl: done! Please let me know if anything is missing or if something is incorrect

fantapop (approved):
LGTM!

fantapop (approved):
LGTM!

PR #229, opened 2024-08-20
pass Plan to CreateCluster if it was specified by the user

andy-kimball:
Previously, the Plan field was not passed to the CreateCluster API, even if
it was explicitly set by the user. This prevented the server from validating
the cluster configuration. For example, the server was not able to return
an error when Plan=BASIC and provisioned_vcpus are set. The fix is to pass
the Plan field to CreateCluster if the user explicitly sets it.

**Commit checklist**
- [X] Changelog
- [X] Integration test(s)
- [X] Acceptance test(s)
- [X] Example(s)


internal/provider/cluster_resource.go
```diff
@@ -19,6 +19,7 @@ package provider
 import (
 	"context"
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
```
rgcase, line 22: nit: Sort these imports.

rgcase (approved):
LGTM with one tiny nit.

PR #231, opened 2024-09-02
provider: Add jwt_issuer resource

pritesh-lahoti:
Aligning to the CC API changes, this PR deprecates the `api_oidc_config` resource in favor of the `jwt_issuer` resource.
As the existing TF resource was not in use, the team decided to make the changes in-place.
Refined and added exhaustive acceptance tests.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)


internal/provider/jwt_issuers_test.go
```diff
@@ -0,0 +1,364 @@
...
+			CcIdentity:    types.StringValue(testCCID1),
+			IsRegex:       types.BoolValue(testIsRegex1),
+		},
+	}
+)
+
+// TestAccJWTIssuerResource attempts to create, check, and destroy
+// a real JWT Issuer. It will be skipped if TF_ACC isn't set.
+// In order to work, the `JwtIssuerEnabled` Feature Flag must be enabled and
+// the test org must have Org SSO enabled (no need for SAML/OIDC).
```
fantapop, line 84: It would be nice if we could enable this via terraform.  @marksoper do you think this is something customers would want?

internal/provider/jwt_issuers_test.go
```diff
@@ -0,0 +1,364 @@
...
+				*respIdentityMap[0].CcIdentity == testIdentityMap[0].CcIdentity.ValueString() &&
+				*respIdentityMap[0].IsRegex == testIdentityMap[0].IsRegex.ValueBool() &&
+				*respIdentityMap[1].TokenIdentity == testIdentityMap[1].TokenIdentity.ValueString() &&
+				*respIdentityMap[1].CcIdentity == testIdentityMap[1].CcIdentity.ValueString() &&
+				*respIdentityMap[1].IsRegex == testIdentityMap[1].IsRegex.ValueBool() {
+				return nil
+			}
+		}
+
+		return fmt.Errorf("JWT Issuer configuration does not have correct values")
```
fantapop, line 305: It might be nice to include the 2 objects for comparison in this error message.  It looks like we use testify in some places.  I guess its not exactly in the right format but if you could convert one of the object it might be nicer to use something like `require.Equal(t, ...)`

internal/provider/jwt_issuers_test.go
```diff
@@ -0,0 +1,364 @@
...
+			return fmt.Errorf("not found: %s", resourceName)
+		}
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("no ID is set")
+		}
+
+		traceAPICall("GetJWTIssuer")
+		if _, _, err := p.service.GetJWTIssuer(context.TODO(), rs.Primary.ID); err == nil {
+			return fmt.Errorf("JWT Issuer not deleted")
```
fantapop, line 359: I'm not quite following what's happening here.  I'm assuming the api should be returning a 404 if the issuer is deleted?  What does that look like from the perspective of the SDK?

docs/resources/jwt_issuer.md
```diff
@@ -0,0 +1,64 @@
...
+### Nested Schema for `identity_map`
+
+Required:
+
+- `cc_identity` (String) Specifies how to map the fetched token identity into CockroachDB Cloud.
+- `token_identity` (String) Specifies how to fetch external identity from the token claim.
+
+Optional:
+
+- `is_regex` (Boolean) Specifies whether the token_identity and cc_identity configurations are in regex format.
```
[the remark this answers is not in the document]
fantapop: I realized that the is_regex is baked into the api so if we wanted to change that we'd probably want to do it in the api as well.   It seems like a good Idea to me but it may not be worth the effort.  What do you all think?

fantapop (commented):
This looks good in general.  I left some comments / questions.

internal/provider/jwt_issuers.go
```diff
@@ -30,73 +30,74 @@ import (
...
 			},
 			"claim": schema.StringAttribute{
 				Optional:    true,
-				Computed:    true,
-				Description: "The JWT claim that should be used as the user identifier. Defaults to the subject.",
+				Description: "Used to identify the user from the external Identity Provider. Defaults to \"sub\".",
 			},
 			"identity_map": schema.ListNestedAttribute{
+				Optional:    true,
+				Description: "A list of mappings to map the external token identity into CockroachDB Cloud.",
```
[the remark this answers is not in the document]
pritesh-lahoti: They remain empty at the backend.
In terms of behaviour, `claim` defaults to "sub", though the value remains empty.

docs/resources/jwt_issuer.md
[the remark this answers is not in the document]
pritesh-lahoti: I agree that `is_regex` is redundant, but we could only do away with `token_identity` and `cc_identity`.

internal/provider/jwt_issuers_test.go
pritesh-lahoti: As part of resource destroy, we expect the TF Provider to issue a `DELETE` call to Console.
This bit validates that calling the `GetIssuer` does not return any issuer (expects it to be destroyed).
Updated to use HTTP status codes for clarity.

internal/provider/jwt_issuers_test.go
pritesh-lahoti: Updated. Thank you!
Did not use `require.Equal(t, ...)` to let the error be reported as per the semantics of `TestCheckFunc`.

fantapop: cool.  Looks good.

docs/resources/jwt_issuer.md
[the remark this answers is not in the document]
fantapop: I'm not sure I follow.  It's not currently redundant right?  It's necessary because we don't know if an identity should be interpreted as a string or a regex.  I'm suggesting instead of verifying it that way that we instead of different field names based on whether its a string or a regex.  Note that I've mixed the usages of these here so illustrate the additional power that it provides.  For example:

```
  identity_map = [
    {
      token_identity = "token_identity"
      cc_identity_pattern    = "\\1@example.com"
    },
    {
      token_identity_pattern = "(.*)"
      cc_identity    = "cc_identity"
    },
```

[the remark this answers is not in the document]
pritesh-lahoti: I see, got it.
However, it is possible to derive whether the input has a regex.
As it turns out, the identity map implementation within CRDB also works this way (without needing to specify whether it's a regex): https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/pgwire/identmap/ident_map.go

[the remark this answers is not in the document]
fantapop: oh I see, that is checking to see if the first token is a slash.  

> If the system-identity field starts with a slash, it will be interpreted as a regular expression

That seems like a better approach as long as a leading slash isn't a valid value for an identity.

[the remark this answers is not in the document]
pritesh-lahoti: Created a PR in managed-service for the same: https://github.com/cockroachlabs/managed-service/pull/18531

docs/resources/jwt_issuer.md
```diff
@@ -0,0 +1,58 @@
...
+  audience   = "test_audience"
+  jwks       = "{\"keys\":[{\"alg\":\"RS256\",\"e\":\"AQAB\",\"kid\":\"test_kid1\",\"kty\":\"RSA\",\"n\":\"09lq1lCEuteonwDJOhGTDak11ThplZuC9JEWQNdBnBSQwlkJQIE7A7nTBO0xTibcsh2HwYkC-N_Gs1jP4iwN3dRqnu5FwG2ct5mY8KLwJiHzToFC0MKenSFQCy0FviNtOnpiObcUlDvR2NDeNtMl_6SPzcQEt7GUTBBYZgoAxPmOgevki6ZNO6Y86xFqx3y6v8EPwW010AiC60r4AHGCTBhYF4uqmq5JH2UU4dDh9Udc-9LZxlSqPwJvnKDG2GjcnD8TsU3wjfEM_nRmx3dnXsrZUXYfNGtdv5dlHywf5AhkJmTavqcsJkgrNA-PNBghFMcCR816_kCIkCYWLWC5vQ\"}]}"
+  claim      = "email"
+  identity_map = [
+    {
+      token_identity = "token_identity"
+      cc_identity    = "cc_identity"
+    },
+    {
+      token_identity = "(.*)"
```
fantapop, line 27: Do we need the preceding slash in terraform/

internal/provider/jwt_issuers.go
```diff
@@ -30,73 +30,70 @@ import (
...
 						},
 						"cc_identity": schema.StringAttribute{
 							Required:    true,
-							Description: "The username (email or service account id) of the CC user that the token should map to.",
-						},
-						"is_regex": schema.BoolAttribute{
-							Optional:    true,
-							Computed:    true,
-							Description: "Indicates that the token_principal field is a regex value.",
+							Description: "Specifies how to map the fetched token identity to an identity in CockroachDB Cloud. In case of a regular expression for token_identity, this must contain a \\1 placeholder for the matched content.",
```
fantapop, line 81: I wonder if there will be a problem understanding if the slashes in the placeholder token should be escaped or not?  I noticed that in the docs they show up as `\1` but in the actual source they need to use `\\1`.  I'm not sure the best way to document this.  Maybe @mdlinville can comment?

internal/provider/jwt_issuers.go
```diff
@@ -109,195 +106,190 @@ func (r *apiOidcConfigResource) Configure(
...
+	var out []client.JWTIssuerIdentityMapEntry
 	for _, mapEntry := range *identityMap {
-		out = append(out, client.ApiOidcIdentityMapEntry{
-			CcIdentity:    mapEntry.CcIdentity.ValueStringPointer(),
-			IsRegex:       mapEntry.IsRegex.ValueBoolPointer(),
+		out = append(out, client.JWTIssuerIdentityMapEntry{
 			TokenIdentity: mapEntry.TokenIdentity.ValueStringPointer(),
+			CcIdentity:    mapEntry.CcIdentity.ValueStringPointer(),
+			// TODO: Need to remove once the field has been removed from the API.
+			IsRegex: types.BoolValue(false).ValueBoolPointer(),
```
fantapop, line 278: Do we need this?  Can't we just leave it empty?

internal/provider/jwt_issuers.go
```diff
@@ -109,195 +106,190 @@ func (r *apiOidcConfigResource) Configure(
...
 	if identityMap == nil {
 		return nil
 	}
 	var out []IdentityMapEntry
 	for _, mapEntry := range *identityMap {
 		out = append(out, IdentityMapEntry{
-			CcIdentity:    types.StringPointerValue(mapEntry.CcIdentity),
-			IsRegex:       types.BoolPointerValue(mapEntry.IsRegex),
 			TokenIdentity: types.StringPointerValue(mapEntry.TokenIdentity),
+			CcIdentity:    types.StringPointerValue(mapEntry.CcIdentity),
```
fantapop, line 292: are these optional? I'm surprised we're using `StringPointerValue` here

internal/provider/jwt_issuers_test.go
```diff
@@ -0,0 +1,385 @@
...
+		{
+			TokenIdentity: types.StringValue(testTokenID1),
+			CcIdentity:    types.StringValue(testCCID1),
+		},
+	}
+)
+
+// TestAccJWTIssuerResource attempts to create, check, and destroy
+// a real JWT Issuer. It will be skipped if TF_ACC isn't set.
+// In order to work, the `JwtIssuerEnabled` Feature Flag must be enabled and
```
fantapop, line 81: I just enabled `JwtIssuerEnabled` for the prod terraform acceptance test org.

internal/provider/jwt_issuers_test.go
```diff
@@ -0,0 +1,385 @@
...
+						tfjsonpath.New("id"),
+						knownvalue.NotNull(),
+					),
+					statecheck.ExpectKnownValue(
+						resourceNameTest,
+						tfjsonpath.New("id"),
+						knownvalue.StringRegexp(
+							regexp.MustCompile(
+								"^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$",
+							),
```
fantapop, line 166: I see this repeated a couple times.   Can you pull it to a named variable?

fantapop (approved):
I left some questions and nits in there but mostly looks good.  @mdlinville Can you look over the docs here?

internal/provider/jwt_issuers_test.go
marksoper: @fantapop what specifically is the feature you're referring to?

fantapop: In this case, I'm referring to being able to enable cloud SSO.   But more generally it seems like there might be a set of org configurations that we could consider allowing for edit.  I see that Org name is allowed to be updated as well.

Screenshot 2024-09-23 at 10 56 55 AM
Screenshot 2024-09-23 at 10 57 52 AM

docs/resources/jwt_issuer.md
```diff
@@ -0,0 +1,58 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "cockroach_jwt_issuer Resource - terraform-provider-cockroach"
+subcategory: ""
```
mdlinville, line 4: Do we need a subcategory? I've never seen this type of generated reference in a PR before, so apologies if it's a silly question.

I know this is a generated file, so these suggestions apply to the source. It's just simpler for me to give them here.

docs/resources/jwt_issuer.md
```diff
@@ -0,0 +1,58 @@
...
+description: |-
+  Configuration to manage external JWT Issuers to access CockroachDB Cloud API via JWT.
+---
+
+# cockroach_jwt_issuer (Resource)
+
+Configuration to manage external JWT Issuers to access CockroachDB Cloud API via JWT.
```
mdlinville, line 11:
```suggestion
Configuration to manage external Java Web Token (JWT) Issuers for authentication to the CockroachDB Cloud API.
```

docs/resources/jwt_issuer.md
```diff
@@ -0,0 +1,58 @@
...
+  ]
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `audience` (String) The intended audience for consuming the JWT.
```
mdlinville, line 39: Is this how these are generally formatted? It wants to be a table or something more structured, I think. Kind of weird to have the type in parentheses in the prose.

docs/resources/jwt_issuer.md
```diff
@@ -0,0 +1,58 @@
...
+### Read-Only
+
+- `id` (String) The unique identifier of the JWT Issuer resource.
+
+<a id="nestedatt--identity_map"></a>
+### Nested Schema for `identity_map`
+
+Required:
+
+- `cc_identity` (String) Specifies how to map the fetched token identity to an identity in CockroachDB Cloud. In case of a regular expression for token_identity, this must contain a \1 placeholder for the matched content.
```
mdlinville, line 57: It would be great to show an example of a plain string and a regex, for this and the next item.

examples/resources/cockroach_jwt_issuer/resource.tf
```diff
@@ -0,0 +1,16 @@
+resource "cockroach_jwt_issuer" "example" {
+  issuer_url = "https://accounts.google.com"
+  audience   = "test_audience"
+  jwks       = "{\"keys\":[{\"alg\":\"RS256\",\"e\":\"AQAB\",\"kid\":\"test_kid1\",\"kty\":\"RSA\",\"n\":\"09lq1lCEuteonwDJOhGTDak11ThplZuC9JEWQNdBnBSQwlkJQIE7A7nTBO0xTibcsh2HwYkC-N_Gs1jP4iwN3dRqnu5FwG2ct5mY8KLwJiHzToFC0MKenSFQCy0FviNtOnpiObcUlDvR2NDeNtMl_6SPzcQEt7GUTBBYZgoAxPmOgevki6ZNO6Y86xFqx3y6v8EPwW010AiC60r4AHGCTBhYF4uqmq5JH2UU4dDh9Udc-9LZxlSqPwJvnKDG2GjcnD8TsU3wjfEM_nRmx3dnXsrZUXYfNGtdv5dlHywf5AhkJmTavqcsJkgrNA-PNBghFMcCR816_kCIkCYWLWC5vQ\"}]}"
+  claim      = "email"
+  identity_map = [
+    {
+      token_identity = "token_identity"
+      cc_identity    = "cc_identity"
```
mdlinville, line 9: These are not very informative examples. It could be good to show some actual values, like you do for the cc_identity field.

mdlinville (approved):
Some small comments to improve the examples hopefully.

internal/provider/jwt_issuers_test.go
marksoper: @fantapop My initial thought is to wait until we have customer demand for doing this. Org level features aren't an obvious fit for Terraform:
- e.g. the org name being under Terraform when there's no org resource
- Enabling SSO is likely a one-time change. If there's configuration associated with SSO that needs to be managed then it would be a good fit. A quick check of customer signal doesn't turn up anything.

fantapop: If we did this, we'd add an Org level resource.   Ack on customer signal .  I was thinking it might be nice to be able to set this up inside the test but I realize it would probably be a painful because the only org we have to access to is the org that's running the tests.

internal/provider/jwt_issuers.go
pritesh-lahoti: The `identity_map` field is optional, but if an identity map is specified, then the `token_identity` and `cc_identity` fields are required. Looks like this has not been marked in the API, so the SDK still uses `*string` for these.

I could take it up as part of the PR to address the other deferred comments from https://github.com/cockroachlabs/managed-service/pull/18531.

docs/resources/jwt_issuer.md
pritesh-lahoti: Looks like we don't have subcategories as of today - https://registry.terraform.io/providers/cockroachdb/cockroach/latest/docs.
Similarly, we don't have subcategories for the API Docs - https://www.cockroachlabs.com/docs/api/cloud/v1#overview.

We had discussed subcategories in the context of APIs, and had concluded to revisit it later, once we have logical grouping to categorize the endpoints (or resources in the case of Terraform).

docs/resources/jwt_issuer.md
pritesh-lahoti: This is standard Terraform documentation rendering - https://registry.terraform.io/providers/cockroachdb/cockroach/latest/docs/resources/allow_list.

The descriptions are indeed in sync with the API field descriptions - https://github.com/cockroachlabs/managed-service/blob/master/console/apipb/api.proto#L3599.

internal/provider/jwt_issuers_test.go
pritesh-lahoti: Thank you, but I am surprised how the acceptance tests worked earlier.

internal/provider/jwt_issuers.go
pritesh-lahoti: Because we did not remove the field from the API yet, the response still contains this field with a default value (`false`).

internal/provider/jwt_issuers.go
pritesh-lahoti: I agree.
@mdlinville did you have a recommendation here?

I'll keep the description here and within the API in sync.

internal/provider/jwt_issuers.go
fantapop: Okay, Ideally it wouldn't be a pointer if the value is always there since it just makes the sdk more annoying to use.  But I get it based on the state we're in now.  If you end up fixing it later, that would be nice.

internal/provider/jwt_issuers.go
mdlinville: I have no idea about this. If you need a second backslash to show up, don't you just escape it? I'm not familiar the nuts and bolts here.

fantapop: The question is how best to document what value the user should use.   The current output will read as this:

```
...for token_identity, this must contain a \1 placeholder for the matched content.
```

But since the value in the string will need to be escaped, the user will actually use `\\1`.  I was suggesting that maybe we should show `\\1` instead.   But asking you if you had guidance on how to make this distinction more clear to the user.

One potential way would be to link them to an example.

pritesh-lahoti: Updated the field description to call this out. Open to suggestions.

pritesh-lahoti:
I'll be going ahead and merging the change to avoid any further merge conflicts.
I could take up any pending suggestions separately.

PR #233, opened 2024-09-11
Adding example for migrating from Basic to Standard

marksoper:
**Commit checklist**
- [ ] Changelog
- [ x] Doc gen (`make generate`)
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [x ] Example(s)


examples/workflows/migrate_from_basic_to_standard/main.tf
```diff
@@ -0,0 +1,62 @@
...
+      source = "cockroachdb/cockroach"
+    }
+  }
+}
+provider "cockroach" {
+  # export COCKROACH_API_KEY with the cockroach cloud API Key
+}
+
+#
+# Step 1 - Remove this cluster resource to migrate from Basic to Standard
```
andy-kimball, line 39: "Remove" sounds a bit like the cluster should be deleted. But what you're really showing here is just that you need to switch plan to/from `BASIC` to/from `STANDARD`, and either add or remove `provisioned_vcpus`. I guess I'd have done something more like:
```
resource "cockroach_cluster" "example" {
  name           = var.cluster_name
  cloud_provider = var.cloud_provider

  # Change "BASIC" to "STANDARD" to upgrade to Standard.
  plan           = "BASIC"

  serverless     = {
    # Uncomment this line to upgrade to Standard.
    # Modify the value based on how many vCPUs you want to provision.
    # provisioned_vcpus = 2
  }
  regions        = [for r in var.cloud_provider_regions : { name = r }]
}
```

fantapop:
@marksoper I updated the cloud-20 branch to have the latest cloud 2.0 SDK and had to force push the cloud-20 base branch as a result.  You'll need to update this branch.  One easy way would be to cherry-pick your commits onto a new branch off of the new HEAD of `cloud-20`.  I can help you with this if you want.

mdlinville:
Mark and I chatted offline and we agree that this isn't a good way to document how to change your plan. That probably needs a tutorial that starts from our Basic recipe and shows what to change to move it to Standard, then eventually the other direction too. I can use the recipe in this PR as source material.

marksoper:
I'm going to abandon this PR - see @mdlinville 's note above. Cc: @fantapop @andy-kimball @rgcase

PR #236, opened 2024-09-16
Add markers to allow remote include of examples within cluster.md

mdlinville:
These markers allow remote inclusion of the part of the file contained within them

**Commit checklist**
- [ ] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)


docs/resources/cluster.md
```diff
@@ -12,6 +12,8 @@ CockroachDB Cloud cluster.
 
 ## Example Usage
 
+<!-- ADVANCED_DOCS_MARKER -->
```
andy-kimball, line 15: I don't think we should have markers that straddle the ``` code marks. You have the begin marker _before_ the code block, but the end marker _in_ the code block. I think both should be inside the code block, and they should be valid Terraform, probably comments.

PR #237, opened 2024-09-16
update docs for cluster_resource upgrade_type

fantapop:
Note that the upgrade_type docs in the cluster data source were not updated as part of this change and remain:

"Dictates the behavior of cockroach major version upgrades."

The reason for this is that the remainder that was added seemed only to pertain to setting the values and not for reading them back.

### DIFF
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -133,7 +133,7 @@ Read-Only:
 Optional:
 
 - `spend_limit` (Number, Deprecated) Spend limit in US cents.
-- `upgrade_type` (String) Dictates the behavior of cockroach major version upgrades. If plan type is 'BASIC', this attribute must be left empty or set to 'AUTOMATIC'. Allowed values are: 
+- `upgrade_type` (String) Dictates the behavior of CockroachDB major version upgrades. Manual upgrades are not supported on CockroachDB Basic. Manual or automatic upgrades are supported on CockroachDB Standard. If you omit the field, it defaults to `AUTOMATIC`. Allowed values are:
   * MANUAL
   * AUTOMATIC
 - `usage_limits` (Attributes) (see [below for nested schema](#nestedatt--serverless--usage_limits))
--- a/internal/provider/cluster_resource.go
+++ b/internal/provider/cluster_resource.go
@@ -194,7 +194,7 @@ func (r *clusterResource) Schema(
 							stringplanmodifier.UseStateForUnknown(),
 						},
 						Validators:  []validator.String{stringvalidator.OneOf(AllowedUpgradeTypeTypeEnumValueStrings...)},
-						MarkdownDescription: "Dictates the behavior of cockroach major version upgrades. If plan type is 'BASIC', this attribute must be left empty or set to 'AUTOMATIC'. Allowed values are: " +
+						MarkdownDescription: "Dictates the behavior of CockroachDB major version upgrades. Manual upgrades are not supported on CockroachDB Basic. Manual or automatic upgrades are supported on CockroachDB Standard. If you omit the field, it defaults to `AUTOMATIC`. Allowed values are:" +
 							formatEnumMarkdownList(AllowedUpgradeTypeTypeEnumValueStrings),
 					},
 				},
--- a/docs/data-sources/cluster.md
+++ b/docs/data-sources/cluster.md
@@ -79,7 +79,7 @@ Read-Only:
 
 - `routing_id` (String) Cluster identifier in a connection string.
 - `spend_limit` (Number, Deprecated) Spend limit in US cents.
-- `upgrade_type` (String) Dictates the behavior of cockroach major version upgrades.
+- `upgrade_type` (String) Dictates the behavior of CockroachDB major version upgrades.
 - `usage_limits` (Attributes) (see [below for nested schema](#nestedatt--serverless--usage_limits))
 
 <a id="nestedatt--serverless--usage_limits"></a>
--- a/internal/provider/cluster_data_source.go
+++ b/internal/provider/cluster_data_source.go
@@ -91,7 +91,7 @@ func (d *clusterDataSource) Schema(
 					},
 					"upgrade_type": schema.StringAttribute{
 						Computed:    true,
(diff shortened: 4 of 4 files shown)


internal/provider/cluster_resource.go
```diff
@@ -194,7 +194,7 @@ func (r *clusterResource) Schema(
 							stringplanmodifier.UseStateForUnknown(),
 						},
 						Validators:  []validator.String{stringvalidator.OneOf(AllowedUpgradeTypeTypeEnumValueStrings...)},
-						MarkdownDescription: "Dictates the behavior of cockroach major version upgrades. If plan type is 'BASIC', this attribute must be left empty or set to 'AUTOMATIC'. Allowed values are: " +
+						MarkdownDescription: "Dictates the behavior of cockroach major version upgrades. Manual upgrades are not supported on CockroachDB Basic. Manual or automatic upgrades are supported on CockroachDB Standard. If you omit the field, it defaults to `AUTOMATIC`. Allowed values are:" +
```
fantapop, line 197: @rgcase what do you think about specifying the default here?  We normally don't specify the defaults for serverside defaults but this one seems unlikely to change.

alyshanjahani-crl (approved):

internal/provider/cluster_resource.go
rgcase: Hmm do we have documentation somewhere that mentions the default? Seems like it would be better to link to that.  I prefer not hard-coding the default here but I don't feel too strongly about it so if you do then it's fine.

docs/resources/cluster.md
```diff
@@ -133,7 +133,7 @@ Read-Only:
 Optional:
 
 - `spend_limit` (Number, Deprecated) Spend limit in US cents.
-- `upgrade_type` (String) Dictates the behavior of cockroach major version upgrades. If plan type is 'BASIC', this attribute must be left empty or set to 'AUTOMATIC'. Allowed values are: 
+- `upgrade_type` (String) Dictates the behavior of cockroach major version upgrades. Manual upgrades are not supported on CockroachDB Basic. Manual or automatic upgrades are supported on CockroachDB Standard. If you omit the field, it defaults to `AUTOMATIC`. Allowed values are:
```
mdlinville, line 136:
```suggestion
- `upgrade_type` (String) Dictates the behavior of cockroachdb major version upgrades. Manual upgrades are not supported on CockroachDB Basic. Manual or automatic upgrades are supported on CockroachDB Standard. If you omit the field, it defaults to `AUTOMATIC`. Allowed values are:
```

internal/provider/cluster_resource.go
mdlinville, line 197:
```suggestion
						MarkdownDescription: "Dictates the behavior of cockroachdb major version upgrades. Manual upgrades are not supported on CockroachDB Basic. Manual or automatic upgrades are supported on CockroachDB Standard. If you omit the field, it defaults to `AUTOMATIC`. Allowed values are:" +
```

mdlinville: This attempts to cover the case where you don't set the field. In both plans, if you don't set the field, you get `AUTOMATIC`.

mdlinville (approved):

internal/provider/cluster_resource.go
fantapop: I believe what Ryan is suggesting is that, maybe we can link out to a docs page here.    We could say something like: 

If the field is omitted, a serverside default will be applied.  See [upgrade_type documentation ](url) for more details.

rgcase: Ya that's what I was suggesting.

@mdlinville the issue is that once this terraform provider is released, it's in the wild forever and we can guarantee that we'd be able to patch it. If we ever change our mind and have the server start setting a different default `upgrade_type` field in some cases, this won't be correct anymore.

My suggestion to link out to docs would be so that if we change the behavior one day we'll update the docs as well, and the user will always get the correct information about the default value.

mdlinville: It seems a bit circular since this thing you are wanting to have a link out from is itself literally the documentation for the API. So you're asking for the reader of the API spec to go from their IDE to their browser and then back to the IDE just to understand what happens with this field if they don't set it. It seems a bit much.

rgcase: This is the documentation for the Terraform provider which is a client of the API, not the API itself. The API should have its own documentation I think?

mdlinville: It seems so weird that there is this limitation to the Terraform registry. The only real place we can link out to on cockroachlabs.com is https://www.cockroachlabs.com/docs/cockroachcloud/provision-a-cluster-with-terraform. The actual best thing to do might be to link out instead to https://github.com/cockroachdb/terraform-provider-cockroach for the latest defaults. Then the Terraform registry doc reader can get to the latest details there.

rgcase: @fantapop I guess a change of the default value would really be a breaking API change, so we might want to cut a new version for it anyway, so we could guarantee that the default is always the same for this version of TF? If so then I guess we can just hard-code it.

marksoper: @fantapop and I were just chatting about this. I'm in favor of explicitly calling out the default values in the Terraform docs, especially for something like `upgrade_type` where we really want customers to choose AUTOMATIC unless they have a clear need for MANUAL - ideally we would say AUTOMATIC is the recommended, not just default, value. We don't want customers wondering about the right thing to do, whether they need to supply a value, etc. and run the risk of them putting in MANUAL when they aren't up to speed on the work they'll need to do to manage that.

I realize there may be other considerations, but wanted to stress that including defaults is the better customer experience.

fantapop: I just chatted with Ryan offline.  We agreed that we can specify the defaults in the documentation here because we should be cutting a new API version to preserve the old defaults if they change anyways.   It means we'll need to remember to update the documentation here if the default ever changes.

Also, I agree with Mark that the current wording doesn't provide very much guidance on what value the user should pick.  If they choose Manual, what ramifications does that have?  If they choose automatic, how often is that applied?  Could it cause an outage?  Do we have a recommendation?

PR #240, opened 2024-09-25
use the allowed upgrades list when validating version upgrades

dcrosta:
Beginning with v24.2, some releases are innovation releases which can be skipped during upgrading. The ListClusterMajorVersions API returns information on allowed upgrade paths, which we now use to validate whether an upgrade is possible in `apply`. Additional validation (cluster state, etc) is performed server-side within the API, so this validation is mostly to catch the most common type of error with a faster, more helpful error message, and not to be exhaustive.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)


internal/provider/cluster_resource.go
```diff
@@ -716,23 +701,64 @@ func (r *clusterResource) Update(
...
 			}
-			upgradeStatus = client.CLUSTERUPGRADESTATUSTYPE_ROLLBACK_RUNNING
-		} else if cmp > 1 {
-			resp.Diagnostics.AddError("Invalid upgrade version", "Can't skip versions. Upgrades must be performed one version at a time.")
+			if IsKnown(plan.UpgradeStatus) && plan.UpgradeStatus.ValueString() != string(client.CLUSTERUPGRADESTATUSTYPE_MAJOR_UPGRADE_RUNNING) {
+				resp.Diagnostics.AddError("Invalid upgrade_status",
+					fmt.Sprintf(
+						"cockroach_version ('%s') would upgrade cluster, but upgrade_status was '%s'. "+
+							"Remove upgrade_status when upgrading via cockroach_version.",
+						planVersion,
```
fantapop, line 735: I was somewhat surprised by this comment as it seems to be describing how upgrade_status in the api is used which should be an internal concern.  Then I was thinking that maybe I was misremembering how upgrades work in terraform so I had to review.  Note [here](https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/internal/provider/cluster_resource.go#L267-L270) that `upgrade_status` is marked as `Computed: true` and does not include `Optional: true,`.  This means they can't actually set this value in the plan.   IIRC, version upgrades in terraform just rely on the settable `cockroach_version` attribute.  if the version in the plan is later than the state, we try to upgrade.  If the version in the plan is earlier, we try to rollback.  This flow relies on the 72 hour auto finalization.  In order to finalize more quickly the `cockroach_finalize_version_upgrade` resource can be used.    Maybe I've misunderstood what's going on here.  Let's chat about it.  Also, side note, I've wanted to redesign how finalization works for some time now.  Maybe now is a good time to do it.  Or it could wait.

dcrosta: Ah, yeah, this was a holdover from my earlier misunderstanding about the `upgrade_status` field -- thanks for catching, I will remove this error checking block, which is moot, I think

Makefile
```diff
@@ -3,7 +3,7 @@ HOSTNAME=registry.terraform.io
 NAMESPACE=cockroachdb
 NAME=cockroach
 BINARY=terraform-provider-${NAME}
-VERSION=1.6.0
+VERSION=1.9.0
```
fantapop, line 6: Oops, Looks like I missed this last time.  Do you know what this controls?

CHANGELOG.md
```diff
@@ -5,8 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [1.9.0] - 2024-10-07
 
+- Added support for skipping [Innovation
+  Releases](https://www.cockroachlabs.com/docs/cockroachcloud/upgrade-policy#innovation-releases)
```
fantapop, line 11: fancy!

internal/provider/cluster_resource.go
```diff
@@ -716,23 +701,65 @@ func (r *clusterResource) Update(
...
+					validDowngrade = true
+					break
+				}
+			}
+			if !validDowngrade {
+				resp.Diagnostics.AddError("Invalid CockroachDB version",
+					fmt.Sprintf(
+						"Cannot roll back to to '%s'.",
+						planVersion))
+				return
```
fantapop, line 756: I was going to ask if we can list the valid rollback version here but... do we actually not know it at this point?  If we let this pass through, it seems like we can give a better error.  WDYT?

internal/provider/cluster_resource.go
```diff
@@ -971,6 +998,45 @@ func simplifyClusterVersion(version string, planSpecifiesPreviewString bool) str
...
+	}
+	toYear, toOrdinal, err := parseMajorVersion(toMajorVersion)
+	if err != nil {
+		return false, err
+	}
+	return toYear > fromYear || (toYear == fromYear && toOrdinal > fromOrdinal), nil
+}
+
+// isDowngrade returns true if toMajorVersion is older than fromMajorVersion
+func isDowngrade(fromMajorVersion, toMajorVersion string) (bool, error) {
```
fantapop, line 1028: Can you add some tests for isUpgrade and is Downgrade?

Makefile
dcrosta: Just the other rules in the Makefile here; it was helpful to have it be correct for `make install` when I was testing locally.

internal/provider/cluster_resource.go
dcrosta: We don't include the rollback version in the CC API clusters response. We could, there's probably no harm in doing so. Do you think that's valuable to add?

fantapop (approved):
nice, thank you

internal/provider/cluster_resource.go
dcrosta: Without all this error handling, we get:

```
cockroach_cluster.example: Modifying... [id=2dc9a589-1f20-412c-952b-522d5184a3f3]
cockroach_cluster.example: Still modifying... [id=2dc9a589-1f20-412c-952b-522d5184a3f3, 10s elapsed]
cockroach_cluster.example: Still modifying... [id=2dc9a589-1f20-412c-952b-522d5184a3f3, 20s elapsed]
╷
│ Error: Error updating cluster
│
│   with cockroach_cluster.example,
│   on main.tf line 49, in resource "cockroach_cluster" "example":
│   49: resource "cockroach_cluster" "example" {
│
│ cluster can only be rolled back to the previous version, expected: v24.1
```

I don't know why this took 20+ seconds to get this? But it seems like a reasonable error, so I'll remove the checks in the tf provider

fantapop: Here is the request: https://us5.datadoghq.com/apm/trace/8538d437c1868d2218b1a60d611750c1?graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=1655416649021547938&timeHint=1728402657004.241


350ms for the actual api request.    I think it probably took 20s because of all the other requests that happen during a tf run.  It looks like there are about 20 requests to the same cluster endpoint.  API caching via terraform isn't a good story currently.

Screenshot 2024-10-09 at 8 41 53 AM

PR #241, opened 2024-10-18
Add backup_config resource

fantapop:
Note that this commit is still a work in progress as the final SDK changes required for its use are not deployed yet but I've tested it locally and it's in a good place for feedback. 

@mdlinville I was particularly hoping to get your feedback on the documentation that I added.  This ended up being a documentation heavy resource.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)

I'm not super clear on how well the large block of documentation will render on the terraform docs site.  Here is the how the value looks with my local rendering.  Note: I've updated this screenshot after the latest batch of edits.

Screenshot 2024-10-22 at 1 18 53 PM


docs/resources/backup_config.md
```diff
@@ -0,0 +1,117 @@
...
+subcategory: ""
+description: |-
+  The backup configuration for a cluster.
+  Each cluster has a backup configuration that determines if backups are enabled,
+  how frequently they are taken and how often they are retained for.  This
+  resource allows configuring those settings.  It's important to note that the
+  existence or lack of this resource doesn't not indicate whether a backup
+  configuration exists or not. Rather, if Terraform is managing it or not.
+  As a result of this, removal of the resource from the Terraform configuration
+  will leave the backup configuration in the state it was before the removal.
```
kathancox, line 13: Does this mean, if you remove the backup_config tf resource, it just won't have any affect on the the configuration at all? If I have understood that correctly, perhaps it could be something like:

> As a result, removing the resource from the Terraform configuration will not affect the settings in the backup configuration.

docs/resources/backup_config.md
```diff
@@ -0,0 +1,117 @@
...
+  retention_days Edit Restrictions
+  retention_days can only be set once.  Further changes require opening a support
+  ticket.  For this reason, managing the value via terraform can require special
+  care. Here are some useful tips:
```
kathancox, line 17:
```suggestion
  `retention_days` can only be set once. It is necessary to open a support ticket to modify the settings again. For this reason, consider the following when managing the value in Terraform:
```

docs/resources/backup_config.md
```diff
@@ -0,0 +1,117 @@
...
+  * Optionally, you can refrain from including retention_days in the
+  terraform resource to fully rely on server side management of the value and not need any
+  special care.
```
kathancox, line 20:
```suggestion
  * (Optional) Refrain from including retention_days in the
  Terraform resource to rely on server-side management of the value instead.
```

docs/resources/backup_config.md
```diff
@@ -0,0 +1,117 @@
...
+  * If the initial value for retention_days in the backupconfig resource is the
```
kathancox, line 21:
```suggestion
  * If the initial value for `retention_days` in the `cockroach_backup_config` resource is the
```

docs/resources/backup_config.md
```diff
@@ -0,0 +1,117 @@
...
+  default value (30), the single edit will not be used up and it can be changed one
+  more time.
```
kathancox, line 23:
```suggestion
  default value (30), it will be possible to modify the retention setting one more time.
```

docs/resources/backup_config.md
```diff
@@ -0,0 +1,117 @@
...
+  * If the initial value set for retention_days in the backupconfig resource is
```
kathancox, line 24:
```suggestion
  * If the initial value set for `retention_days` in the `cockroach_backup_config` resource is
```

docs/resources/backup_config.md
```diff
@@ -0,0 +1,117 @@
...
+  not the default, the single edit will be used up on the first apply. Further
```
kathancox, line 25:
```suggestion
  not the default, it will not be possible to modify the retention setting again. Further
```

docs/resources/backup_config.md
```diff
@@ -0,0 +1,117 @@
...
+  edits will require a support ticket.
```
kathancox, line 26:
```suggestion
  modifications will require a support ticket.
```

docs/resources/backup_config.md
```diff
@@ -0,0 +1,117 @@
...
+  Changing the value of retention_days after using your one change will be a
+  multi-step operation. Here are two workflows that will work.  Both of these
+  paths assume you already have a backup config resource that doesn't have any
+  updates left:
+  Change it and then open a ticket before applying:
+  
+  Update retention_days to the new desired value in terraform.Before applying the run, contact support to change the retention_days to the desired value.Apply the changes in terraform, a terraform READ operation will be done and see the existing value and update the tf state.Temporarily remove management of it from tf, update it via ticket and then add it back.
+  
+  Remove management of retention_days from the terraform resourceRun the apply, nothing will change but terraform is no longer managing that value.Open a support ticket to update retention_days to the desired value.Optionally, add retention_days back to the terraform config with the new value and apply the no-op update.
```
kathancox, line 35:
```suggestion
  Changing the value of retention_days after using your one change will be a
  multi-step operation. Refer to the two following workflows. Both of these
  paths assume you already have a cockroach_backup_config resource that cannot be modified again. Change it and then open a ticket before applying:
  
  Update retention_days to the new value in Terraform. Before applying the run, contact support to change the retention_days to the new value. Apply the changes in Terraform, a Terraform READ operation will complete, recognize the existing value, and update the tf state. Temporarily remove management of it from tf, update it via ticket, and then add it back.
  
  Remove management of retention_days from the Terraform resource. Run the apply, nothing will change but Terraform is no longer managing that value. Open a support ticket to update retention_days to the new value. Optionally, add retention_days back to the Terraform config with the new value and apply the no-op update.
```

docs/resources/backup_config.md
```diff
@@ -0,0 +1,117 @@
...
+---
+
+# cockroach_backup_config (Resource)
+
+The backup configuration for a cluster.
```
kathancox, line 40:
```suggestion
The backup settings for a cluster.
```

In the docs, we are using "backup settings". I don't know if we need to keep this consistent, so this is a total nit.

docs/resources/backup_config.md
kathancox: Ah I see the duplicate content in the lower part of this file that is formatted in lists, I've added the language updates here to the latter part of the file.

kathancox (commented):
Added some comments for the doc component of this. I wasn't sure of the constraints here, so feel free to take the comments as you need. I can take another look when you need it.

docs/resources/backup_config.md
```diff
@@ -0,0 +1,117 @@
...
+  enabled = false
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `enabled` (Boolean) Indicates whether the backup configuration is enabled. If set to false, no backups will be created.
```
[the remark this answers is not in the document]
fantapop: yes it indicates if backups are enabled.  Is the edit here:

`Indicates whether backups are enabled.` ?

docs/resources/backup_config.md
fantapop: yes that's right. The suggestion sounds good to me.

docs/resources/backup_config.md
fantapop: Since this only applies to this one setting, I've taken the suggestion but updated it to:

modify the setting again.

docs/resources/backup_config.md
```diff
@@ -0,0 +1,117 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "cockroach_backup_config Resource - terraform-provider-cockroach"
+subcategory: ""
+description: |-
+  The backup configuration for a cluster.
+  Each cluster has a backup configuration that determines if backups are enabled,
+  how frequently they are taken and how often they are retained for.  This
+  resource allows configuring those settings.  It's important to note that the
```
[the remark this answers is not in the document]
fantapop: rereading this, I realized, 

and how often they are retained for

should be:

and how long they are retained for

docs/resources/backup_config.md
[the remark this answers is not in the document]
mdlinville: Drive by to say yes I think that's what she means, to remove the indirection of "the configuration" being enabled vs backups being enabled.

docs/resources/backup_config.md
fantapop: yes this is a bit confusing.  This md file is generated from the content in the resource.  The description in the frontmatter looks to be a stripped down version of the contents of the file.

fantapop:
@kathancox I made a round of changes.  You can see the diff here:  https://github.com/cockroachdb/terraform-provider-cockroach/compare/48c107eae8a9bdc446d998fda84c1b9d1daa4d93..5a50b9916d3b01d54a25b5ee51a051f2d4bc34c9

I think I've incorporated everything but I may have missed something.  Please feel free to make more comments if you have further edits.

kathancox (approved):
The back_config.md doc lgtm

fantapop:
I was just able to do some local testing of this and it's performing as expected.  One oddity which I called out in the behavior document I created is that when you remove the resource, no delete actually occurs.  It's just removing management of the backup configuration.  This may be slightly surprising to a user but It's probably an unlikely scenario to run into.  A terraform removal looks like this in the CLI.  

Screenshot 2024-10-22 at 5 03 08 PM

I think this is okay and is probably our best option here.  A couple other approaches I could think of:

1.  have backups disabled by default and have the lack of a resource indicate that they are disabled. This would mean that addition and removal of the config aligns with whether they are on or off. That seems undesirable because disabled by default is hostile to the user.
2. have backup configuration be a setting on the cluster resource. In this case, we're not adding or removing a resource in terraform which correlates better to backup  configuration always existing.  This is a potentially viable option and may be more natural for a user.  The 2 main drawbacks here are 1.  The cluster resource is quite complex already  2. I didn't consider this before the existing approach so migrating to the other approach would be more effort.

fantapop:
> 2. have backup configuration be a setting on the cluster resource. In this case, we're not adding or removing a resource in terraform which correlates better to backup  configuration always existing.  This is a potentially viable option and may be more natural for a user.  The 2 main drawbacks here are 1.  The cluster resource is quite complex already  2. I didn't consider this before the existing approach so migrating to the other approach would be more effort.

I've decided to dedicate a couple hours to exploring this.

fantapop:
closing in favor of https://github.com/cockroachdb/terraform-provider-cockroach/pull/242

PR #246, opened 2024-10-29
[CC-30315] rely on cluster api for validating parent_id

fantapop:
Previously, the cluster create and update operations would validate the parent_id manually via a separate API call.  This validation is also done within the create and update api calls so these additional calls were unnecessary.  We remove them here and add tests around setting this value which were previously missing.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/internal/provider/cluster_resource_test.go
+++ b/internal/provider/cluster_resource_test.go
@@ -184,16 +184,16 @@ var secondUpdatedBackupConfig = &client.BackupConfiguration{
 	FrequencyMinutes: 1440,
 }
 
-// TestAccBackupConfig is an acceptance test focused only on testing the
-// backup_config parameter. It will be skipped if TF_ACC isn't set.
+// TestAccClusterWithBackupConfig is an acceptance test focused only on testing
+// the backup_config parameter. It will be skipped if TF_ACC isn't set.
 func TestAccClusterWithBackupConfig(t *testing.T) {
 	t.Parallel()
 	clusterName := fmt.Sprintf("%s-serverless-%s", tfTestPrefix, GenerateRandomString(2))
 	testClusterWithBackupConfig(t, clusterName, false /* useMock */)
 }
 
-// TestAccBackupConfig is an acceptance test focused only on testing the
-// backup_config parameter. It will be skipped if TF_ACC isn't set.
+// TestIntegrationClusterWithBackupConfig is an integration test focused only on
+// testing the backup_config parameter.
 func TestIntegrationClusterWithBackupConfig(t *testing.T) {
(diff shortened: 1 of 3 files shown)


internal/provider/cluster_resource_test.go
```diff
@@ -470,6 +470,198 @@ func checkBackupConfig(clusterResourceName string, expected *client.BackupConfig
...
+func TestAccClusterWithParentID(t *testing.T) {
+	t.Parallel()
+	clusterName := fmt.Sprintf("%s-parent-c-%s", tfTestPrefix, GenerateRandomString(2))
+	folderName := fmt.Sprintf("%s-parent-f-%s", tfTestPrefix, GenerateRandomString(2))
+	testClusterWithParentID(t, clusterName, folderName, false /* useMock */)
+}
+
+// TestIntegrationClusterWithParentID is an integration test focused only on
+// testing the parent_id parameter.
+func TestIntegrationClusterWithParentID(t *testing.T) {
```
rgcase, line 486: Do you think we need a whole separate test for this? Now the `parent_id` field should behave just like the other fields on the cluster so it seems like it could be covered in a more general test?

fantapop: parent_id does have some additional complexity such as validation that wasn't previously being tested.  For example the field should accept either "root" or something that matches a uuid.   Also, there is a test showing that the error message for a not found folder is reasonable.  

It felt easier and less complex to have a set of tests dedicated only to this field.  

I think if I were to add the tests into the existing framework that exists for testing the serverless clusters it would go here for integration tests:  https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/internal/provider/cluster_resource_test.go#L796-L819

and it would go somewhere here for acceptance tests: https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/internal/provider/cluster_resource_test.go#L71-L75

Neither of these areas seem ideal to inject a set of serial checks on the parent_id and it looks like I'd need two different versions of the tests to cover them across integration and acceptance.

I think your overall point here is that hopefully we don't need a dedicated set of tests for each new field because that could explode the amount of total test code for this resource.   If you feel strongly about this specific case, let's talk offline about it.

rgcase: > I think your overall point here is that hopefully we don't need a dedicated set of tests for each new field because that could explode the amount of total test code for this resource. If you feel strongly about this specific case, let's talk offline about it.

Ya right this was where I was getting at. I think you've convinced me that this field is significantly different though so I'm okay with this, thanks.

PR #248, opened 2024-11-14
Add a github action for triggering acceptance tests

fantapop:
This action is very similar to the pre-release-ci action which runs when a tag is created, but it can be triggered manually on any branch. This allow us to easily run the acceptance tests in other convenient times such as before landing a branch or before creating a tag.

**Commit checklist**
- [ ] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)


rgcase (approved):
LGTM

PR #252, opened 2024-12-03
[CC-30640] allow manual version control for standard clusters

fantapop:
STANDARD clusters now accept the cockroach_version attribute on cluster update for manual upgrades control.  The serverless.upgrade_type attribute must be set to MANUAL.

As part of this an API call to validate cockraoch_version for ADVANCED clusters prior to the update was removed.  This logic provided some nicer error messages for advanced clusters but the logic was complex, it required an additional api call to fetch the valid versions and standard tier version checking is not supported.  We now rely on the serverside validation for checking versions.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/internal/provider/cluster_resource.go
+++ b/internal/provider/cluster_resource.go
@@ -23,7 +23,6 @@ import (
 	"regexp"
 	"sort"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/cockroachdb/cockroach-cloud-sdk-go/v5/pkg/client"
@@ -123,7 +122,7 @@ func (r *clusterResource) Schema(
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.UseStateForUnknown(),
 				},
-				Description: "Major version of CockroachDB running on the cluster.",
+				MarkdownDescription: "Major version of CockroachDB running on the cluster. This value can be used to orchestrate version upgrades. Supported for ADVANCED and STANDARD clusters (when `serverless.upgrade_type` set to 'MANUAL').",
 			},
 			"account_id": schema.StringAttribute{
 				Computed: true,
@@ -377,6 +376,27 @@ func (r *clusterResource) ConfigValidators(_ context.Context) []resource.ConfigV
 	}
 }
 
+func derivePlanType(cluster *CockroachCluster) (client.PlanType, error) {
+	var planType client.PlanType
+	if IsKnown(cluster.Plan) {
+		planType = client.PlanType(cluster.Plan.ValueString())
+	} else if cluster.DedicatedConfig != nil {
+		planType = client.PLANTYPE_ADVANCED
+	} else if cluster.ServerlessConfig != nil {
+		if cluster.ServerlessConfig.UsageLimits != nil && IsKnown(cluster.ServerlessConfig.UsageLimits.ProvisionedVirtualCpus) {
+			planType = client.PLANTYPE_STANDARD
+		} else {
+			planType = client.PLANTYPE_BASIC
+		}
+	} else {
+		return "", fmt.Errorf("could not derive plan type, plan must contain either a ServerlessConfig or a DedicatedConfig")
+	}
+	if !planType.IsValid() {
+		return "", fmt.Errorf("invalid plan type %q", cluster.Plan.ValueString())
+	}
+	return planType, nil
+}
+
 func (r *clusterResource) Create(
 	ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse,
 ) {
@@ -436,6 +456,20 @@ func (r *clusterResource) Create(
 				resp.Diagnostics.AddError("Invalid Upgrade Type", err.Error())
 			} else {
 				serverless.SetUpgradeType(*upgradeType)
+			}
+		}
+
+		if IsKnown(plan.CockroachVersion) {
+			planType, err := derivePlanType(&plan)
+			if err != nil {
+				resp.Diagnostics.AddError("Invalid Configuration", err.Error())
+				return
+			}
+
+			if planType == client.PLANTYPE_BASIC {
+				resp.Diagnostics.AddError("Invalid Attribute Combination", "cockroach_version is not supported for BASIC clusters.")
+			} else {
+				resp.Diagnostics.AddError("Unsupported Attribute", "cockroach_version is not supported during cluster creation for STANDARD clusters. STANDARD clusters are automatically created using the latest available version.")
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


internal/provider/cluster_resource_test.go
```diff
@@ -1219,6 +1414,8 @@ func onDemandSingleRegionClusterWithLimitsStep(
 	planType client.PlanType,
 	requestUnitLimit int64,
 	storageMibLimit int64,
+	upgradeType *client.UpgradeTypeType,
```
fantapop, line 1417: this pattern of adding new fields for each type that needs to be controlled on occasion doesn't scale well.  I'm open to suggestions here.   One thing I was considering is using a config object pattern similar to what is done for backup configuration: https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/internal/provider/cluster_resource_test.go#L1912-L1917

internal/provider/cluster_resource.go
```diff
@@ -123,7 +122,7 @@ func (r *clusterResource) Schema(
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.UseStateForUnknown(),
 				},
-				Description: "Major version of CockroachDB running on the cluster.",
+				MarkdownDescription: "Major version of CockroachDB running on the cluster. This value can be used to orchestrate version upgrades. Supported for ADVANCED and STANDARD clusters (when `serverless.upgrade_type` set to 'MANUAL').",
```
fantapop, line 125: @kathancox thanks in advanced for looking over this. I'm not sure it's enough or whether we have other dedicated documentation on updating the cockroach version with terraform.  I could link out to an external doc if we have one.

internal/provider/cluster_resource.go
```diff
@@ -439,6 +438,23 @@ func (r *clusterResource) Create(
...
+			if clusterSpec.Plan != nil {
+				planType = *clusterSpec.Plan
+			} else if serverless.UsageLimits != nil && serverless.UsageLimits.ProvisionedVirtualCpus != nil {
+				planType = client.PLANTYPE_STANDARD
+			} else {
+				planType = client.PLANTYPE_BASIC
+			}
+
+			if planType == client.PLANTYPE_BASIC {
+				resp.Diagnostics.AddError("Invalid Attribute Combination", "cockroach_version is not supported for BASIC clusters.")
```
fantapop, line 452: Also, @kathancox you'll find quite a few errors strings that could use a docs review in this file.  They should all look something like this one.  Just look for `AddError`.  If you want I can tag you on each one.

internal/provider/cluster_resource_test.go
```diff
@@ -975,6 +982,17 @@ func TestIntegrationServerlessClusterResource(t *testing.T) {
 			},
 			finalCluster: singleRegionClusterWithUnlimited("BASIC"),
 		},
+		{
+			name: "upgrade cockroach_version for STANDARD cluster",
```
fantapop, line 986: This is the only real automated test for a successful standard tier upgrade.  Our api doesn't allow setting up these tests since we need to be able to build a host cluster and then upgrade it.  For that reason, we're currently limited to integration tests here.   Given the complication of upgrades it would be great to figure out a way to build an acceptance test for this.  @andy-kimball Do you have any ideas for that?

internal/provider/cluster_resource.go
```diff
@@ -439,6 +438,23 @@ func (r *clusterResource) Create(
 			}
 		}
 
+		if IsKnown(plan.CockroachVersion) {
+			var planType client.PlanType
+			if clusterSpec.Plan != nil {
```
andy-kimball, line 443: This plan type derivation logic seems like something that should be a utility function.

internal/provider/cluster_resource.go
```diff
@@ -788,66 +804,45 @@ func (r *clusterResource) Update(
...
-		// current version, according to the ListMajorClusterVersions API.
-		if upgrading, err := isUpgrade(stateVersion, planVersion); upgrading {
-			var currentVersionInfo client.ClusterMajorVersion
-			for _, v := range apiResp.Versions {
-				if v.Version == stateVersion {
-					currentVersionInfo = v
-					break
+		if state.ServerlessConfig != nil {
+			stateUpgradeTypeString := state.ServerlessConfig.UpgradeType.ValueString()
+			planUpgradeTypeString := plan.ServerlessConfig.UpgradeType.ValueString()
```
andy-kimball, line 826: Is it guaranteed that `plan.ServerlessConfig` will be non-nil here?

internal/provider/cluster_resource.go
```diff
@@ -788,66 +804,45 @@ func (r *clusterResource) Update(
...
+			if stateUpgradeTypeString != planUpgradeTypeString {
+				resp.Diagnostics.AddError("Error updating cluster", "upgrade_type and version must not be changed during the same plan.")
+				return
+			} else {
```
andy-kimball, line 830: NIT: I'd just get rid of the `else` since you're already returning from the `if`.

internal/provider/cluster_resource_test.go
andy-kimball: @alyshanjahani-crl, do we support creating a Serverless cluster at the previous version? If so, we could do that, and then upgrade it to the next version, all without touching the host cluster version.

internal/provider/cluster_resource_test.go
andy-kimball: Sure, having a struct might be nice, where we can have default values that don't need to be set on every call.

kathancox (approved):
LGTM! One tiny nit, and a small comment/question.

internal/provider/cluster_resource_test.go
fantapop: We're currently limited creating serverless clusters at the latest version. This is why cockroach_version is not accepted for serverless clusters during the Create phase.  But some kind of ability in the future would be welcome.  We could hide it behind a feature flag if need be.

internal/provider/cluster_resource.go
```diff
@@ -788,66 +804,45 @@ func (r *clusterResource) Update(
...
+				stateUpgradeTypePtr, err := client.NewUpgradeTypeTypeFromValue(stateUpgradeTypeString)
+				if err != nil {
+					resp.Diagnostics.AddError("Invalid value for upgrade_type found in state", err.Error())
```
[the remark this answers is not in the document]
fantapop: looking closer at this, I noticed this is the "error summary" field.  We generally don't add periods for the summary.  But it does seem like this summary is probably too specific.  Here is what would be returned by `err.Error()`:  

```
"invalid value 'BAD_VALUE' for UpgradeTypeType: valid values are ['AUTOMATIC', 'MANUAL']
```

So the error would should up as:

"Invalid value for upgrade_type found in state"
"invalid value 'BAD_VALUE' for UpgradeTypeType: valid values are ['AUTOMATIC', 'MANUAL']

This is repetitive.  I'll update the summary to be "Error updating cluster" like the rest of the errors in this file.   It would be nice to eventually do an audit of the full provider and make a strategy for what to use for the summaries.

internal/provider/cluster_resource.go
```diff
@@ -788,66 +804,45 @@ func (r *clusterResource) Update(
...
+		// Unfortunately the update cluster api endpoint doesn't allow updating
+		// the version and anything else in the same call. Since version may
+		// rely on upgrade_type and plan, if they are updated at the same time
+		// we need to fail out.
+		// We could explore doing the version upgrade after the other updates to
+		// avoid this 2 step apply requirement.
+
+		if client.PlanType(state.Plan.ValueString()) == client.PLANTYPE_BASIC {
+			if client.PlanType(plan.Plan.ValueString()) != client.PLANTYPE_BASIC {
+				resp.Diagnostics.AddError("Error updating cluster", "plan and version must not be changed during the same plan.")
```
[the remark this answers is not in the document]
fantapop: The plan I'm referring to is basically a batch of updates. Not to be confused with the cluster plan (i.e. basic, standard, advanced).

With terraform you make changes to your configuration, then you can run a "plan" in which terraform will spit out what the changes will be.  Or you can run an "apply", which includes a "plan" but then asks if the plan looks good before proceeding with the actually update.

If this is a point of confusion I'm happy to change it.  We could say something like "must not be changed at the same time"  or "... at once" or, "please apply changes to the upgrade_type before upgrading", or many other things...

internal/provider/cluster_resource.go
fantapop: Okay, I've moved it out to its own function and added a test for it.

internal/provider/cluster_resource.go
fantapop: I didn't know this so I wrote a test to try it out.  Apparently we're protected against this [here](https://github.com/cockroachdb/terraform-provider-cockroach/blob/3f8d115a1056c6fa8dde59600e7b8dea1f122518/internal/provider/cluster_resource.go#L700-L707).  ModifyPlan is run before the update should it should never be the case that `state.ServerlessConfig` is set but `plan.ServerlessConfig` is not.

internal/provider/cluster_resource_test.go
fantapop: Okay, I've done some refactoring.  Let me know what you think.  I added it to a [later commit](https://github.com/cockroachdb/terraform-provider-cockroach/pull/252/commits/e95da011a8916523fa22648b85bdaa1bb75d6eee).

I could make further convenience functions to save additional space. I was thinking about something like:

```go
func provisionedClusterStep(
	clusterName string,
	config slsConfig,
) resource.TestStep {
    config.vcpus = ptr(6)
    return serverlessClusterStep(clusterName, client.PLANTYPE_STANDARD, config)
}
```

andy-kimball: The changes LGTM.

internal/provider/cluster_resource.go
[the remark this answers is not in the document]
kathancox: Great thanks!

internal/provider/cluster_resource.go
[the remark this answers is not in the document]
kathancox: I forgot about the `plan` command and the concept generally within Terraform. I think `must not be changed at the same time`, or maybe the original is fine. I kind of like `plan and version must not be changed during the same terraform plan.` But maybe that is too heavy handed. I'll leave it to you.

internal/provider/cluster_resource.go
```diff
@@ -377,6 +376,27 @@ func (r *clusterResource) ConfigValidators(_ context.Context) []resource.ConfigV
...
 }
 
+func derivePlanType(cluster *CockroachCluster) (client.PlanType, error) {
+	var planType client.PlanType
+	if IsKnown(cluster.Plan) {
+		planType = client.PlanType(cluster.Plan.ValueString())
+	} else if cluster.DedicatedConfig != nil {
+		planType = client.PLANTYPE_ADVANCED
+	} else if cluster.ServerlessConfig != nil {
+		if cluster.ServerlessConfig.UsageLimits != nil && IsKnown(cluster.ServerlessConfig.UsageLimits.ProvisionedVirtualCpus) {
```
rgcase, line 386: Is `IsKnown` the same as non-zero? Do we reject specs that set a usage limit and also explicitly set `provisioned_virtual_cpus == 0`?

rgcase (approved):
LGTM

internal/provider/cluster_resource.go
rgcase: Oops I meant to delete this. I think the `ConfigValidators` on the cluster resource guarantee that they can't be set together.

PR #254, opened 2024-12-06
[CC-30766] disallow 0 as a value for disk_iops

fantapop:
0 is a valid value to pass into the api for disk_iops.  That, along with omitting the value in the api call results in the default iops value.  However, the api returns the actual value for the iops that was chosen.  This results in a         configuration conflict with the set value of 0 and the actual value that was returned from the server.  For this reason we now disallow 0 from being set and we clarify in the docs that omitting the value is the correct way to ask for the      default.

**Commit checklist**
- [x] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)


fantapop:
This looks big because it includes a vendor change that resulted in a lot of updates. The commit to review is this one: 
6052f571cd5d73ed34b0521dff0aa23d521e59a1

docs/resources/cluster.md
```diff
@@ -139,7 +139,7 @@ Optional:
 Optional:
 
 - `cidr_range` (String) The IPv4 range in CIDR format that will be used by the cluster. This is supported only on GCP, and must have a subnet mask no larger than /19. Defaults to "172.28.0.0/14". This cannot be changed after cluster creation.
-- `disk_iops` (Number) Number of disk I/O operations per second that are permitted on each node in the cluster. Zero indicates the cloud provider-specific default.
+- `disk_iops` (Number) Number of disk I/O operations per second that are permitted on each node in the cluster. Omitting this attribute will result in the cloud provider-specific default.
```
fantapop, line 142: @kathancox here is the doc change I was hoping to get your eyes on.

internal/provider/cluster_resource.go
```diff
@@ -218,7 +219,13 @@ func (r *clusterResource) Schema(
 					"disk_iops": schema.Int64Attribute{
 						Optional:    true,
 						Computed:    true,
-						Description: "Number of disk I/O operations per second that are permitted on each node in the cluster. Zero indicates the cloud provider-specific default.",
+						Validators:  []validator.Int64{
+							// If supplied this value must be non-zero. 0 is a
+							// valid api value indicating the default being
+							// returned but it causes a provider inconsistency.
+							int64validator.AtLeast(1),
```
rgcase, line 226: Hmm this is a breaking change if anyone is actually setting this right? Is it really important to change this? It seems like anyone setting zero today knows what they're getting. I don't feel that strongly about it though so if you want to merge it then go for it.

rgcase (approved):

kathancox (approved):

internal/provider/cluster_resource.go
fantapop: As discussed offline, I don't think this is a breaking change because using 0 as was previously mentioned in the doc causes an error.  When 0 is written to the tf configuration, it is then passed via the api.  The api returns the actual (non-zero) default depending on the disk size.  The 0 in the configuration causes a state inconsistency error with the returned value.  I just noticed the wrong commit message was added to this PR since I broke out some vendor additions to a separate commit.  I'll fix the PR message. Hopefully that helps clear this point up.

PR #255, opened 2024-12-07
[CC-30768] don't allow node_count for serverless cluster

fantapop:
Previously it was possible to pass in the node_count value for serverless region because the type is shared with dedicated clusters. This ended up causing a state inconsistency however becase node_count is coerced to 0 by the api which conflicts with a value added in the config. To solve this, we simply validate that node_count is not used unless cluster plan is Advanced.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -113,7 +113,7 @@ Required:
 
 Optional:
 
-- `node_count` (Number) Number of nodes in the region. Will always be 0 for serverless clusters.
+- `node_count` (Number) Number of nodes in the region. Valid for Advanced clusters only.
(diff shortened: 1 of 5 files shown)


docs/resources/cluster.md
```diff
@@ -113,7 +113,7 @@ Required:
 
 Optional:
 
-- `node_count` (Number) Number of nodes in the region. Will always be 0 for serverless clusters.
+- `node_count` (Number) Number of nodes in the region. Valid for Advance clusters only.
```
fantapop, line 116: @kathancox this one should be pretty straightforward.  Just this line to check. Thank you!

rgcase (approved):

kathancox (approved):

PR #265, opened 2025-02-12
Fix prometheus metric export test

aa-joshi:
**Commit checklist**
- [ ] Changelog
- [x] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/internal/provider/metric_export_prometheus_config_resource_test.go
+++ b/internal/provider/metric_export_prometheus_config_resource_test.go
@@ -56,7 +56,7 @@ func TestIntegrationMetricExportPrometheusConfigResource(t *testing.T) {
 		Regions: []client.Region{
 			{
 				Name:      "us-east1",
-				NodeCount: 3,
+				NodeCount: 1,
 			},
 		},
 	}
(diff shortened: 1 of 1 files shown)


internal/provider/metric_export_prometheus_config_resource_test.go
```diff
@@ -85,21 +80,7 @@ func TestIntegrationMetricExportPrometheusConfigResource(t *testing.T) {
 		Return(createdPrometheusClusterInfo, nil, nil)
 	s.EXPECT().GetPrometheusMetricExportInfo(gomock.Any(), clusterID).
 		Return(createdPrometheusClusterInfo, nil, nil).
-		Times(3)
-
-	// Update
-	s.EXPECT().GetCluster(gomock.Any(), clusterID).
```
aa-joshi, line 91: We removed tf update validation as prometheus metric export integration doens't need update flow. It is either enable and disable.

internal/provider/metric_export_prometheus_config_resource_test.go
```diff
@@ -201,28 +175,7 @@ resource "cockroach_cluster" "test" {
...
-  name           = "%s"
-  cloud_provider = "GCP"
-  dedicated = {
-    storage_gib = 35
-  	num_virtual_cpus = 4
-  }
-  regions = [{
-    name = "us-east1"
-    node_count: 3
+    node_count: 1
```
aa-joshi, line 178: The node count is reduced to 1 to reduce the cost of additional node provisioning.

PR #266, opened 2025-02-27
remove plan modifier from state parameter in cluster resource

linhcrl:
This commit removes the UseStateForUnknown plan modifier from the cluster resource's state attribute. This change resolves inconsistent results after apply caused by mismatched values between the state and plan.

**Commit checklist**
- [x] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.11.3] - 2025-02-27
+
+### Fixed
+
+- Remove UseStateForUnknown plan modifier for state parameter in cluster resource.
+
 ## [1.11.2] - 2025-02-07
 
 ### Fixed
--- a/internal/provider/cluster_resource.go
+++ b/internal/provider/cluster_resource.go
@@ -217,9 +217,9 @@ func (r *clusterResource) Schema(
 						Description: "Storage amount per node in GiB.",
 					},
 					"disk_iops": schema.Int64Attribute{
-						Optional:    true,
-						Computed:    true,
-						Validators:  []validator.Int64{
+						Optional: true,
+						Computed: true,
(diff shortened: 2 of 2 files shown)


linhcrl:
@fantapop my IDE applied some formatting changes to the file. The only actual changes I made was removing the plan modifier. 

Please note I tried to repro using tests but couldn't really figure out how to do it. 

However, I was able to test _locally_. I followed the steps to reproduce the error before the change (removal of plan modifier) and then I verified that after the change, we no longer received the initial error.

fantapop (commented):
Did you try adding / removing different variations of the backup_config?  IIRC the reason I ended up using the UseStateForUnknown modifier was due to plan churn.

linhcrl:
> Did you try adding / removing different variations of the backup_config? IIRC the reason I ended up using the UseStateForUnknown modifier was due to plan churn.

Oh to be clear I only removed the plan modifier from the _state_ attribute because that's the one that caused this specific bug/error and it's the only one I know how to reproduce and test locally.

 I didn't touch backup configs at all. Should I remove it from those fields as well?

fantapop:
> > Did you try adding / removing different variations of the backup_config? IIRC the reason I ended up using the UseStateForUnknown modifier was due to plan churn.
> 
> @fantapop Oh to be clear I only removed the plan modifier from the _state_ attribute because that's the one that caused this specific bug/error and it's the only one I know how to reproduce and test locally.
> 
> I didn't touch backup configs at all. Should I remove it from those fields as well?

Sorry, I missed this comment.  I wasn't trying to suggest we remove it there.  Just that we test to ensure there wasn't churn from various backup config states.  The change for updating state to unknown was in the same PR as the other and I started making those changes due to churn.

linhcrl:
After further thought, I don't think this change is necessary. We initially thought of doing this because on terraform apply, after a  field got changed externally, terraform reported a difference between the value stored in the state and the value returned by the API. 

I have worked on a separate bug where the field didn't have a plan modifier and it still errored with the same message, so this change in the PR likely won't make a difference. I'm going to close this, as the API fix we put in a while ago solved the original issue.

PR #270, opened 2025-04-08
[CC-30705] add labels attribute to cluster and folder resources

linhcrl:
This commit adds support for managing labels through the cockroach_cluster and cockroach_folder resources in the Cockroach Terraform provider. Labels defined in Terraform are authoritative and will overwrite any labels added or removed outside of Terraform.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.12.0] - 2025-04-30
+
 ### Added
+
+- Added support for label management in the `cockroach_cluster` and
+  `cockroach_folder` resources.
 
 - Added `full_version` attribute to the cluster data source and resource for
   fetching the full version string. (e.g. v25.1.0)
--- a/internal/provider/cluster_resource.go
+++ b/internal/provider/cluster_resource.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/cockroachdb/cockroach-cloud-sdk-go/v6/pkg/client"
+	"github.com/cockroachdb/terraform-provider-cockroach/internal/utils"
 	"github.com/cockroachdb/terraform-provider-cockroach/internal/validators"
 	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/resourcevalidator"
@@ -339,6 +340,13 @@ func (r *clusterResource) Schema(
 					},
 				},
 			},
+			"labels": schema.MapAttribute{
+				Computed:    true,
+				Optional:    true,
+				ElementType: types.StringType,
+				Description: "Map of key-value pairs used to organize and categorize resources. If unset, labels will not be managed by Terraform. If set, labels defined in Terraform will overwrite any labels configured outside this platform.",
+				Validators:  labelsValidator,
+			},
 		},
 	}
 }
@@ -387,7 +395,9 @@ func (r *clusterResource) ConfigValidators(_ context.Context) []resource.ConfigV
 	}
 }
 
-func (r *clusterResource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+func (r *clusterResource) ValidateConfig(
+	ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse,
+) {
 
 	// ValidateConfig seems to be called multiple times, once before any values
 	// are known.  When all values are unknown we can't convert to our cluster
@@ -573,6 +583,17 @@ func (r *clusterResource) Create(
 		clusterSpec.SetParentId(parentID)
 	}
 
+	if IsKnown(plan.Labels) {
+		labels, err := utils.ToStringMap(plan.Labels)
+		if err != nil {
+			resp.Diagnostics.AddError(
+				"Error processing cluster labels",
+				fmt.Sprintf("Could not convert cluster labels: %v", err),
+			)
+		}
+		clusterSpec.SetLabels(labels)
+	}
+
 	deleteProtection := client.DELETEPROTECTIONSTATETYPE_DISABLED
 	if plan.DeleteProtection.ValueBool() {
 		deleteProtection = client.DELETEPROTECTIONSTATETYPE_ENABLED
@@ -604,7 +625,7 @@ func (r *clusterResource) Create(
 	// configuration so it will be there no matter what in case of failure.
 	// The backup configuration can be synced in a later apply if need be.
 	var newState CockroachCluster
-	diags = loadClusterToTerraformState(clusterObj, nil, &newState, &plan)
+	diags = loadClusterToTerraformState(ctx, clusterObj, nil, &newState, &plan)
 	resp.Diagnostics.Append(diags...)
 	diags = resp.State.Set(ctx, newState)
 	resp.Diagnostics.Append(diags...)
@@ -662,7 +683,7 @@ func (r *clusterResource) Create(
 		}
 	}
 
-	diags = loadClusterToTerraformState(clusterObj, remoteBackupConfig, &newState, &plan)
+	diags = loadClusterToTerraformState(ctx, clusterObj, remoteBackupConfig, &newState, &plan)
 	resp.Diagnostics.Append(diags...)
 	diags = resp.State.Set(ctx, newState)
 	resp.Diagnostics.Append(diags...)
@@ -734,7 +755,7 @@ func (r *clusterResource) Read(
 	// We actually want to use the current state as the plan here, since we're
 	// trying to see if it changed.
 	var newState CockroachCluster
-	diags = loadClusterToTerraformState(clusterObj, remoteBackupConfig, &newState, &state)
+	diags = loadClusterToTerraformState(ctx, clusterObj, remoteBackupConfig, &newState, &state)
 	resp.Diagnostics.Append(diags...)
 	diags = resp.State.Set(ctx, newState)
 	resp.Diagnostics.Append(diags...)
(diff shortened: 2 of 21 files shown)


mock/service.go
```diff
@@ -293,10 +293,10 @@ func (mr *MockServiceMockRecorder) CreateServiceAccount(arg0, arg1 interface{})
 }
```
linhcrl, line 293: This will go into the PR that updates the SDK. Added it to this PR to have a record of what changes need to be made..

internal/utils/utils.go
linhcrl, line 1: Needed to add this package because putting it in the `providers/utils.go` file caused an import cycle

mock/service.go
linhcrl, line 1: This will go into the PR that updates the SDK. Added it to this PR to have a record of what changes need to be made..

internal/provider/cluster_resource_test.go
```diff
@@ -2434,3 +2434,289 @@ func TestDerivePlanType(t *testing.T) {
...
+				},
+				Config:      getTestClusterWithLabels(folderName, clusterName, labelsAboveLimit, false /*moveToFolder*/),
+				ExpectError: regexp.MustCompile("Attribute labels must contain at most 50 key-value pairs, with each key"),
+			},
+			{
+				PreConfig: func() {
+					traceMessageStep("create cluster with invalid labels")
+				},
+				Config:      getTestClusterWithLabels(folderName, clusterName, invalidLabels, false /*moveToFolder*/),
+				ExpectError: regexp.MustCompile("Attribute labels must contain at most 50 key-value pairs, with each key"),
```
linhcrl, line 2644: Couldn't figure out how to match the entire string because the validator logs the validation error message in multiple lines. So I decided to just match against the beginning of the message.

internal/provider/cluster_resource_test.go
```diff
@@ -2434,3 +2434,289 @@ func TestDerivePlanType(t *testing.T) {
...
+					traceMessageStep("move cluster and leave labels unchanged")
+				},
+				Config: getTestClusterWithLabels(folderName, clusterName, newLabels, true /*moveToFolder*/),
+				Check:  testCheckLabels(serverlessResourceName, newLabels),
+			},
+		},
+	})
+}
+
+func getTestClusterWithLabels(folderName, clusterName string, labels map[string]string, moveToFolder bool) string {
```
fantapop, line 2685: I think you should make labels a pointer here and incorporate some tests where you stop passing a labels object into the resource.  If labels is excluded, it means we shouldn't clear it.  If labels is passed  but empty, we should clear it.

internal/provider/folder_resource.go
```diff
@@ -47,6 +49,15 @@ func (r *folderResource) Schema(
...
 				},
 			},
+			"labels": schema.MapAttribute{
+				Computed:    true,
+				Optional:    true,
+				ElementType: types.StringType,
+				Description: "Map of key-value pairs used to organize and categorize resources.",
+				Validators: []validator.Map{
+					validators.Labels(),
+				},
```
fantapop, line 59: This syntax is slightly heavy.  I think including everything in a single define is kind of nice.  

take a look at:

https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/internal/provider/api_key_resource.go#L46

this could be something like: 
```go
Validators: labelsValidators,
```
instead

internal/provider/folder_resource_test.go
```diff
@@ -204,11 +264,16 @@ func getTestFolderResourceUpdateConfig(parentFolderName, childFolderName string)
...
 	parent_id = "root"
+	labels = {}
 }
 
 resource "cockroach_folder" "test_folder_child" {
     name = "%s"
 	parent_id = "root"
+	labels = {
+        managed_by = "engineer_b"
+        campaign = "87654"
```
fantapop, line 275: nit: indentation here would be nice

internal/utils/utils.go
fantapop: utils/utils.go is pretty henious.  Do you know what it was that was causing the import cycle you're not importing much...

fantapop (commented):
nice start on this!  It seems pretty close.

internal/utils/utils.go
linhcrl: because I'm calling the helper `ToStringMap()` from this file in the `validators` package. And the `validators` package is used in the `providers` package and the original utils.go file is in the `providers` package

internal/provider/cluster_resource_test.go
linhcrl: Done! we now have one test case that clears labels and one that doesn't pass labels and thus leaves it unchanged

internal/provider/folder_resource.go
linhcrl: added a `labelsValidator` var in `/providers/utils.go` (which is where `uuidValidator` is defined)

internal/provider/folder_resource_test.go
linhcrl: did not see this in my IDE, thanks for pointing out

internal/provider/cluster_resource.go
```diff
@@ -1266,6 +1298,16 @@ func loadClusterToTerraformState(
 		state.ParentId = types.StringValue(*clusterObj.ParentId)
 	}
 
+	var allDiags diag.Diagnostics
+
+	if clusterObj.Labels == nil {
+		state.Labels = types.MapNull(types.StringType)
```
fantapop, line 1304: It feels like we should be use types.MapValue here with an empty map.  What do you think?

internal/provider/cluster_resource_test.go
```diff
@@ -19,6 +19,7 @@ package provider
 import (
 	"context"
 	"fmt"
+	"github.com/cockroachdb/terraform-provider-cockroach/internal/validators"
```
fantapop, line 22: i'm surprised this is sorting here.  I guess we're not running any kind of formatter on this code...  we should consider doing that at some point...

internal/provider/cluster_resource_test.go
```diff
@@ -2432,3 +2433,303 @@ func TestDerivePlanType(t *testing.T) {
...
+				},
+				Config:      getTestClusterWithLabels(folderName, clusterName, &labelsAboveLimit, false /*moveToFolder*/),
+				ExpectError: regexp.MustCompile("Attribute labels must contain at most 50 key-value pairs, with each key"),
+			},
+			{
+				PreConfig: func() {
+					traceMessageStep("create cluster with invalid labels")
+				},
+				Config:      getTestClusterWithLabels(folderName, clusterName, &invalidLabels, false /*moveToFolder*/),
+				ExpectError: regexp.MustCompile("Attribute labels must contain at most 50 key-value pairs, with each key"),
```
fantapop, line 2648: Is this the right error message?

internal/provider/folder_resource.go
```diff
@@ -47,6 +49,13 @@ func (r *folderResource) Schema(
 					validators.FolderParentID(),
 				},
 			},
+			"labels": schema.MapAttribute{
+				Computed:    true,
+				Optional:    true,
+				ElementType: types.StringType,
+				Description: "Map of key-value pairs used to organize and categorize resources.",
```
fantapop, line 56: I think we should consider including in the description what happens if labels is not included.   Something like... "if unset, labels will not be managed by terraform."  although that's kind of a mouthful.  I wonder if we should have a separate section describing how labels works in terraform.   We have [this guide](https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/docs/guides/updating-backup-retention.md) about backup config.  We could have something similar.

For example we may to detail that setting labels via terraform is authoritative and will overwrite any labels set on resources set in the UI.

Maybe there's not enough info to warrant that.   Eventually if we had non-authoritative usages or support he lifecycle methods like aws mentions [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/resource-tagging#ignoring-changes-in-individual-resources), maybe there will be more to talk about. 

We should have @mikeCRL look at these docs strings as well.

internal/provider/folder_resource_test.go
```diff
@@ -190,25 +242,38 @@ func getTestFolderResourceConfig(parentFolderName, childFolderName string) strin
...
+        "cost-center" = "12345"
+	}
 }
 
 resource "cockroach_folder" "test_folder_child" {
     name = "%s"
 	parent_id = cockroach_folder.test_folder_parent.id
+	labels = {
+        managed_by = "engineer_a"
+        campaign = "98765"
```
fantapop, line 256: spacing on a some of these is off

internal/provider/cluster_resource_test.go
linhcrl: yes, there's a single error message which mentions the limit AND the valid format. It just happens to start with mentioning the limit

internal/provider/cluster_resource.go
linhcrl: updated

internal/provider/cluster_resource_test.go
linhcrl: fixed these and applied crlfmt to some of these files too

internal/provider/folder_resource_test.go
linhcrl: ah weird, IDE doesnt show any of this. Im guessing it's due to a mixed usage of spaces and tabs. I think it should all be update by now. Thanks again!

internal/provider/cluster_resource_test.go
fantapop: ah I see.  Ideally there would be a different error message because we don't really know from this test that it's not being triggered for the wrong reason.  If that's hard, we could just add a comment here saying that the error is for both too many and for invalid labels but we're just matching the start of it.

internal/provider/folder_resource.go
fantapop: That sounds good to me.  But we should vet it with @mikeCRL.

internal/provider/cluster_resource_test.go
linhcrl: Okay, I added a separate check for label limit and return a different error message if there are more labels than allowed.

Now we have separate errors for invalid labels and label limit

internal/provider/folder_resource.go
linhcrl: Ok cool! I updated the docs to reflect my proposal, but let me know if you'd like to modify @mikeCRL

mikeCRL (approved):

CHANGELOG.md
```diff
@@ -5,8 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.12.0] - 2025-04-30
```
linhcrl, line 8: I'll aim to release the changes on this day

fantapop: I merged down that PR with full_version.  You'll need to resolve the conflict in this file.

fantapop: oh you did it already .   :-)

fantapop (approved):
LGTM

PR #271, opened 2025-04-08
[FEB-93] add cluster attribute for full version string

fantapop:
In order to orchestrate version upgrades we allow specifying the major version of the cluster in the cluster resource.  Originally the cluster data source exposed the full version string but having this value be a different format from the resource (major vs full) was causing confusion so the data source was changed to match the resource.

At this point, the full version string was not available in any resource or data source so we add it back here as "full_version".

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)


linhcrl (commented):
LGTM!

linhcrl (approved):
Again, meant to approve

PR #275, opened 2025-04-24
add AWS external_id support for metric/log export

arjunmahishi:
This commit implements external_id support for both CloudWatch metric export
and log export configurations. External ID is used when assuming AWS IAM
roles and provides additional security for cross-account access.

Changes:

  * Upgrade cockroach-cloud-sdk-go to  v6.1.0
  * Add external_id field to CloudWatch metric export resource
  * Add aws_external_id field to log export resource
  * Update tests for both resources
  * Addresses an issue where fields incorrectly marked as computed led to
    miscalculation of diffs when fields were commented out or removed from the
    terraform config
  * Ensures backward compatibility for existing configurations

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] ~Acceptance test(s)~ Acc tests have been skipped for these resources. I've manually tested this extensively on a staging cluster. Will create a separate follow up ticket to enable the Acc tests if required. 
- [x] Example(s)

### DIFF
--- a/internal/provider/log_export_config_resource.go
+++ b/internal/provider/log_export_config_resource.go
@@ -60,7 +60,6 @@ var logExportAttributes = map[string]schema.Attribute{
 	"region": schema.StringAttribute{
 		Optional:            true,
 		MarkdownDescription: "Controls whether all logs are sent to a specific region in the customer sink.",
-		Computed:            true,
 	},
 	"groups": schema.ListNestedAttribute{
 		Optional: true,
@@ -109,6 +108,10 @@ var logExportAttributes = map[string]schema.Attribute{
 		Optional:    true,
 		ElementType: types.StringType,
 		Description: "Controls what CRDB channels do not get exported.",
+	},
+	"aws_external_id": schema.StringAttribute{
+		Optional:    true,
+		Description: "The external ID to use when assuming the AWS role.",
 	},
 }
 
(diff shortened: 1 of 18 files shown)


internal/provider/log_export_config_resource.go
```diff
@@ -311,18 +314,21 @@ func loadLogExportIntoTerraformState(
 		apiRedact = types.BoolValue(spec.GetRedact())
 	}
 
-	var apiRegion types.String
-	if spec.Region == nil {
-		apiRegion = types.StringNull()
-	} else {
-		apiRegion = types.StringValue(spec.GetRegion())
+	apiRegion := types.StringNull()
+	if spec.Region != nil && *spec.Region != "" {
```
arjunmahishi, line 318: We need to check for both nil and empty string. This mainly what caused the bug mentioned in the commit message

internal/provider/log_export_config_resource_test.go
```diff
@@ -300,6 +305,7 @@ resource "cockroach_log_export_config" "test" {
   type           = "AWS_CLOUDWATCH"
   redact         = false
   region 		 = "us-east-1"
```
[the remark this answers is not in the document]
arjunmahishi: done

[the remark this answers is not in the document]
arjunmahishi: I just formatted it with [hclfmt](https://github.com/fatih/hclfmt). Hence the spacing diff.

aa-joshi (approved):
Changes LGTM.

We should start running the integration tests for both log and metric export(not necessarily in this PR.)

- https://github.com/cockroachdb/terraform-provider-cockroach/blob/f5ba3854c613385d395ee9f5c4336b5e0e398f3f/internal/provider/metric_export_cloudwatch_config_resource_test.go#L37
- https://github.com/cockroachdb/terraform-provider-cockroach/blob/f5ba3854c613385d395ee9f5c4336b5e0e398f3f/internal/provider/log_export_config_resource_test.go#L37

internal/provider/log_export_config_resource.go
```diff
@@ -110,6 +109,10 @@ var logExportAttributes = map[string]schema.Attribute{
 		ElementType: types.StringType,
 		Description: "Controls what CRDB channels do not get exported.",
 	},
+	"aws_external_id": schema.StringAttribute{
```
Abhinav1299, line 112: In reference to the slack thread which discuss this issue, the default value of computed is then "false" I believe

arjunmahishi: Yes (follows default of `bool` in golang)

arjunmahishi:
TFTR!

PR #278, opened 2025-04-30
[CLOUDOPS-13734] handle external deletion of service account in user role grant resource

linhcrl:
Previously, the external deletion of a service account resource managed in Terraform caused errors on apply if there was a user role grant resource that referenced the user ID associated with the deleted service account. This PR fixes this issue by adding a RequireReplace plan modifier on the UserID attribute of the UserRoleGrant resource.

**Commit checklist**
- [x] Changelog
- [ ] Doc gen (`make generate`) - not relevant
- [x] Integration test(s)
- [x] Acceptance test(s)
- [ ] Example(s) - not relevant

### DIFF
--- a/internal/provider/user_role_grant_resource_test.go
+++ b/internal/provider/user_role_grant_resource_test.go
@@ -18,6 +18,7 @@ package provider
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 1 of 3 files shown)


internal/provider/user_role_grant_resource_test.go
```diff
@@ -498,3 +499,221 @@ resource "cockroach_user_role_grant" "a_duplicate_grant" {
...
+				RefreshState:       true,
+				ExpectNonEmptyPlan: true,
+			},
+			// This step reapplies the exact same configuration. Because drift was
+			// detected in the previous step, Terraform will create a new service
+			// account.
+			{
+				Config: getTestRoleGrantResourceWithNewUserConfig(),
+				Check:  resource.ComposeTestCheckFunc(checkOrgRole),
+			},
```
linhcrl, line 669: I tried to combine these two steps (add `PreConfig` and `ExpectNonEmptyPlan` to the third step and remove the second step) but I couldn't get it to work.

I did test this three-step version and without the bug fix, it gave the same error as the one our client saw, so I think it should be sufficient

internal/provider/user_role_grant_resource_test.go
```diff
@@ -498,3 +499,221 @@ resource "cockroach_user_role_grant" "a_duplicate_grant" {
...
+	if serviceAccountID == "" {
+		t.Fatalf("expected service account ID to be set")
+	}
+	p := testAccProvider.(*provider)
+	p.service = NewService(cl)
+	_, httpResp, err := p.service.DeleteServiceAccount(context.Background(), serviceAccountID)
+	if err != nil || httpResp.StatusCode != http.StatusOK {
+		t.Error(err != nil)
+		t.Error(httpOk.StatusCode)
+		t.Fatalf("failed to delete service account")
```
fantapop, line 717: It seems like these are a bit vague and different to how we commonly do error handling.  Can we use require here?  I guess similarly for the t.Fatalf above.
```suggestion
	require.NoError(t, err)
	require.Equal(t, httpResponse.StatusCode, http.StatusOK)
```

fantapop (commented):
Can you add the ticket to the commit header?  It would be good to have a place where the details of the bug are discussed.  I think the commit message is probably a good place for that. It's somewhat vague right now.

internal/provider/user_role_grant_resource_test.go
linhcrl: Updated

fantapop (approved):
LGTM

PR #285, opened 2025-06-12
Add UseStateForUnknown plan modifier to sql_dns and internal_dns

linhcrl:
The sql_dns and internal_dns fields are assigned at cluster creation and remain static afterward. However, the Terraform provider did not previously reflect this behavior, causing these fields to be marked as "known after apply" in every plan and apply.

This commit adds the UseStateForUnknown plan modifier to both fields so that after the initial apply (cluster creation), the provider will use the values stored in the state.

**Commit checklist**
- [ ] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/internal/provider/cluster_resource.go
+++ b/internal/provider/cluster_resource.go
@@ -77,6 +77,9 @@ var regionSchema = schema.NestedAttributeObject{
 		"sql_dns": schema.StringAttribute{
 			Computed:    true,
 			Description: "DNS name of the cluster's SQL interface. Used to connect to the cluster with IP allowlisting.",
+			PlanModifiers: []planmodifier.String{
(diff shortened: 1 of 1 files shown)


linhcrl:
Note: I tried adding some tests but couldn't come up with anything meaningful because to my knowledge it's hard to test the output of a tf plan. I did test locally by creating a cluster and making changes to it before running terraform plan - it did not show those two fields as "known after apply" after I added these changes

fantapop (approved):
This looks good to me.  It seems like you'll need to update one of the tests however.

linhcrl:
As discussed, I'm going to merge this and investigate the test failure separately. It's an unrelated failure that affects other PRs as well

PR #291, opened 2025-07-10
Upgrade Go and various dependencies

davidwding:
**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)


.github/workflows/ci.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.22
+          go-version: 1.23
```
fantapop, line 18: There are go versions hardcoded in each workflow file.  It probably makes sense to update them all in this PR at once.

davidwding:
> Based on the failures it looks like you might need to run a build. Try `make build`.
> 
> Also given that this one is doing a bunch of upgrades it's probably worth running a full test suite on it before merging.

Yeah, I had forgotten to include a file in the last commit, which is there now. I'm currently running the full test suite on my branch: https://github.com/cockroachdb/terraform-provider-cockroach/actions/runs/16203470625/job/45747978814

fantapop (approved):
LGTM

PR #292, opened 2025-07-10
Implement support for physical cluster replication streams

davidwding:
This implements support for replication stream creation and failover.
Deletion will come later.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)


internal/provider/physical_replication_stream.go
```diff
@@ -0,0 +1,384 @@
...
+	if apiObj.ActivatedAt != nil {
+		state.ActivatedAt = types.StringValue(apiObj.GetActivatedAt().String())
+	} else {
+		state.ActivatedAt = types.StringNull()
+	}
+	state.Status = types.StringValue(string(apiObj.GetStatus()))
+
+	if state.FailoverImmediately.ValueBool() {
+		state.FailoverImmediately = types.BoolValue(true)
+	}
```
fantapop, line 225: This seems a bit odd.  If `state.FailoverImmediately.ValueBool()` is true, then `state.FailoverImmediately` should already equal `types.BoolValue(true)` right?  I would think this is a no-op

internal/provider/physical_replication_stream.go
```diff
@@ -0,0 +1,384 @@
...
+
+	// Create update spec with FAILING_OVER status
+	failingOverStatus := client.REPLICATIONSTREAMSTATUSTYPE_FAILING_OVER
+	updateSpec := &client.UpdatePhysicalReplicationStreamSpec{
+		Status: &failingOverStatus,
+	}
+
+	// If failover_at is specified in the plan, set it in the update spec
+	// Only set failover_at if failover_immediately is not true
+	if !plan.FailoverAt.IsNull() && !plan.FailoverImmediately.ValueBool() {
```
fantapop, line 314: nit: it might help readability if you pulled these out into some local variables.  Each of them has been used multiple times now.  Could be something like:
```
isFailoverAt := !plan.FailoverAt.IsNull()
isFailoverImmediately := !plan.FailoverImmediately.IsNull()

if isFailoverAt && isFailoverImmediately  {
 // invalid configuration
}

if isFailoverImmediately && !plan.FailoverImmediately.ValueBool() {
   // .. invalid value for failover immediately
}

// now you know that isFailoverImmediately also indicates the value is valid
if isFailoverAt {
   // set the failoverAt value on the spec
}
```

internal/provider/physical_replication_stream_test.go
```diff
@@ -0,0 +1,229 @@
...
+		return false
+	}
+	return req.Name == m.ExpectedName
+}
+
+func (m clusterReqMatcher) String() string {
+	return fmt.Sprintf("clusterReqMatcher %s", m.ExpectedName)
+}
+
+func TestIntegrationPhysicalReplicationStreamResource(t *testing.T) {
```
fantapop, line 33: It's preferable to have both integration and acceptance tests for each feature.  Was your concern about the length or cost of spinning up 2 advanced clusters?

internal/provider/physical_replication_stream.go
```diff
@@ -0,0 +1,384 @@
...
+	ID                    types.String `tfsdk:"id"`
+	CreatedAt             types.String `tfsdk:"created_at"`
+	ReplicatedTime        types.String `tfsdk:"replicated_time"`
+	ReplicationLagSeconds types.Int64  `tfsdk:"replication_lag_seconds"`
+	RetainedTime          types.String `tfsdk:"retained_time"`
+	Status                types.String `tfsdk:"status"`
+	FailoverAt            types.String `tfsdk:"failover_at"`
+	FailoverImmediately   types.Bool   `tfsdk:"failover_immediately"`
+	ActivatedAt           types.String `tfsdk:"activated_at"`
+}
```
fantapop, line 31: Do you mind moving this to the internal/provider/models.go file?   

I wouldn't actually mind having them in their specific resource files but for now, the convention is to have them there.

fantapop (commented):
Added some feedback for you.  One other question I have is whether we should have a data source for these.

internal/provider/physical_replication_stream_test.go
davidwding: I've added acceptance tests.

davidwding (commented):
PTAL!

> One other question I have is whether we should have a data source for these.

I can add a data source. Mind if I do it in a separate PR to keep this one smaller?

internal/provider/physical_replication_stream.go
```diff
@@ -0,0 +1,380 @@
...
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"PCR update failed",
+			fmt.Sprintf("PCR stream failed to update: %s", formatAPIErrorMessage(err)),
+		)
+		return
+	}
+
+	state.FailoverAt = plan.FailoverAt
+	state.FailoverImmediately = plan.FailoverImmediately
```
fantapop, line 349: Is there any way that the FailoverAt value coming back from the update response could be different from what is set in the plan?  

For example, what if FailoverAt was set manually via the api to a value before or after the plan value.  etc...

internal/provider/physical_replication_stream.go
```diff
@@ -0,0 +1,380 @@
...
+	ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse,
+) {
+	var state PhysicalReplicationStream
+	diags := req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if state.Status != types.StringValue(string(client.REPLICATIONSTREAMSTATUSTYPE_COMPLETED)) {
```
fantapop, line 367: It looks like Canceled is in the api.proto also.  Is there not a reason to check for that here also?

fantapop (commented):
I'm still working back through this but have another meeting.  I'll submit a couple questions I came up with.

internal/provider/physical_replication_stream.go
davidwding: Good question. I just checked and yes, the value could actually be updated manually via a separate API call. Should we also handle this with a separate field? Something like `failover_requested_at`?

fantapop (commented):
I just noticed we're missing the example file for this resource. Also left a handful of nits.

There should be:

examples/phyical_replication_stream/resource.tf
examples/phyical_replication_stream/import.sh

You can look at the other files for how they should look.

internal/provider/physical_replication_stream_test.go
```diff
@@ -0,0 +1,267 @@
...
+    name           = "%s"
+    cloud_provider = "GCP"
+    dedicated = {
+        storage_gib = 15
+        num_virtual_cpus = 4
+        cidr_range = "172.28.0.0/19"
+    }
+    regions = [{
+        name = "us-central1"
+				node_count = 3
```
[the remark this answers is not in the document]
davidwding: Ugh, tabs vs. spaces. Fixed

internal/provider/physical_replication_stream.go
davidwding: As discussed separately, this is fine as is.

davidwding (commented):
> I just noticed we're missing the example file for this resource. Also left a handful of nits.
> 
> There should be:
> 
> examples/phyical_replication_stream/resource.tf examples/phyical_replication_stream/import.sh
> 
> You can look at the other files for how they should look.

Done

davidwding:
I've verified that the acceptance tests still pass after refactoring the tests as you suggested.

fantapop (approved):
cool.  This all looks good to me now. Thank you!

PR #293, opened 2025-07-16
Add support_physical_cluster_replication flag to Cluster

davidwding:
**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)

### DIFF
--- a/internal/provider/cluster_resource.go
+++ b/internal/provider/cluster_resource.go
@@ -268,6 +268,16 @@ func (r *clusterResource) Schema(
 							stringplanmodifier.UseStateForUnknown(),
 						},
 					},
+					"support_physical_cluster_replication": schema.BoolAttribute{
+						Optional: true,
+						// This is false since our SDK does not currently provide
+						// any indicator of whether a cluster supports PCR or not.
+						Computed:            false,
+						MarkdownDescription: "This field specifies whether a cluster should be started using an architecture that supports physical cluster replication. This field is restricted to Limited Access usage; see our documentation for details: https://www.cockroachlabs.com/docs/cockroachcloud/physical-cluster-replication.html",
+						PlanModifiers: []planmodifier.Bool{
+							boolplanmodifier.UseStateForUnknown(),
+						},
+					},
 				},
 			},
 			"regions": schema.ListNestedAttribute{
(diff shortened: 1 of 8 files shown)


internal/provider/cluster_resource.go
```diff
@@ -1363,6 +1377,11 @@ func loadClusterToTerraformState(
 			PrivateNetworkVisibility: types.BoolValue(clusterObj.GetNetworkVisibility() == client.NETWORKVISIBILITYTYPE_PRIVATE),
 			CidrRange:                types.StringValue(clusterObj.CidrRange),
 		}
+		if plan != nil && plan.DedicatedConfig != nil {
+			// We want to make sure to set this bool to false if it was
+			// omitted entirely.
+			state.DedicatedConfig.SupportPhysicalClusterReplication = types.BoolValue(plan.DedicatedConfig.SupportPhysicalClusterReplication.ValueBool())
+		}
```
fantapop, line 1384: This part is somewhat confusing.  I wonder if we could make it a bit better.  I think partially it's confusing because defaulting to false is mentioned in the comment and then the value from the plan is chosen. Presumably the value could be true?   

The other thing that I think is confusing is about when the plan would be nil.  Inspecting the callers reveals that on a Read the state is passed in as the plan. I think that is the only case where it might be nil? But still that seems a bit strange.   If you've figured this out during development it would be good to leave a more detailed comment about this situation.  

I'm also wondering if we can incorporate it into the setting of `state.DedicatedConfig` above instead of after the fact.

internal/provider/cluster_resource.go
```diff
@@ -1363,6 +1379,16 @@ func loadClusterToTerraformState(
...
 		}
+		// As noted in the comment of this function, plan may be nil if we
+		// are reading a datasource or importing a resource. In these cases,
+		// since there is currently no indication from the SDK whether a
+		// cluster supports PCR or not, we leave the field blank.
+		if plan != nil && plan.DedicatedConfig != nil {
+			// Otherwise, if we have explicitly specified this flag in the
+			// plan, then we add it to the state. If it was null, we keep it
+			// as null.
+			state.DedicatedConfig.SupportPhysicalClusterReplication = plan.DedicatedConfig.SupportPhysicalClusterReplication
```
fantapop, line 1390: I'm wondering if it's fine that we don't return the value back from the cluster.  It means for instance that if someone were to import a cluster with this property set on it, it wouldn't be verifiable.  Do you know what happens in this case?

fantapop (approved):
LGTM

PR #295, opened 2025-08-06
[CNSL-1392] add cockroach_restore resource

linhcrl:
**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)


internal/provider/restore_resource.go
```diff
@@ -0,0 +1,353 @@
...
+			fmt.Sprintf("Could not create restore: %s", formatAPIErrorMessage(err)),
+		)
+		return
+	}
+
+	// Poll until restore operation completes.
+	err = retry.RetryContext(ctx, createRestoreEndpointTimeout,
+		waitForRestoreReadyFunc(ctx, plan.DestinationClusterID.ValueString(), restoreObj.Id.ValueString(), r.provider.service, restoreObj))
+	if err != nil {
+		resp.Diagnostics.AddError("restore operation timed out and may have failed", formatAPIErrorMessage(err))
```
fantapop, line 201: nit: do we know this was a timeout?  I assume if the endpoint request returns a permanent error right away it wouldn't be a timeout.

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,353 @@
...
+						},
+						"schema": schema.StringAttribute{
+							MarkdownDescription: "The schema portion of the name of the object to be restored. In the case of `tpcc.public.warehouse`, this would be `public`.",
+							Optional:            true,
+						},
+						"table": schema.StringAttribute{
+							MarkdownDescription: "The table portion of the name of the object to be restored. In the case of `tpcc.public.warehouse`, this would be `warehouse`.",
+							Optional:            true,
+						},
+					},
```
fantapop, line 74: This method of inputting data seem pretty painful.  I was hoping we could figure out a way to pull data directly from a backup object to populate this.   I don't know exactly what that would look like.  Do you have any ideas?

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,353 @@
...
+	restoreID := state.ID.ValueString()
+
+	// In case this was an import, validate the ID format.
+	if !uuidRegex.MatchString(restoreID) {
+		resp.Diagnostics.AddError(
+			"Unexpected restore ID format",
+			fmt.Sprintf("'%s' is not a valid restore ID format. Expected UUID.", restoreID),
+		)
+		return
+	}
```
fantapop, line 243: It seems like this could be done with a validator.  I think there is already a UUID validator.  Can you experiment with this?

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,353 @@
...
+
+	traceAPICall("GetRestore")
+	restoreObj, httpResp, err := r.provider.service.GetFolder(ctx, restoreID)
```
fantapop, line 246: I assume this is  a copy paste issue.
```suggestion
	traceAPICall("GetRestore")
	restoreObj, httpResp, err := r.provider.service.GetRestore(ctx, restoreID)
```

internal/provider/restore_resource.go
linhcrl: You are right, it's not obvious whether it's a timeout. Since this is a unique case of an operation just taking too long for terraform to keep polling, I'm not sure what the appropriate messaging would be here.

"Operation failed or timed out." ?

internal/provider/restore_resource.go
linhcrl, line 1: In many cases, I needed to check whether a field was actually set before assigning a value to it. This ensures that the translation between the client object and the original input (and vice versa) remains as accurate as possible.

That’s why you’ll see a lot of `IsUnknown()` and `if variable != nil {}` checks throughout the code.

internal/provider/restore_resource.go
linhcrl: I left a separate comment explaining my solution

internal/provider/restore_resource_test.go
```diff
@@ -0,0 +1,571 @@
...
+	%s
+	%s
+}
+`, restoreType, constructRestoreItems(objects), constructRestoreOpts(opts))
+	}
+	return resources
+}
+
+func constructRestoreItems(objects []RestoreItem) string {
+	var parts []string
```
fantapop, line 436: parts is a bit generic of a name.  How about something like `objectBlocks`.  Or another fine option would be to use a name like config and just append the the string to itself.

```
config := ""
for _, o := range objects {
     ....
     config += fmt.Sprintf(`objects {
     ... 
```

internal/provider/restore_resource_test.go
```diff
@@ -0,0 +1,571 @@
...
+				tables = append(tables, table.String())
+			}
+
+			tablesStr := fmt.Sprintf(`[%s]`, strings.Join(tables, `, `))
+			fields = append(fields, fmt.Sprintf(`tables = %s`, tablesStr))
+		}
+
+		parts = append(parts, fmt.Sprintf(`objects {
+		%s
+	}`, strings.Join(fields, "\n\t\t")))
```
fantapop, line 460: This function is a not as clear as some of the other functions for generating hcl. It's obvsiouly more complicated since it's dynamic but I was trying to figure how to make it more clear.

I think using the client types as objects will help (as mentioned above) But I also think breaking out `fields` into multiple variables would also help.  I think the shape the config will be more clear and you won't need the \n\t\t anymore.  Something like:

```
config =+ fmt.Sprintf(`
objects {
  %s
  %s
  %s
}
`, database, schema, tables)
```

Or maybe breaking it into 2 separate blocks depending on whether it was database or tables like:

```
if tables != nil {
config =+ fmt.Sprintf(`
objects {
  database = "%s"
  schema = "%s"
  tables = ["%s"]
}
`, database, schema, strings.Join(tables, `", "`))
} else {

config =+ fmt.Sprintf(`
objects {
  database = "%s"
}
`, database)
}
```

internal/provider/restore_resource_test.go
```diff
@@ -0,0 +1,571 @@
...
+		restoreResourceName = "cockroach_restore.test_restore"
+		clusterResourceName = "cockroach_cluster.test_cluster"
+	)
+	resource.Test(t, resource.TestCase{
+		IsUnitTest:               useMock,
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: getTestRestoreResourceConfig(clusterName, false /*includeRestore*/, restoreType, objects, opts, isAdvancedCluster),
```
fantapop, line 340: It would be nice to comment what each of these steps is doing.   It looks like the first step is creating a cluster and enabling backups and then the second step is actually creating the restore?  You could just use a comment.  Or optionally, you could use the step printing mechanism I set up.  Here Is an example: https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/internal/provider/cluster_resource_test.go#L334-L336

When you run the tests with the TRACE_API_CALLS env set, it will print out the steps along with all the api calls so you can easily copy them for the integration tests.  Discussed brief in the testing section of the [DEVELOPMENT.md](https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/DEVELOPMENT.md#testing)

internal/provider/restore_resource_test.go
```diff
@@ -0,0 +1,571 @@
...
+	}
+
+	createRestoreReq := &client.CockroachCloudCreateRestoreRequest{
+		Type:            client.RESTORETYPETYPE_TABLE,
+		SourceClusterId: &clusterID,
+		Objects: &[]client.RestoreItem{
+			{Database: ptr("test_db"), Schema: ptr("public"), Table: ptr("table1")},
+			{Database: ptr("test_db"), Schema: ptr("public"), Table: ptr("table2")},
+			{Database: ptr("test_db2"), Schema: ptr("public"), Table: ptr("table3")},
+		},
```
fantapop, line 284: nit: creating these is a bit unwieldy.  You should consider making some helpers.  I was thinking something like:
```
Objects: &[]client.RestoreItem{
    dbRestoreItem("test_db"),
    tablesRestoreItem("test_db", "public", []string{"table1", "table2"})
}
```

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,400 @@
...
+		return
+	}
+
+	// Poll until restore operation completes.
+	err = retry.RetryContext(ctx, createRestoreEndpointTimeout,
+		waitForRestoreReadyFunc(ctx, plan.DestinationClusterID.ValueString(), restoreObj.Id, r.provider.service, restoreObj))
+	if err != nil {
+		if err.Error() == createRestorePendingErr {
+			resp.Diagnostics.AddError("Restore operation timed out", formatAPIErrorMessage(err))
+		} else {
```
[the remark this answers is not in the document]
linhcrl: makes sense I'll update it accordingly. `errors.Is(err, retry.TimoutError)` doesn't work since some methods arent implemented for `retry.TimeoutError` that are required to be acceptable for `errors.Is`

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,400 @@
...
+							Required:            true,
+						},
+						"schema": schema.StringAttribute{
+							MarkdownDescription: "The schema portion of the name of the objects to be restored. In the case of `tpcc.public.warehouse`, this would be `public`.",
+							Optional:            true,
+						},
+						"tables": schema.ListAttribute{
+							MarkdownDescription: "The table portion of the name of the objects to be restored. In the case of `tpcc.public.warehouse`, this would be `warehouse`.",
+							Optional:            true,
+							ElementType:         types.StringType,
```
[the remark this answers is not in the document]
linhcrl: this field is a list so we need to specify what kind of elements are in the list

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,400 @@
...
+		Type: client.RestoreTypeType(plan.Type.ValueString()),
+	}
+
+	if !plan.BackupID.IsUnknown() {
+		createRestoreReq.BackupId = plan.BackupID.ValueStringPointer()
+	}
+
+	if !plan.SourceClusterID.IsUnknown() {
+		createRestoreReq.SourceClusterId = plan.SourceClusterID.ValueStringPointer()
+	}
```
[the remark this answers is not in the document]
linhcrl: I think think I would prefer to rely on the API handling all the validation since all the other nuanced checks already live there. Since we don't make any db calls before the validation, I'm not too concerned about it

internal/provider/restore_resource_test.go
```diff
@@ -0,0 +1,571 @@
...
+		Regions: []client.Region{
+			{
+				Name:      "us-central1",
+				NodeCount: 1,
+			},
+		},
+		CidrRange:        "172.28.0.0/16",
+		CloudProvider:    client.CLOUDPROVIDERTYPE_GCP,
+		CockroachVersion: "v25.2",
+		Plan:             client.PLANTYPE_ADVANCED,
```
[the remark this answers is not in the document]
linhcrl: yeah but since this is integration test, I figured it didn't matter. For acceptance test I am using standard. Let me know what you think

internal/provider/restore_resource_test.go
```diff
@@ -0,0 +1,571 @@
...
+	}]
+	backup_config = {
+		enabled           = true
+		frequency_minutes = 5
+		retention_days    = 30
+	}
+}
+`, clusterName)
+
+	if isAdvancedCluster {
```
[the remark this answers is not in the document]
linhcrl: for the purpose of the integration test it doesn't matter as much because we can return anything in the mocked response. But technically, we cannot have table and database restores on standard clusters so in order to represent actual scenarios, I'm using advanced clusters for integration tests

internal/provider/restore_resource_test.go
```diff
@@ -0,0 +1,571 @@
...
+	}
+	return resources
+}
+
+func constructRestoreItems(objects []RestoreItem) string {
+	var parts []string
+	for _, o := range objects {
+		var fields []string
+
+		if !o.Database.IsNull() && !o.Database.IsUnknown() {
```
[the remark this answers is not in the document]
linhcrl: fair point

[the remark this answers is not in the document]
linhcrl: Actually on a second thought I wolud rather keep the existing types. The reason is that the terraform types follow the same structure that is needed for the HCL config.

`client.RestoreItem` is a flattened out structure where each object is one database, one schema, and one table. To convert that to the structure the config accepts, we would need to keep track of which database and schema combination has been seen already and collect associated tables, which is a bit of an overkill for tests

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,400 @@
...
+	// responsible for removing the resource state.
+	resp.State.RemoveResource(ctx)
+}
+
+func (r *restoreResource) ImportState(
+	ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse,
+) {
+	resp.Diagnostics.AddError(
+		"Import not supported",
+		"Restore resources cannot be imported. Restore operations are one-time operations that can be created through Terraform.",
```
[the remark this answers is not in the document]
linhcrl: I removed the second half to keep it simple and easy to understand

internal/provider/restore_resource_test.go
linhcrl: I went with the first option, I find that easier to follow/simple

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,400 @@
...
+					"skip_localities_check": schema.BoolAttribute{
+						Description: "Allows the restore operation to continue in the event that there are mismatched localities between the backup and target cluster. Useful when restoring multi-region tables to a cluster missing some localities.",
+						Optional:    true,
+					},
+					"skip_missing_foreign_keys": schema.BoolAttribute{
+						Description: "Allows a table to be restored even if it has foreign key constraints referencing rows that no longer exist in the target cluster.",
+						Optional:    true,
+					},
+					"skip_missing_sequences": schema.BoolAttribute{
+						Description: "Allows a table to be restored even if it contains a column whose `DEFAULT` value depends on a sequence. (See https://www.cockroachlabs.com/docs/stable/show-sequences)",
```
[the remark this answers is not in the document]
linhcrl: done

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,400 @@
...
+	if !plan.BackupID.IsUnknown() {
+		createRestoreReq.BackupId = plan.BackupID.ValueStringPointer()
+	}
+
+	if !plan.SourceClusterID.IsUnknown() {
+		createRestoreReq.SourceClusterId = plan.SourceClusterID.ValueStringPointer()
+	}
+
+	var objects []client.RestoreItem
+	if len(plan.Objects) > 0 {
```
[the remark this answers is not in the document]
linhcrl: done

internal/provider/restore_resource_test.go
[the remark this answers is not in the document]
linhcrl: left a comment

internal/provider/restore_resource_test.go
linhcrl: updated to use step printing mechanism

internal/provider/restore_resource_test.go
linhcrl: done

internal/provider/restore_resource_test.go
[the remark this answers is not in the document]
fantapop: If reusing `client.RestoreItem` you could take advantage of your new helpers for creating the items and you could drop all the `isKnown` checks and `.ValueString()` calls.  As you mentioned, you'd probably want a method to group the api format by table.

If there were more of these data structures being created, I'd feel stronger about either reusing the client types or creating some intermediate testing types but since there are only 2 it seems fine to leave it how you have.  I'll resolve this one.

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,399 @@
...
+	state.Status = types.StringValue(string(restoreObj.Status))
+	state.CreatedAt = types.StringValue(restoreObj.CreatedAt.String())
+	state.CompletionPercent = types.Float32Value(restoreObj.CompletionPercent)
+
+	if sourceClusterId != nil {
+		state.SourceClusterID = *sourceClusterId
+	}
+	if destinationClusterId != nil {
+		state.DestinationClusterID = *destinationClusterId
+	}
```
fantapop, line 366: It strikes me as surprising these these are pointers.  Can you using something like 
```
IsKnown(plan.SourceClusterID)
```
 instead of passing it as a pointer?

fantapop (approved):
I have a couple of remaining nits but we should wait for @jhlodin to sign off on the docs strings.  Also,  I forget, will there be a separate PR for the example?

internal/provider/restore_resource.go
linhcrl: This method is used both in the Create and Read functions. These resource fields get set in Create function and then remain unchanged for the rest of the lifetime of the resource. I made the variables pointers so that in the Read function, we don't try to reset these fields. The Read method does not know the Source/Destination cluster since GetRestore doesn't contain this info

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,399 @@
...
+		)
+		return
+	}
+
+	// Poll until restore operation completes.
+	err = retry.RetryContext(ctx, createRestoreEndpointTimeout,
+		waitForRestoreReadyFunc(ctx, plan.DestinationClusterID.ValueString(), restoreObj.Id, r.provider.service, restoreObj))
+	if err != nil {
+		if errors.Is(err, createRestorePendingErr) {
+			resp.Diagnostics.AddError("Restore operation timed out", "The restore operation has not been marked as failed and will continue to progress in the background.")
```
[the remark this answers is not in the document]
linhcrl: done

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,403 @@
...
+
+func NewRestoreResource() resource.Resource {
+	return &restoreResource{}
+}
+
+func (r *restoreResource) Schema(
+	_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse,
+) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: "CockroachDB Cloud restore. Can be used to kick off a restore operation.",
```
jhlodin, line 35:
```suggestion
		MarkdownDescription: "CockroachDB Cloud restore. Used to start a restore operation from a managed backup.",
```

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,403 @@
...
+				Required:    true,
+				Validators:  uuidValidator,
+			},
+			"type": schema.StringAttribute{
+				MarkdownDescription: "The type of restore to be performed. Allowed values are: `CLUSTER` or `DATABASE` or `TABLE`.",
+				Required:            true,
+				Validators:          []validator.String{stringvalidator.OneOf("CLUSTER", "DATABASE", "TABLE")},
+			},
+			"backup_id": schema.StringAttribute{
+				Description: "The ID of the backup from which data will be restored. If backup_id is not set, the restore will use the latest available backup from source_cluster_id.",
```
jhlodin, line 51:
```suggestion
				Description: "ID of the managed backup from which data will be restored. If this value is not set, the restore operation uses the most recent available backup on the source cluster.",
```

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,403 @@
...
+				Optional:    true,
+				Computed:    true,
+				Validators:  uuidValidator,
+			},
+			"source_cluster_id": schema.StringAttribute{
+				Description: "The ID of the cluster containing the backup to be restored.",
```
jhlodin, line 57:
```suggestion
				Description: "ID of the source cluster containing the managed backup to be restored.",
```

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,403 @@
...
+				Optional:    true,
+				Validators:  uuidValidator,
+			},
+			"restore_opts": schema.SingleNestedAttribute{
+				Description: "Additional options controlling the behavior of the restore operation.",
+				Optional:    true,
+				Attributes: map[string]schema.Attribute{
+					"new_db_name": schema.StringAttribute{
+						Description: "Specifies the name of the database to create during a database restore operation. If not set, the name defaults to the original database name from the source cluster.",
```
jhlodin, line 66:
```suggestion
						Description: "Optionally specifies the name of a new database to create as the target of a database restore operation. If not set, the name defaults to the original database name from the source cluster.",
```

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,403 @@
...
+						Optional:    true,
+					},
+					"into_db": schema.StringAttribute{
+						Description: "Specifies the target database to restore the table into during a table restore operation. If not set, the table will be restored into the database it belonged to in the source backup.",
```
jhlodin, line 70:
```suggestion
						Description: "Optionally specifies a target database to restore the table into during a table restore operation. If not set, the table is restored into the database it belonged to in the source backup.",
```

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,403 @@
...
+					"skip_missing_foreign_keys": schema.BoolAttribute{
+						Description: "Allows a table to be restored even if it has foreign key constraints referencing rows that no longer exist in the target cluster.",
+						Optional:    true,
+					},
+					"skip_missing_sequences": schema.BoolAttribute{
+						MarkdownDescription: "Allows a table to be restored even if it contains a column whose `DEFAULT` value depends on a sequence. More information can be found [here](https://www.cockroachlabs.com/docs/stable/show-sequences).",
+						Optional:            true,
+					},
+					"schema_only": schema.BoolAttribute{
+						Description: "If set, only the schema will be restored and no user data will be included.",
```
jhlodin, line 86:
```suggestion
						Description: "If set, only the schema without any user data is restored.",
```

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,403 @@
...
+			"status": schema.StringAttribute{
+				Description: "The current status of the restore operation.",
+				Computed:    true,
+			},
+			"created_at": schema.StringAttribute{
+				Description: "The time at which the restore operation was initiated.",
+				Computed:    true,
+			},
+			"completion_percent": schema.Float32Attribute{
+				Description: "The percentage of the restore operation that has been completed. The value ranges from 0 to 1.",
```
jhlodin, line 100:
```suggestion
				Description: "Decimal value showing the percentage of the restore operation that has been completed. Value ranges from 0 to 1.",
```

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,403 @@
...
+				Computed:    true,
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"objects": schema.ListNestedBlock{
+				Description: "The list of objects to restore. Required when type is DATABASE or TABLE.",
```
jhlodin, line 106:
```suggestion
				Description: "The list of objects to restore. Required when the restore operation type is DATABASE or TABLE.",
```

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,403 @@
...
+				NestedObject: schema.NestedBlockObject{
+					Attributes: map[string]schema.Attribute{
+						"database": schema.StringAttribute{
+							MarkdownDescription: "The database portion of the name of the objects to be restored. In the case of `tpcc.public.warehouse`, this would be `tpcc`.",
```
jhlodin, line 110:
```suggestion
							MarkdownDescription: "The database name in the fully qualified name of the objects to be restored. In the case of an operation that restores `tpcc.public.warehouse`, this value would be `tpcc`.",
```

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,403 @@
...
+							Required:            true,
+						},
+						"schema": schema.StringAttribute{
+							MarkdownDescription: "The schema portion of the name of the objects to be restored. In the case of `tpcc.public.warehouse`, this would be `public`.",
```
jhlodin, line 114:
```suggestion
							MarkdownDescription: "The schema name in the fully qualified name of the objects to be restored. In the case of an operation that restores `tpcc.public.warehouse`, this value would be `public`.",
```

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,403 @@
...
+							Optional:            true,
+						},
+						"tables": schema.ListAttribute{
+							MarkdownDescription: "The table portion of the name of the objects to be restored. In the case of `tpcc.public.warehouse`, this would be `warehouse`.",
```
jhlodin, line 118:
```suggestion
							MarkdownDescription: "The table name in the fully qualified name of the objects to be restored. In the case of an operation that restores `tpcc.public.warehouse`, this would value be `warehouse`.",
```

jhlodin:
Please tag me for review again when the markdown file is generated. Thanks!

internal/provider/restore_resource.go
```diff
@@ -0,0 +1,403 @@
...
+			"id": schema.StringAttribute{
+				Computed: true,
+			},
+			"destination_cluster_id": schema.StringAttribute{
+				Description: "The ID of the cluster where the backup will be restored.",
+				Required:    true,
+				Validators:  uuidValidator,
+			},
+			"type": schema.StringAttribute{
+				MarkdownDescription: "The type of restore to be performed. Allowed values are: `CLUSTER` or `DATABASE` or `TABLE`.",
```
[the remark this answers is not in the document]
linhcrl: I’m a bit concerned about using the word _resource_ in this field description, since in Terraform that term has a very specific meaning (it refers to the managed items like cockroach_restore). I worry this could cause confusion for users reading the docs. Do you have another suggestion perhaps?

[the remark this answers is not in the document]
jhlodin: Object?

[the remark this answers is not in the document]
linhcrl: we use objects to refer to the actual items to restore (e.g. database name or  table name)

How about "The _type of objects_ to be restored"

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.0]
+
+### Added
+
+- Added cockroach_restore resource for starting restore operations from a managed backup.
```
jhlodin, line 12:
```suggestion
- Added cockroach_restore resource for starting restore jobs from a managed backup.
```

docs/resources/restore.md
```diff
@@ -0,0 +1,102 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "cockroach_restore Resource - terraform-provider-cockroach"
+subcategory: ""
+description: |-
+  CockroachDB Cloud restore. Used to start a restore operation from a managed backup.
```
jhlodin, line 6:
```suggestion
  CockroachDB Cloud restore. Used to start a restore job from a managed backup.
```

docs/resources/restore.md
```diff
@@ -0,0 +1,102 @@
...
+---
+
+# cockroach_restore (Resource)
+
+CockroachDB Cloud restore. Used to start a restore operation from a managed backup.
```
jhlodin, line 11:
```suggestion
CockroachDB Cloud restore. Used to start a restore job from a managed backup.
```

docs/resources/restore.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Example Usage
+
+```terraform
+resource "cockroach_restore" "database_restore" {
+  destination_cluster_id = "12345678-1234-4123-8123-123456789abc"
+  type                   = "DATABASE"
+  source_cluster_id      = "deadbeef-dead-4bee-8def-deadbeefcafe"
```
jhlodin, line 19: Dead beef cafe? No change needed but curious about the name

docs/resources/restore.md
```diff
@@ -0,0 +1,102 @@
...
+
+- `destination_cluster_id` (String) ID of the cluster where the backup will be restored.
+- `type` (String) The type of restore to be performed. Allowed values are: `CLUSTER` or `DATABASE` or `TABLE`.
+
+### Optional
+
+- `backup_id` (String) ID of the managed backup from which data will be restored. If this value is not set, the restore operation uses the most recent available backup on the source cluster.
+- `objects` (Block List) The list of objects to restore. Required when the restore operation type is DATABASE or TABLE. (see [below for nested schema](#nestedblock--objects))
+- `restore_opts` (Attributes) Additional options controlling the behavior of the restore operation. (see [below for nested schema](#nestedatt--restore_opts))
+- `source_cluster_id` (String) ID of the source cluster containing the managed backup to be restored.
```
jhlodin, line 70:
```suggestion
- `backup_id` (String) ID of the managed backup from which data will be restored. If this value is not set, the restore job uses the most recent available backup on the source cluster.
- `objects` (Block List) The list of objects to restore. Required when the restore job type is DATABASE or TABLE. (see [below for nested schema](#nestedblock--objects))
- `restore_opts` (Attributes) Additional options controlling the behavior of the restore job. (see [below for nested schema](#nestedatt--restore_opts))
- `source_cluster_id` (String) ID of the source cluster containing the managed backup to be restored.
```

docs/resources/restore.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Read-Only
+
+- `completion_percent` (Number) Decimal value showing the percentage of the restore operation that has been completed. Value ranges from 0 to 1.
+- `created_at` (String) The time at which the restore operation was initiated.
+- `id` (String) The ID of this resource.
+- `status` (String) The current status of the restore operation.
```
jhlodin, line 77:
```suggestion
- `completion_percent` (Number) Decimal value showing the percentage of the restore job that has been completed. Value ranges from 0 to 1.
- `created_at` (String) The time at which the restore job was initiated.
- `id` (String) The ID of this resource.
- `status` (String) The current status of the restore job.
```

docs/resources/restore.md
```diff
@@ -0,0 +1,102 @@
...
+
+<a id="nestedblock--objects"></a>
+### Nested Schema for `objects`
+
+Required:
+
+- `database` (String) The database name in the fully qualified name of the objects to be restored. In the case of an operation that restores `tpcc.public.warehouse`, this value would be `tpcc`.
```
jhlodin, line 84:
```suggestion
- `database` (String) The database name in the fully qualified name of the objects to be restored. In the case of an job that restores `tpcc.public.warehouse`, this value would be `tpcc`.
```

docs/resources/restore.md
```diff
@@ -0,0 +1,102 @@
...
+
+Optional:
+
+- `schema` (String) The schema name in the fully qualified name of the objects to be restored. In the case of an operation that restores `tpcc.public.warehouse`, this value would be `public`.
+- `tables` (List of String) The table name in the fully qualified name of the objects to be restored. In the case of an operation that restores `tpcc.public.warehouse`, this would value be `warehouse`.
```
jhlodin, line 89:
```suggestion
- `schema` (String) The schema name in the fully qualified name of the objects to be restored. In the case of a job that restores `tpcc.public.warehouse`, this value would be `public`.
- `tables` (List of String) The table name in the fully qualified name of the objects to be restored. In the case of a job that restores `tpcc.public.warehouse`, this would value be `warehouse`.
```

jhlodin (approved):
Need to use the term "job" instead of "operation", see slack thread for context - https://cockroachlabs.slack.com/archives/C08QU70N8AX/p1755878413494069

Otherwise LGTM

linhcrl:
Thanks @jhlodin updated all the references to operation

internal/provider/restore_resource.go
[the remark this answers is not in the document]
fantapop: How about something like "Type describes at what level the restore will take place.  When using a DATABASE or TABLE level restore additional detail passed in the `objects` attribute is required.  Allowed values are: ..."

examples/resources/cockroach_restore/resource.tf
```diff
@@ -0,0 +1,39 @@
+resource "cockroach_restore" "database_restore" {
```
fantapop, line 1: it would be nice to have a cluster level restore in the example as well

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.0]
```
fantapop, line 8: can you remove the tag?

linhcrl: done

internal/provider/restore_resource.go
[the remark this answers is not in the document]
jhlodin: Discussed offline with Linh, building on your suggestion - 

"Type describes the scope of the restore job. When using a DATABASE or TABLE restore, additional details must be passed in the `objects` attribute. Allowed values are: ..."

PR #300, opened 2025-08-22
Add egress_private_endpoint resource

mwstobo:
This commit implements an egress private endpoint as a Terraform resource. See below for official Cockroach Cloud documentation for the feature. This feature is currently in limited access and is not yet available to all customers.

Note that support for adding custom domain names to an egress private endpoint will come in a later commit.

https://www.cockroachlabs.com/docs/cockroachcloud/egress-private-endpoints.html

**Commit checklist**
- [ ] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)


internal/provider/egress_private_endpoint_resource_test.go
```diff
@@ -0,0 +1,271 @@
...
+			},
+		}
+	}
+
+	resource.Test(t, resource.TestCase{
+		IsUnitTest:               useMock,
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+		Steps:                    steps,
+	})
```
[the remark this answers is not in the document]
mwstobo: Good call. I initially had this as part of the integration tests, but it got complicated when I got to the acceptance tests. I'll add it back for integration tests only.

docs/resources/egress_private_endpoint.md
```diff
@@ -0,0 +1,178 @@
...
+description: |-
+  Egress private endpoint for a CockroachDB cluster. Egress private endpoints allow CockroachDB clusters to connect to external services through private network connections without traversing the public internet.
+  Warning: Use of egress private endpoints requires >=v1.14.0
+  If you intend to use this provider to provision egress private endpoints you must install/upgrade to version 1.14.0 https://github.com/cockroachdb/terraform-provider-cockroach/releases/tag/v1.14.0 or later https://registry.terraform.io/providers/cockroachdb/cockroach/latest
+  Terraform resource docs https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/docs/resources/egress_private_endpoint.mdCockroach Cloud docs https://www.cockroachlabs.com/docs/cockroachcloud/egress-private-endpoints.html
+---
+
+# cockroach_egress_private_endpoint (Resource)
+
+Egress private endpoint for a CockroachDB cluster. Egress private endpoints allow CockroachDB clusters to connect to external services through private network connections without traversing the public internet.
```
angeladietz, line 14: nit: CockroachDB Cloud cluster. Same thing everywhere else in the pr

docs/resources/egress_private_endpoint.md
```diff
@@ -0,0 +1,178 @@
...
+
+# Example: GCP endpoint for an application service
+resource "cockroach_egress_private_endpoint" "gcp_app_service" {
+  cluster_id                = cockroach_cluster.my_gcp_cluster.id
+  region                    = "us-east1"
+  target_service_type       = "PRIVATE_SERVICE"
+  target_service_identifier = "projects/project/regions/us-east1/serviceAttachments/0123456789abcdef"
+}
+
+
```
angeladietz, line 41: nit: extra blank line here

chenbh (approved):
LGTM

examples/resources/cockroach_egress_private_endpoint/resource.tf
```diff
@@ -0,0 +1,110 @@
...
+  target_service_identifier = "arn:aws:kafka:us-west-2:123456789012:cluster/my-iam-cluster/87654321-4321-4321-4321-210987654321-3"
+}
+
+# Example: AWS MSK with TLS authentication
+resource "cockroach_egress_private_endpoint" "msk_tls" {
+  cluster_id                = cockroach_cluster.my_aws_cluster.id
+  region                    = "eu-west-1"
+  target_service_type       = "MSK_TLS"
+  target_service_identifier = "arn:aws:kafka:eu-west-1:123456789012:cluster/my-tls-cluster/abcdef12-abcd-abcd-abcd-abcdef123456-4"
+}
```
angeladietz, line 40: nit: since `us-east-1` is the only region in `my_aws_cluster`, i think we should use that for the examples instead of `us-west-2` and `eu-west-1`

angeladietz (commented):
Overall lgtm, but this is the first time i've looked at terraform provider code so i'll leave that to Chris for review.

internal/provider/egress_private_endpoint_resource.go
```diff
@@ -0,0 +1,554 @@
...
+	); err != nil {
+		resp.Diagnostics.AddError(
+			"Error deleting egress private endpoint",
+			fmt.Sprintf("Could not delete egress private endpoint: %s", formatAPIErrorMessage(err)),
+		)
+		return
+	}
+
+	if err := retry.RetryContext(ctx, egressPrivateEndpointDeleteTimeout,
+		waitForEgressEndpointDeletedFunc(
```
fantapop, line 275: same feedback as above.  It seems like it would be clearer to incorporate the wait into the delete function above.

internal/provider/egress_private_endpoint_resource_test.go
```diff
@@ -0,0 +1,280 @@
...
+	clusterName := fmt.Sprintf("%s-epe-%s", tfTestPrefix, GenerateRandomString(2))
+
+	cluster := client.Cluster{
+		Name: clusterName,
+	}
+
+	endpoint := client.EgressPrivateEndpoint{
+		Region:                  "us-east-2",
+		TargetServiceIdentifier: "com.amazonaws.us-east-2.logs",
+		TargetServiceType:       client.EGRESSPRIVATEENDPOINTTARGETSERVICETYPETYPE_PRIVATE_SERVICE,
```
fantapop, line 42: what a brutal generated variable name....

internal/provider/egress_private_endpoint_resource_test.go
```diff
@@ -0,0 +1,280 @@
...
+	"testing"
+
+	"github.com/cockroachdb/cockroach-cloud-sdk-go/v6/pkg/client"
+	mock_client "github.com/cockroachdb/terraform-provider-cockroach/mock"
+	"github.com/golang/mock/gomock"
+	"github.com/google/uuid"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+func TestAccEgressPrivateEndpointResource(t *testing.T) {
```
fantapop, line 31: does the acceptance test rely on cloud resources?  I didn't see anything specific so I was wondering how this would work.

fantapop (commented):
Hey nice work on this!  It's always interesting to me to see some new development patterns.  I've left a handful of questions / suggestions / nits.

internal/provider/egress_private_endpoint_resource_test.go
fantapop: nice.  Thanks for the explanation.

internal/provider/egress_private_endpoint_resource.go
```diff
@@ -0,0 +1,554 @@
...
+			},
+			"endpoint_address": schema.StringAttribute{
+				MarkdownDescription: "endpoint_address is the cloud-provider generated address (either IP address or DNS name) through which the endpoint is accessible",
+				Computed:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"state": schema.StringAttribute{
+				MarkdownDescription: "state represents the current state of the endpoint. Potential values are:" + formatEnumMarkdownList(client.AllowedEgressPrivateEndpointStateTypeEnumValues),
```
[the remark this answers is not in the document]
fantapop: I see that this one is computed, so in retrospect, using the word "potential" as opposed to "allowed" makes sense.

internal/provider/egress_private_endpoint_resource.go
```diff
@@ -0,0 +1,554 @@
...
+
+	var state EgressPrivateEndpoint
+	diags := req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Get all endpoints with pagination
+	allEndpoints, err := getAllEgressEndpointsForCluster(ctx, r.provider.service, state.ClusterID.ValueString())
```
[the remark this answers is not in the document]
fantapop: @mwstobo are we able to use the new endpoint here now?

internal/provider/egress_private_endpoint_resource.go
```diff
@@ -0,0 +1,554 @@
...
+		}
+
+		traceAPICall("CreateEgressPrivateEndpoint")
+		endpoint, httpResp, createErr := cl.CreateEgressPrivateEndpoint(
+			ctx,
+			clusterID,
+			&createReq,
+		)
+		if createErr != nil {
+			if httpResp != nil && httpResp.StatusCode == http.StatusConflict {
```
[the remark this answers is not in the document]
fantapop: @mwstobo any update on this?  I see the source code here is still checking for Unavailable in various places.  I also recently noticed we're inconsistent with how treat locking errors and filed this. https://cockroachlabs.atlassian.net/browse/CC-33641 which I hope someone on CP will collaborate on.   if we ended up using 409, I'd be fine just checking the code because in any case, I think we'd want to handle 409s the same.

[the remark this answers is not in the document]
mwstobo: Ah sorry, I misunderstood your first message, thinking we'd need to check the response body if we kept using 409. I'll check for "lock" in the response body as well.

As for the attached ticket, I'm not certain we could use 409 for cluster locks exclusively, as we do return it in other cases (conflicting private endpoints for example). I agree though, we should definitely land on a clearer approach!

internal/provider/egress_private_endpoint_resource.go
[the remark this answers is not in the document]
mwstobo: https://github.com/cockroachdb/cockroach-cloud-sdk-go/pull/81

internal/provider/egress_private_endpoint_resource.go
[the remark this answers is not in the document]
mwstobo: Sorry, the response body change got lost in the shuffle. I've updated it now!

internal/provider/egress_private_endpoint_resource.go
```diff
@@ -0,0 +1,504 @@
...
+	clusterID string,
+	endpointID string,
+) retry.RetryFunc {
+	return func() *retry.RetryError {
+		endpoint, err := getEgressPrivateEndpoint(ctx, cl, clusterID, endpointID)
+		if err != nil {
+			return retry.RetryableError(
+				fmt.Errorf("encountered an error while getting egress private endpoint, retrying: %s", err.Error()),
+			)
+		}
```
fantapop, line 475: sorry, I must have missed this before, usually we stop retrying on client errors in general with some exceptions (like removing resources if not found, or retrying if locked, etc).  It looks like getEgressPrivateEndpoints is returning standard errors for any client error (except not found) and those will end up being retried.  For example, if the service account didn't have access to the cluster we would return a 403 and it would continue to be retried.

jhlodin (approved):
Suggested some edits to the generated docs and accompanying source, otherwise docs LGTM

docs/resources/egress_private_endpoint.md
```diff
@@ -0,0 +1,173 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "cockroach_egress_private_endpoint Resource - terraform-provider-cockroach"
+subcategory: ""
+description: |-
+  Egress private endpoint for a CockroachDB Cloud cluster. Egress private endpoints allow CockroachDB Cloud clusters to connect to external services through private network connections without traversing the public internet.
+  - Terraform resource docs https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/docs/resources/egress_private_endpoint.md
```
jhlodin, line 7: Suggest renaming to "Terraform resource docs source" if we want to keep this link at all

docs/resources/egress_private_endpoint.md
```diff
@@ -0,0 +1,173 @@
...
+  - Cockroach Cloud docs https://www.cockroachlabs.com/docs/cockroachcloud/egress-private-endpoints.html
+---
+
+# cockroach_egress_private_endpoint (Resource)
+
+Egress private endpoint for a CockroachDB Cloud cluster. Egress private endpoints allow CockroachDB Cloud clusters to connect to external services through private network connections without traversing the public internet.
+- [Terraform resource docs](https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/docs/resources/egress_private_endpoint.md)
```
jhlodin, line 14: Suggest renaming to "Terraform resource docs source" if we want to keep this link at all

PR #301, opened 2025-08-26
[CNSL-659] Add error message for Cloudwatch on Basic plan

jaywritescode:
Resolves [CNSL-659](https://cockroachlabs.atlassian.net/browse/CNSL-659).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed an issue where enabling log export configs did not account for lock errors.
+- Exporting Cloudwatch metrics on a serverless Cluster no longer generates an error.
+- Exporting Cloudwatch metrics while on the Basic plan now generates an error.
 
 ## [1.13.1]
 
--- a/internal/provider/metric_export_cloudwatch_config_resource.go
+++ b/internal/provider/metric_export_cloudwatch_config_resource.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
-	"time"
 
 	"github.com/cockroachdb/cockroach-cloud-sdk-go/v6/pkg/client"
 	"github.com/hashicorp/terraform-plugin-framework/path"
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.13.2]
```
fantapop, line 8: There are a bunch of tf changes being worked on right now.  Let's remove this line and bundle it with a larger deploy.

fantapop (commented):
Hey this is looking good @jaywritescode !  As part of this can you also update the tests to use standard clusters instead of an advanced cluster?  We generally prefer to use standard where possible because they are quicker to run.

internal/provider/metric_export_cloudwatch_config_resource_test.go
```diff
@@ -61,7 +61,7 @@ func TestIntegrationMetricExportCloudWatchConfigResource(t *testing.T) {
 		Id:               clusterID,
 		Name:             clusterName,
 		CockroachVersion: "v22.2.0",
-		Plan:             "ADVANCED",
+		Plan:             "STANDARD",
 		CloudProvider:    "AWS",
 		State:            "CREATED",
 		Config: client.ClusterConfig{
```
fantapop, line 67: There are a few more updates you'll need to change this to a STANDARD cluster as the parameters are different.  Here is a pretty minimal example I think: https://github.com/cockroachdb/terraform-provider-cockroach/blob/main/internal/provider/private_endpoint_connection_resource_test.go#L112-L130

jaywritescode: Updated, I believe. But please let me know if there are other settings that are incompatible with a Standard cluster so I can update them too. Thanks!

jaywritescode:
Figured out the problem — I forgot to `go install .` the project into my local `go/bin` directory. With that fix, I'm seeing the `CloudWatch metric export services are not available on the Basic plan` error message.

internal/provider/metric_export_cloudwatch_config_resource.go
```diff
@@ -185,7 +185,7 @@ func (r *metricExportCloudWatchConfigResource) Create(
 
 func retryEnableCloudWatchMetricExport(
 	ctx context.Context,
-	timeout time.Duration,
+	_ time.Duration,
```
fantapop, line 188: If it's no longer used we can drop it from the function signature

jaywritescode: Done.

PR #281, opened 2025-05-22
Expose private_endpoint_dns in region list

sorliem:
Expose private_endpoint_dns in the region list. It is necessary for creating a Google PSC connection.

Closes https://github.com/cockroachdb/terraform-provider-cockroach/issues/280

**Commit checklist**
- [ ] Changelog
- [x] Doc gen (`make generate`)
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

PR #305, opened 2025-09-02
[CNSL-1406] Expose private_endpoint_dns in cluster region list

nicholas-tao:
Previously, `private_endpoint_dns` was not exposed in the cluster region list. Since it is used to connect a cluster to GCP Private Service Connect, customers needed to manually retrieve the `private_endpoint_dns` from the CockroachCloud UI or API, breaking end-to-end automation.

This change adds `private_endpoint_dns` to the cluster region list to allow customers to directly access it via Terraform.

This commit builds on #281.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/internal/provider/cluster_resource_test.go
+++ b/internal/provider/cluster_resource_test.go
@@ -1069,7 +1069,7 @@ func TestIntegrationServerlessClusterResource(t *testing.T) {
 			initialCluster: singleRegionClusterWithUnlimited("BASIC"),
 			updateStep: func() resource.TestStep {
 				return resource.TestStep{
-					Config:      getTestDedicatedClusterResourceConfig(clusterName, latestClusterMajorVersion, false, 4, nil),
+					Config:      getTestDedicatedClusterResourceConfig(clusterName, latestClusterMajorVersion, false, 4, nil, false),
 					ExpectError: regexp.MustCompile("Cannot update cluster plan type"),
 				}
 			},
@@ -1121,7 +1121,7 @@ func TestIntegrationServerlessClusterResource(t *testing.T) {
 		{
 			name: "use serverless primary region on dedicated cluster",
 			createStep: func() resource.TestStep {
-				config := getTestDedicatedClusterResourceConfig(clusterName, latestClusterMajorVersion, false, 4, nil)
+				config := getTestDedicatedClusterResourceConfig(clusterName, latestClusterMajorVersion, false, 4, nil, false)
 				config = strings.Replace(config, "node_count: 1", "primary: true", -1)
 				return resource.TestStep{
 					Config:      config,
@@ -1303,7 +1303,7 @@ func TestIntegrationServerlessClusterResource(t *testing.T) {
 			name: "attempt to update an advanced cluster to a standard cluster in place",
 			createStep: func() resource.TestStep {
 				return resource.TestStep{
-					Config: getTestDedicatedClusterResourceConfig(clusterName, latestClusterMajorVersion, false, 4, nil),
+					Config: getTestDedicatedClusterResourceConfig(clusterName, latestClusterMajorVersion, false, 4, nil, false),
 				}
 			},
 			initialCluster: client.Cluster{
@@ -1831,8 +1831,9 @@ func TestIntegrationDedicatedClusterResource(t *testing.T) {
 		},
 		Regions: []client.Region{
 			{
-				Name:      "us-central1",
-				NodeCount: 1,
+				Name:               "us-central1",
+				NodeCount:          1,
+				PrivateEndpointDns: "test-private-endpoint-dns.gcp-us-central1.crdb-test.io",
 			},
 		},
 		CidrRange: "172.28.0.0/16",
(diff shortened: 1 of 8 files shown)


linhcrl:
@jhlodin there are some docs changes here. Could you please take a look and let us know if you have any feedback?

internal/provider/cluster_resource_test.go
linhcrl, line 1: Not too familiar with private endpoints but you will likely have to actually modify the `getTestDedicatedClusterResourceConfig` to create a terraform cluster config that has private endpoints. @fantapop can you advise on how this works?

internal/provider/cluster_resource_test.go
```diff
@@ -2107,6 +2149,7 @@ resource "cockroach_cluster" "test" {
 	  num_virtual_cpus = %d
 	  cidr_range = "172.28.0.0/16"
 	  supports_cluster_virtualization = true
+	  private_network_visibility = true
```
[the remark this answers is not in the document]
nicholas-tao: turns out we don't actually need to set `private_network_visibility` so i removed it from the tests

internal/provider/cluster_resource_test.go
```diff
@@ -2090,13 +2122,23 @@ func testCheckCockroachClusterExists(resourceName string) resource.TestCheckFunc
...
 ) string {
 	var deleteProtectionConfig string
 	if deleteProtectionEnabled != nil {
 		deleteProtectionConfig = fmt.Sprintf("\ndelete_protection = %t\n", *deleteProtectionEnabled)
 	}
 
+	var privateEndpointServicesResource string
+	var dataSourceDependsOn string
+	if privateEndpointServicesEnabled != nil && *privateEndpointServicesEnabled {
+		privateEndpointServicesResource = `resource "cockroach_private_endpoint_services" "test" {
```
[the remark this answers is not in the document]
linhcrl: Also, `depends_on = [cockroach_cluster.test]` is not needed. By setting the cluster_id to the ID of the `cockroach_cluster.test` resource, you are implicitly creating a dependency

internal/provider/cluster_resource_test.go
```diff
@@ -2001,12 +2016,18 @@ func testDedicatedClusterResource(
...
 				resource.TestCheckResourceAttr(resourceName, "cockroach_version", latestClusterMajorVersion),
 				resource.TestCheckResourceAttr(dataSourceName, "cockroach_version", latestClusterMajorVersion),
 				resource.TestMatchResourceAttr(resourceName, "full_version", startsWithLatestMajorVersionRE),
 				resource.TestMatchResourceAttr(dataSourceName, "full_version", startsWithLatestMajorVersionRE),
+
+				// We check for private_endpoint_dns here because it's not available during initial
+				// cluster creation. The private_endpoint_dns field is only populated after a
+				// PrivateEndpointService resource is created, which happens after the cluster is created.
+				resource.TestCheckResourceAttrSet(resourceName, "regions.0.private_endpoint_dns"),
+				resource.TestCheckResourceAttrSet(dataSourceName, "regions.0.private_endpoint_dns"),
```
linhcrl, line 2030: thanks for leaving this comment here. It's super helpful.

internal/provider/cluster_resource_test.go
```diff
@@ -2034,8 +2055,17 @@ func testDedicatedClusterResource(
...
+			Config: getTestDedicatedClusterResourceConfig(clusterName, latestClusterMajorVersion, false, 4, ptr(true), true),
+			Check: resource.ComposeTestCheckFunc(
+				resource.TestCheckResourceAttr(resourceName, "delete_protection", "true"),
+
+				// This state check is necessary because the test configuration creates a PrivateEndpointService
+				// resource, which temporarily locks the cluster (state becomes "LOCKED") during creation.
+				// After the PrivateEndpointService creation completes, the cluster state returns to "CREATED".
+				// Without this check, Terraform detects the state change from "LOCKED" to "CREATED" as an
+				// inconsistent result and fails the test with "Provider produced inconsistent result after apply".
+				resource.TestCheckResourceAttr(resourceName, "state", "CREATED"),
```
linhcrl, line 2067: I have a [PR](https://github.com/cockroachdb/terraform-provider-cockroach/pull/308) that should handle the issue mentioned in the highlighted comment. 

Until my PR merges, you could incorporate my change from `internal/provider/cluster_resource.go` in my PR and see if it resolves it. These lines shouldn't be necessary

internal/provider/cluster_resource_test.go
```diff
@@ -2090,13 +2120,24 @@ func testCheckCockroachClusterExists(resourceName string) resource.TestCheckFunc
...
 		deleteProtectionConfig = fmt.Sprintf("\ndelete_protection = %t\n", *deleteProtectionEnabled)
 	}
 
+	var privateEndpointServicesResource string
+	var dataSourceDependsOn string
+	if privateEndpointServicesEnabled {
+		privateEndpointServicesResource =
+			`resource "cockroach_private_endpoint_services" "test" {
+				cluster_id = cockroach_cluster.test.id
+				depends_on = [cockroach_cluster.test]
```
linhcrl, line 2136: this line is extraneous. The line right above already creates an implicit dependency

internal/provider/cluster_resource_test.go
nicholas-tao: makes sense! i'll move the data source check and re-run the acceptance test

nicholas-tao: you're right! moved the data source check to the previous test step and updated the comment.

also tried moving the resource check to the previous step too and it failed (as expected), so left that as-is.

linhcrl (approved):
LGTM! good job!

PR #308, opened 2025-09-03
remove plan modifier from `state` field in cluster resource

linhcrl:
This commit removes the `UseStateForUnknown` plan modifier from the `state` field to prevent errors when the computed field drifts due to external changes.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)


fantapop (commented):
Is there a ticket this is related to that can be added to the headline?

PR #310, opened 2025-09-11
[CNSL-1482] add acceptance test for cockroach_backups data source

linhcrl:
This commit involves adding an acceptance test for the cockroach_backups data source.

**Commit checklist**
- [x] Changelog
- [ ] Integration test(s)
- [x] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/internal/provider/backups_data_source_test.go
+++ b/internal/provider/backups_data_source_test.go
@@ -12,9 +12,7 @@ import (
 	mock_client "github.com/cockroachdb/terraform-provider-cockroach/mock"
 	"github.com/golang/mock/gomock"
 	"github.com/google/uuid"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
-	"github.com/hashicorp/terraform-plugin-testing/terraform"
 )
 
 // TestAccBackupsDataSource attempts to list backups for a real cluster.
@@ -27,7 +25,10 @@ func TestAccBackupsDataSource(t *testing.T) {
 	p.service = NewService(cl)
 	clusterName := fmt.Sprintf("%s-cluster-with-backups-%s", tfTestPrefix, GenerateRandomString(4))
 
-	endTime := time.Now().Truncate(time.Second).UTC()
+	// Use 2 days from now as the end time so that date-only inputs (without timestamp)
+	// still include all of today’s data and the test passes regardless of what time of
+	// day it's run.
+	endTime := time.Now().Add(2 * 24 * time.Hour).Truncate(time.Second).UTC()
 	startTime := endTime.Add(-3 * 24 * time.Hour)
 
 	testBackupsDataSource(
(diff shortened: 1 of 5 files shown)


internal/provider/backups_data_source_test.go
```diff
@@ -27,7 +25,9 @@ func TestAccBackupsDataSource(t *testing.T) {
 	p.service = NewService(cl)
 	clusterName := fmt.Sprintf("%s-cluster-with-backups-%s", tfTestPrefix, GenerateRandomString(4))
 
-	endTime := time.Now().Truncate(time.Second).UTC()
+	// Use tomorrow as the end time so that date-only inputs (without timestamp)
+	// still include all of today’s data.
+	endTime := time.Now().Add(1 * 24 * time.Hour).Truncate(time.Second).UTC()
 	startTime := endTime.Add(-3 * 24 * time.Hour)
```
fantapop, line 31: I'm unclear if it will be a problem but If the tests run at certain times of day, it will span 2 days right?

internal/provider/backups_data_source_test.go
```diff
@@ -134,31 +136,9 @@ func testBackupsDataSource(
 				PreConfig: func() {
 					traceMessageStep("creating cluster and waiting for backups to become available")
 				},
-				Config: getTestBackupsDataSourceConfig(clusterName, startTime.Format(time.DateOnly), endTime.Format(time.DateOnly)),
+				Config: testGetStandardClusterConfig(clusterName, true /*frequentBackup*/),
 				Check: resource.ComposeTestCheckFunc(
```
fantapop, line 140: You can remove these when there is only one TestCheck function

internal/provider/restores_data_source_test.go
```diff
@@ -55,56 +84,81 @@ func TestIntegrationRestoresDataSource(t *testing.T) {
...
-	s.EXPECT().GetBackupConfiguration(gomock.Any(), clusterId).
+	s.EXPECT().GetBackupConfiguration(gomock.Any(), clusterID).
 		Return(backupConfig, httpOkResponse, nil).AnyTimes()
+
+	// Kick off restore operation.
+	s.EXPECT().CreateRestore(gomock.Any(), clusterID, gomock.Any()).Return(restorePending, httpOkResponse, nil)
+	// Restore operation is pending.
+	s.EXPECT().GetRestore(gomock.Any(), clusterID, restoreID).Return(restorePending, httpOkResponse, nil).Times(3)
+	// Restore operation is successful.
+	s.EXPECT().GetRestore(gomock.Any(), clusterID, restoreID).Return(restoreSuccess, httpOkResponse, nil).AnyTimes()
```
fantapop, line 132: for GetRestore and ListRestores below can you figure out the number of times it's called and put in a number? I'm surprised this is working as is actually.  Supposedly gomock does not guarantee the order of mocks that are called.  Since these calls have the same parameters, I think you should use [gomock.InOrder](https://pkg.go.dev/github.com/golang/mock/gomock#InOrder).

internal/provider/backups_data_source_test.go
linhcrl: yep. You mean if it runs at like 23:59:59 right? I could add two days to be safe if you'd like

internal/provider/restores_data_source_test.go
linhcrl: Updated

PR #335, opened 2026-04-20
Fix role grants Create to use API response for state

adityamCRL:
The Create function discarded the SetRolesForUser API  response and saved plan data directly to state. If the API normalized  anything (field ordering, defaults), state wouldn't match what Read  returns, causing drift. Fixed by capturing the response and calling  loadRolesToTerraformState, matching the Update function's existing behavior.                                                                     
                                                                              
The Read function used ListRoleGrants, a paginated endpoint that scans ALL users in the org, Then searched for the target user ID across pages. The managed-service PaginateLocal implementation uses binary search for page navigation on unsorted data, which can miss entries for orgs with 100+ users. When the user wasn't found, Read         removed the resource from state, causing Terraform to want to recreate it.                                                                           
                                                                                
Fixed by replacing the ListRoleGrants pagination loop with a direct GetAllRolesForUser(userId) call. This matches the approach already used by the singular user_role_grant resource.

The April 17 CC deploy likely added users/service accounts to the test organization (from MFA feature work), pushing the total past the 100 user pagination boundary where the binary search bug manifests. On April 15 there were fewer users, so the first page contained everyone and pagination wasn't needed.

**Commit checklist**
- [x] Changelog
- [ ] Doc gen (`make generate`) (N/a no schema changes).
- [x] Integration test(s)
- [x] Acceptance test(s)
- [ ] Example(s) (N/a bug fix).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `s3_vpc_endpoint_id` computed attribute to cluster regions, exposing the AWS S3 VPC gateway endpoint ID for configuring S3 bucket policies. Only populated for Advanced clusters on AWS.
 
+### Fixed
+
+- Fixed `cockroach_user_role_grants` resource drift on refresh after creation. The Create function now uses the API response to populate state, matching the Update function's behavior.
+- Fixed `cockroach_user_role_grants` Read to use a direct user lookup instead of paginated full-org scan, preventing state removal on large orgs.
+
 ## [1.19.0] - 2026-04-15
 
 ### Changed
--- a/internal/provider/user_role_grants_resource.go
+++ b/internal/provider/user_role_grants_resource.go
@@ -29,8 +29,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
-const roleGrantPaginationLimit = 100
-
 type userRoleGrantsResource struct {
 	provider *provider
 }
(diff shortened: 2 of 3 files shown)


internal/provider/user_role_grants_resource.go
```diff
@@ -168,51 +167,37 @@ func (r *userRoleGrantsResource) Read(
...
-				state.ID = state.UserId
-				diags = resp.State.Set(ctx, state)
-				resp.Diagnostics.Append(diags...)
-				return
-			}
-		}
+	// Look up the user's roles directly by user ID rather than scanning
+	// the full paginated list of all users in the org via ListRoleGrants.
+	// The direct lookup is both more efficient and avoids potential issues
+	// with ListRoleGrants pagination on large orgs.
```
mwstobo: nit: I think we should remove the reference to the previous implementation, as it won't mean anything to readers outside of this commit.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `s3_vpc_endpoint_id` computed attribute to cluster regions, exposing the AWS S3 VPC gateway endpoint ID for configuring S3 bucket policies. Only populated for Advanced clusters on AWS.
 
+### Fixed
+
+- Fixed `cockroach_user_role_grants` resource drift on refresh after creation. The Create function now uses the API response to populate state, matching the Update function's behavior.
```
mwstobo, line 16: nit: I think we should leave a note about fixing how user role grants are read as well!

mwstobo (commented):
Just a couple small non-functional nits!

CHANGELOG.md
adityamCRL: Good catch, the initial fix was just for the Create function, I added the read update after and forget to update the changelog.

internal/provider/user_role_grants_resource.go
adityamCRL: Good idea, I just simplified it to a one line comment just explaining what the call does.

mwstobo (approved):
Thanks for the fix!

adityamCRL:
TFTR!

PR #337, opened 2026-04-28
Add post test cleanup for orphaned acceptance test clusters

adityamCRL:
This PR creates a GitHub Actions script to clean up real clusters created during acceptance test runs, on both successful and failed runs. The post-run cleanup script is idempotent: clusters already cleaned up as part of a successful acceptance test run result in a no-op.

When acceptance tests time out (Go's -timeout 120m kills all goroutines without running defer blocks), clusters created during the test run are left orphaned in the testing org. This adds automatic cleanup that runs after every acceptance test workflow, whether tests pass, fail, or time out.

The provider now tracks cluster IDs in a file during test runs via `CLEANUP_TRACKING_FILE`. The tracking wrapper is only constructed in provider.Configure when this env var is set, so customer runs of the provider are unaffected. A post-test cleanup step (if: always()) reads the tracking file and deletes any clusters that were created but not destroyed. Each workflow run uses its own tracking file, so concurrent runs only clean up their own clusters.

**Commit checklist**
- [ ] Changelog (N/a internal CI change)
- [ ] Doc gen (`make generate`) (N/a no schema changes)
- [x] Integration test(s)
- [x] Acceptance test(s) [Link](https://github.com/cockroachdb/terraform-provider-cockroach/actions/runs/25071942710) shows the logs with the new script working.
- [ ] Example(s) (N/a no new resources)

### DIFF
--- a/scripts/cleanup-test-clusters.sh
+++ b/scripts/cleanup-test-clusters.sh
@@ -1,0 +1,220 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# GitHub Actions log annotations.
+log_error()    { echo "::error title=${2:-Error}::$1"; }
+log_warning()  { echo "::warning title=${2:-Warning}::$1"; }
+log_notice()   { echo "::notice title=${2:-Notice}::$1"; }
+log_group()    { echo "::group::$*"; }
+log_endgroup() { echo "::endgroup::"; }
+
+# Write to the GitHub Actions step summary (no-op outside CI).
+write_summary() {
+  if [[ -n "${GITHUB_STEP_SUMMARY:-}" ]]; then
+    echo "$1" >> "$GITHUB_STEP_SUMMARY"
+  fi
+}
+
+# Verify required commands are available.
+for cmd in jq curl; do
+  if ! command -v "$cmd" >/dev/null 2>&1; then
+    log_error "Required command not found: ${cmd}" "Missing Dependency"
+    write_summary "### Cluster Cleanup"$'\n'"**Error:** Required command \`${cmd}\` not found."
+    exit 1
+  fi
+done
+
+TRACKING_FILE="${CLEANUP_TRACKING_FILE:-}"
+API_URL="${COCKROACH_SERVER:-https://cockroachlabs.cloud}"
+DRY_RUN="${DRY_RUN:-0}"
+MAX_RETRIES=3
+RETRY_DELAY=5
+
+if [[ -z "${COCKROACH_API_KEY:-}" ]]; then
+  log_error "COCKROACH_API_KEY is not set" "Configuration"
+  write_summary "### Cluster Cleanup"$'\n'"**Error:** COCKROACH_API_KEY is not set."
+  exit 1
+fi
+
+auth_header="Authorization: Bearer ${COCKROACH_API_KEY}"
+response_file=$(mktemp)
+
+cleanup_temp_files() {
+  rm -f "$response_file"
+  if [[ -n "$TRACKING_FILE" && -f "$TRACKING_FILE" ]]; then
+    rm -f "$TRACKING_FILE"
+    echo "Removed tracking file: ${TRACKING_FILE}"
+  fi
+}
+trap cleanup_temp_files EXIT
+
+echo "=== Cleanup Orphaned Test Clusters ==="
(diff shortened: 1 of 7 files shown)


scripts/cleanup-test-clusters.sh
```diff
@@ -0,0 +1,252 @@
...
+
+PREFIX="tftest"
+API_URL="${COCKROACH_SERVER:-https://cockroachlabs.cloud}"
+DRY_RUN="${DRY_RUN:-0}"
+TRACKING_FILE="${CLEANUP_TRACKING_FILE:-}"
+MAX_RETRIES=3
+RETRY_DELAY=5
+
+if [[ -z "${COCKROACH_API_KEY:-}" ]]; then
+  echo "ERROR: COCKROACH_API_KEY is not set"
```
fantapop: Since this is being used from a github action it would be nice to use the syntax to get their error highlighting.  For example: 

echo "::error:: COCKROACH_API_KEY is not set"

or maybe add a function like log_error()

We do something similar here: https://github.com/cockroachdb/actions/blob/e95bfaf2ef50f0f02456f8f9a7bdca8021f792db/actions_helpers.sh#L5

scripts/cleanup-test-clusters.sh
```diff
@@ -0,0 +1,252 @@
...
+  local page=""
+  while true; do
+    local url="${API_URL}/api/v1/clusters?pagination.limit=100"
+    if [[ -n "$page" ]]; then
+      url="${url}&pagination.page=${page}"
+    fi
+
+    local response
+    response=$(curl -sf -H "$auth_header" "$url") || {
+      echo "ERROR: Failed to list clusters from API" >&2
```
fantapop: error notation?

scripts/cleanup-test-clusters.sh
```diff
@@ -0,0 +1,252 @@
...
+    page=$(echo "$response" | jq -r '.pagination.next_page // empty')
+    if [[ -z "$page" ]]; then
+      break
+    fi
+  done
+
+  echo "$all_clusters"
+}
+
+# Determine mode: targeted (tracking file) or fallback (list all tftest-*).
```
fantapop: It seems like the fallback mode would interfere with concurrently running tests is it really necessary?  We could have an additional scheduled action that deletes clusters over a day old.

fantapop (commented):
I'm not quite done, but left a bit of feedback for you.

scripts/cleanup-test-clusters.sh
adityamCRL: Ah thank you I didn't know about this, I've updated the syntax with the functions like in the example you gave me and updated more call sites throughout the code.

scripts/cleanup-test-clusters.sh
adityamCRL: Fixed, using a `log_error()` now.

scripts/cleanup-test-clusters.sh
adityamCRL: I agree with this the concurrent test issue could be a huge issue with the file fallback path.

As for your second point, I thought of something similar but as a cron job that would regularly clean out clusters that started with the prefix `tftest-*` in that org. I could do that in a separate follow up PR if you think that's a good idea/

fantapop (commented):
Hey, sorry I lost track of this.  Here's what I'm thinking:
* it would be nice to not run this code when customers run the provider.  Can we check an env var to choose the tracking service instead?
* I'm concerned that the fallback deletion could disturb runs concurrently.  Can we just do the file based one as a best effort?  We still have the cluster minder to catch stragglers.

adityamCRL:
@fantapop I'm thinking on the exact same lines as you. 

For your first point, I've updated the code so this will never run when the customer runs the provider:
```
svc := NewService(cl)
if filePath := os.Getenv(clusterTrackingEnvVar); filePath != "" {
	svc = newTrackingService(svc, filePath)
}
p.service = svc
```

For the second point, the code actually never had a fallback path, if the tracking file was not found for whatever reason there would be nothing to clean up, so the entire cleanup method is just a best effort attempt using the file tracking method. I've updated the PR description to be cleaner.

fantapop (approved):
LGTM

PR #338, opened 2026-05-07
[CNSL-2196] add with_empty_ip_allowlist option for serverless cluster creation

jaywritescode:
Allow customers to opt out of the default open IP allowlist (0.0.0.0/0) when creating serverless clusters. The field is create-only and cannot be changed after cluster creation.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)


internal/provider/cluster_resource.go
```diff
@@ -263,6 +263,14 @@ func (r *clusterResource) Schema(
...
 							formatEnumMarkdownList(AllowedUpgradeTypeTypeEnumValueStrings),
 					},
+					"with_empty_ip_allowlist": schema.BoolAttribute{
+						Optional: true,
+						Computed: true,
+						PlanModifiers: []planmodifier.Bool{
+							boolplanmodifier.UseStateForUnknown(),
+						},
+						MarkdownDescription: "Set to true to create the cluster with no IP allowlist entries. By default, serverless clusters are created with a default open allowlist (0.0.0.0/0) that allows connections from any IP address. This can only be set during cluster creation and cannot be changed later.",
+					},
```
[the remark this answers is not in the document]
jaywritescode: Changed to `Computed: false`.

internal/provider/cluster_data_source.go
```diff
@@ -138,6 +138,10 @@ func (d *clusterDataSource) Schema(
 						Computed:    true,
 						Description: "Dictates the behavior of CockroachDB major version upgrades.",
 					},
+					"with_empty_ip_allowlist": schema.BoolAttribute{
+						Computed:            true,
+						MarkdownDescription: "Whether the cluster was created with no IP allowlist entries.",
+					},
```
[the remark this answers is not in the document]
jaywritescode: No longer defaults to false. Leaves the field as null/unknown when plan is unavailable (import/data source).

internal/provider/cluster_resource.go
```diff
@@ -1515,6 +1534,12 @@ func loadClusterToTerraformState(
 			}
 		}
 
+		if plan != nil && plan.ServerlessConfig != nil && IsKnown(plan.ServerlessConfig.WithEmptyIpAllowlist) {
+			serverlessConfig.WithEmptyIpAllowlist = plan.ServerlessConfig.WithEmptyIpAllowlist
+		} else {
+			serverlessConfig.WithEmptyIpAllowlist = types.BoolValue(false)
```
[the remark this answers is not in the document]
jaywritescode: Kept the attribute as Computed: true (required for struct compatibility, will be null since the API can't return it: same pattern as `supports_cluster_virtualization`)

internal/provider/cluster_resource.go
```diff
@@ -962,6 +973,13 @@ func (r *clusterResource) ModifyPlan(
 				"Cannot update whether a cluster supports cluster virtualization",
 				"Changing the supports_cluster_virtualization field of a cluster is not allowed. Please explicitly destroy this cluster before changing this value.")
 		}
+		if serverless := plan.ServerlessConfig; serverless != nil &&
+			state.ServerlessConfig != nil &&
+			serverless.WithEmptyIpAllowlist != state.ServerlessConfig.WithEmptyIpAllowlist {
```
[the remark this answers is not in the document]
jaywritescode: Now compares `ValueBool()` so null and false are treated equivalently — users can freely add/remove `with_empty_ip_allowlist = false` without hitting the guard.

docs/data-sources/cluster.md
```diff
@@ -134,6 +134,7 @@ Read-Only:
 - `spend_limit` (Number, Deprecated) Spend limit in US cents.
 - `upgrade_type` (String) Dictates the behavior of CockroachDB major version upgrades.
 - `usage_limits` (Attributes) (see [below for nested schema](#nestedatt--serverless--usage_limits))
+- `with_empty_ip_allowlist` (Boolean) Whether the cluster was created with no IP allowlist entries.
```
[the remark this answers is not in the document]
jaywritescode: Auto-regenerated with the updated description.

internal/provider/cluster_data_source.go
```diff
@@ -138,6 +138,10 @@ func (d *clusterDataSource) Schema(
 						Computed:    true,
 						Description: "Dictates the behavior of CockroachDB major version upgrades.",
 					},
+					"with_empty_ip_allowlist": schema.BoolAttribute{
+						Computed:            true,
+						MarkdownDescription: "Whether the cluster was created with no IP allowlist entries.",
```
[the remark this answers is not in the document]
jaywritescode: Clarifies the field may be null for data sources and imported resources since the API doesn't return it.

docs/resources/cluster.md
```diff
@@ -82,6 +82,21 @@ resource "cockroach_cluster" "basic" {
...
+  serverless = {
+    with_empty_ip_allowlist = true
+  }
+  regions = [
+    {
+      name = "us-east1"
+    }
+  ]
+  delete_protection = false
+}
```
fantapop: I'm not sure about the wording but I think a better example would include an additional allowlist. Maybe something like:

```suggestion
resource "cockroach_cluster" "basic_locked_down" {
  name           = "cockroach-basic-restricted"
  cloud_provider = "GCP"
  plan           = "BASIC"
  serverless = {
    with_empty_ip_allowlist = true
  }
  regions = [
    {
      name = "us-east1"
    }
  ]
  delete_protection = false
}
resource "cockroach_allow_list" "home" {
  name       = "home"
  cidr_ip    = "123.123.1.1"
  cidr_mask  = 32
  ui         = true
  sql        = true
  cluster_id = cockroach_cluster.basic_locked_down.id
}
```

internal/provider/cluster_data_source.go
```diff
@@ -138,6 +138,10 @@ func (d *clusterDataSource) Schema(
 						Computed:    true,
 						Description: "Dictates the behavior of CockroachDB major version upgrades.",
 					},
+					"with_empty_ip_allowlist": schema.BoolAttribute{
+						Computed:            true,
+						MarkdownDescription: "Whether the cluster was created with no IP allowlist entries. Not returned by the API, so this value may be null for data sources and imported resources.",
+					},
```
fantapop, line 144: This should actually just be removed from the data source.  It doesn't seem to serve any purpose. There's no rule that the data source needs to support everything the resource does.

internal/provider/cluster_resource.go
```diff
@@ -263,6 +263,13 @@ func (r *clusterResource) Schema(
 						MarkdownDescription: "Dictates the behavior of CockroachDB major version upgrades. Manual upgrades are not supported on CockroachDB Basic. Manual or automatic upgrades are supported on CockroachDB Standard. If you omit the field, it defaults to `AUTOMATIC`. Allowed values are:" +
 							formatEnumMarkdownList(AllowedUpgradeTypeTypeEnumValueStrings),
 					},
+					"with_empty_ip_allowlist": schema.BoolAttribute{
+						Optional: true,
+						PlanModifiers: []planmodifier.Bool{
+							boolplanmodifier.UseStateForUnknown(),
+						},
+						MarkdownDescription: "Set to true to create the cluster with no IP allowlist entries. By default, serverless clusters are created with a default open allowlist (0.0.0.0/0) that allows connections from any IP address. This can only be set during cluster creation and cannot be changed later.",
+					},
```
fantapop: * It doesn't seem like we're using any Markdown here.  I think `Description` is an option if it's just plain text.
* serverless is the old terminology.  You should refer to BASIC and STANDARD clusters.  @jhlodin can you look at the strings here?

internal/provider/cluster_resource.go
```diff
@@ -962,6 +973,13 @@ func (r *clusterResource) ModifyPlan(
 				"Cannot update whether a cluster supports cluster virtualization",
 				"Changing the supports_cluster_virtualization field of a cluster is not allowed. Please explicitly destroy this cluster before changing this value.")
 		}
+		if serverless := plan.ServerlessConfig; serverless != nil &&
+			state.ServerlessConfig != nil &&
+			serverless.WithEmptyIpAllowlist.ValueBool() != state.ServerlessConfig.WithEmptyIpAllowlist.ValueBool() {
+			resp.Diagnostics.AddError(
+				"Cannot update with_empty_ip_allowlist",
+				"The with_empty_ip_allowlist field can only be set during cluster creation and cannot be changed afterward. Please explicitly destroy this cluster before changing this value.")
+		}
```
fantapop, line 983: I feel like we should allow removing this field but not adding it.  The reality is that how the cluster was created is just noise later on.  They could have added the default back and removed it again so it's inconsequential.

fantapop (commented):
I left a handful of feedback here. It would be nice if we could get some acceptance tests for this.  Was there a reason that you didn't add that?

docs/resources/cluster.md
jaywritescode: Updated.

internal/provider/cluster_data_source.go
jaywritescode: Unfortunately removing it from the data source produces a runtime error: "Struct defines fields not found in object: with_empty_ip_allowlist". 

Claude says that in order to remove this field from the data source, we would need to create a separate data-source-specific struct or split the shared model. Either of those would be a bigger refactor. The same situation exists for `supports_cluster_virtualization` on dedicated clusters, which is also in the data source schema as `Computed: true` for the same reason.

internal/provider/cluster_resource.go
jaywritescode: Updated, pending input from @jhlodin.

internal/provider/cluster_resource.go
jaywritescode: Updated so we can remove the field (go from true to false/nil) but not add it (go from false/nil to true).

internal/provider/cluster_data_source.go
jaywritescode: Follow up: I asked Claude what it would take to split the shared model, and it's a lot. We would have to duplicate `CockroachCluster` (17 fields), `ServerlessClusterConfig` (5 fields), `DedicatedClusterConfig` (8 fields), plus about a hundred lines of code with `loadClusterToTerraformState`.

Unfortunately, making the value a nullable pointer isn't an option either, since Terraform enforces a strict 1:1 match between struct fields and schema's object attributes.

internal/provider/cluster_resource.go
```diff
@@ -263,6 +263,13 @@ func (r *clusterResource) Schema(
 						MarkdownDescription: "Dictates the behavior of CockroachDB major version upgrades. Manual upgrades are not supported on CockroachDB Basic. Manual or automatic upgrades are supported on CockroachDB Standard. If you omit the field, it defaults to `AUTOMATIC`. Allowed values are:" +
 							formatEnumMarkdownList(AllowedUpgradeTypeTypeEnumValueStrings),
 					},
+					"with_empty_ip_allowlist": schema.BoolAttribute{
+						Optional: true,
+						PlanModifiers: []planmodifier.Bool{
+							boolplanmodifier.UseStateForUnknown(),
+						},
+						Description: "Set to true to create the cluster with no IP allowlist entries. By default, BASIC and STANDARD clusters are created with a default open allowlist (0.0.0.0/0) that allows connections from any IP address. This can only be set during cluster creation and cannot be changed later.",
```
fantapop: I think the wording here should be more like.   "This field is read only during cluster creation and ignored on update".  But maybe @jhlodin can weigh in here.

internal/provider/cluster_data_source.go
fantapop: Okay.  For now let's document that this will always be empty.  I'm thinking something like:
```
Description: "This value is not available for cluster datasource's and will always be nil",
```

@jhlodin what do you think?

internal/provider/cluster_resource.go
fantapop: wait, going from true to false seems like it should error.   But going from true to empty or false to empty is fine.

internal/provider/cluster_data_source.go
jhlodin: I'd want the description to be a little more clear that this is just internal implementation detail which is why it will be empty/should be ignored.

If it is in fact going to be null in any situation where the customer sees the field, something like:
`Description: "This field is set to null for internal reasons and can be safely ignored."`

internal/provider/cluster_resource.go
jhlodin: In practice, I'm assuming if I wanted to add an open allowlist after the cluster had been created with this value set to False I could just add allowlist entries manually?

I'd suggest something descriptive like:
> Set to true to create the cluster with no default entries to the IP allowlist, blocking inbound connections until entries are added manually. This field is not used after cluster creation; to add new allowlist entries, configure your cluster as normal.

If you can't manually add an open `0.0.0.0/0` rule after cluster creation then we'll need to call that out

jhlodin (approved):
Left suggestions, LGTM'd to not block on docs review

fantapop (approved):
This LGTM.  pending the suggestions Joe made.

PR #340, opened 2026-05-24
ci: declare workflow-level `contents: read` on ci and build workflows

arpitjain099:
Both workflows run pure Go build/test (`ci.yaml` runs lint + tests, `build.yaml` does the Go build). No GitHub API writes from either workflow.

Same post-CVE-2025-30066 (`tj-actions/changed-files`) hardening pattern. YAML validated locally.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -6,6 +6,9 @@ on:
     tags:
       - 'v*.*.*'
 
+permissions:
+  contents: read
+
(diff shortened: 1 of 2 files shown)


arpitjain099:
Consolidating to avoid a duplicate: #339 already declares the same read-only permissions on these workflow(s) (and others), so I'm closing this one in its favor. Apologies for the overlap.

PR #344, opened 2026-05-26
fix with_empty_ip_allowlist losing state on plan updates

jaywritescode:
Without Computed: true, UseStateForUnknown never fires because Terraform plans an omitted Optional-only attribute as null (not unknown). This caused the [acceptance test to fail](https://github.com/cockroachdb/terraform-provider-cockroach/actions/runs/26471447488/job/77945707870) at step 2 when the attribute was removed from config but expected to persist.

Guard the state copy with IsKnown so that on initial create (where there is no prior state and the plan value is unknown) the value resolves to null instead of leaking an unknown into state.

**Commit checklist**
- [ ] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/internal/provider/cluster_resource.go
+++ b/internal/provider/cluster_resource.go
@@ -265,6 +265,7 @@ func (r *clusterResource) Schema(
 					},
 					"with_empty_ip_allowlist": schema.BoolAttribute{
 						Optional: true,
+						Computed: true,
 						PlanModifiers: []planmodifier.Bool{
 							boolplanmodifier.UseStateForUnknown(),
 						},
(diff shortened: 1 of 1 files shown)


internal/provider/cluster_resource.go
```diff
@@ -265,6 +265,7 @@ func (r *clusterResource) Schema(
 					},
 					"with_empty_ip_allowlist": schema.BoolAttribute{
 						Optional: true,
+						Computed: true,
 						PlanModifiers: []planmodifier.Bool{
 							boolplanmodifier.UseStateForUnknown(),
```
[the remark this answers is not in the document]
jaywritescode: Fixed.

internal/provider/cluster_resource.go
```diff
@@ -1534,7 +1535,7 @@ func loadClusterToTerraformState(
 			}
 		}
 
-		if plan != nil && plan.ServerlessConfig != nil {
+		if plan != nil && plan.ServerlessConfig != nil && IsKnown(plan.ServerlessConfig.WithEmptyIpAllowlist) {
```
linhcrl, line 1541: would be nice to have a small comment right above explaining why this is necessary

linhcrl (approved):
This looks good to me with one small suggestion

internal/provider/cluster_resource.go
jaywritescode: Comment added.

PR #345, opened 2026-05-27
[CC-36163]: add enable_sending_queue option to log export configuration

aa-joshi:
This adds the `enable_sending_queue` boolean field to the `groups` block on the `cockroach_log_export_config` resource. When enabled on a log export group, logs are buffered to a persistent disk-backed sending queue in the OTel collector. This allows exported logs to survive pod restarts and temporary sink outages.

The API enforces that only one group per cluster can have `enable_sending_queue` set to true; server-side validation rejects configurations with multiple groups enabled.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `enable_sending_queue` option to the `groups` block on `cockroach_log_export_config`. When enabled, logs in the group are buffered to a persistent disk-backed queue, allowing them to survive pod restarts and temporary sink outages. Only one group per cluster can have this option enabled.
 - Added `with_empty_ip_allowlist` option to the `serverless` block on `cockroach_cluster`. Set to `true` to create a serverless cluster with no default IP allowlist entries, requiring explicit allowlist configuration. This is a create-only field and cannot be changed after cluster creation.
 
 ## [1.20.0] - 2026-04-22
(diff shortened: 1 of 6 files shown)


vendor/github.com/cockroachdb/cockroach-cloud-sdk-go/v7/pkg/client/model_log_export_group.go
```diff
@@ -105,3 +107,17 @@ func (o *LogExportGroup) GetRedact() bool {
 func (o *LogExportGroup) SetRedact(v bool) {
 	o.Redact = &v
 }
+
+// GetEnableSendingQueue returns the EnableSendingQueue field value if set, zero value otherwise.
```
arjunmahishi: Was this modified manually or done through `go get -u`?
It should be the later. I think that would result in a `go.mod` / `go.sum` change too. 

[The SDK](https://github.com/cockroachdb/cockroach-cloud-sdk-go) should be updated first (using code gen) and then we should run `go get -u` or similar command here to get the new SDK version.

aa-joshi: valid point. SDK version has been bumped up to v8.0.0 which has the `enable_sending_queue` flag changes. I am now relying on the same version instead of vendor updates.

arjunmahishi (approved):
LGMT

PR #347, opened 2026-06-01
fix with_empty_ip_allowlist allowing true→false change after creation

jaywritescode:
The validation only checked for false→true, so explicitly setting the attribute to false after creation was silently accepted.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


fantapop (commented):
I just want to verify that we can still remove it after creation.  I think the rules should be:

* you can't change from true to false or false to true
* you can remove it

jaywritescode:
I added three tests to check that true→false and false→true fail with an error, and that removing the attribute works.

PR #348, opened 2026-06-21
[CC-36355] sql_user: add write-only password_wo alongside password

souravcrl:
## Summary

Adds a write-only `password_wo` attribute (plus its `password_wo_version` rotation trigger) to the `cockroach_sql_user` resource and marks the existing `password` attribute as deprecated. The two paths coexist so consumers can migrate at their own pace; they are mutually exclusive per resource via `Conflicting` validators, and `password_wo` + `password_wo_version` are bound with `RequiredTogether`.

- **Schema**: `password` stays `Optional` + `Sensitive` and now carries a `DeprecationMessage`. `password_wo` is `Optional` + `Sensitive` + `WriteOnly` (Terraform CLI 1.11+). `password_wo_version` is an `Optional` `Int64` that acts as the rotation trigger — Terraform cannot diff a write-only value, so the integer is what makes a rotation visible.
- **Create**: reads `password_wo` from `req.Config` (write-only attributes are absent from plan and state), `password` from the plan, and falls back to a randomly generated value when neither is set. An unknown `password_wo` is rejected rather than silently producing an empty password.
- **Update**: rotates on a `password` edit (diff against state) or on a `password_wo_version` bump while `password_wo` is set in config. Clearing `password` keeps the historical warning that the underlying password is unchanged — the warning is appended to the response diagnostics but does not gate `resp.State.Set`, so the cleared value still lands in state (an early-return on the warning would have looped indefinitely).
- **Validators**: `resource.ConfigValidators` enforces `Conflicting(password, password_wo)`, `Conflicting(password, password_wo_version)`, and `RequiredTogether(password_wo, password_wo_version)` — the last ensures the version trigger never floats free of the write-only value (which would otherwise silently fall through to the random-password branch).
- **Tests**:
  - `TestIntegrationSqlUserResource` creates users via `password`, `password_wo` + `password_wo_version = 1`, and no password, then rotates both managed paths (edit `password`, bump version to `2`) and asserts the corresponding `UpdateSQLUserPassword` calls fire.
  - `TestIntegrationSqlUserResource_PasswordWOVersionRequiresPasswordWO` asserts the `RequiredTogether` validator rejects a config that sets `password_wo_version` alone.
  - `TestIntegrationSqlUserResource_ClearLegacyPasswordDoesNotLoop` pins the migration scenario — creates with `password`, then clears it, and relies on the framework's post-apply non-empty-plan check to verify state was written (a regression would loop).
  - A runtime `skipIfTerraformBelowWriteOnly` check skips cleanly on Terraform CLIs below 1.11 before any gomock expectations are registered.
- **Example & docs**: example shows `password_wo` + `password_wo_version` with the legacy block as a commented-out reference. Generated resource doc regenerated via `go generate`.
- CHANGELOG.md is intentionally left out so the entry can be folded into the next release commit.

## Deprecation and migration

Terraform providers do not have REST-style API versioning. Deprecation flows through `DeprecationMessage` on the schema attribute (surfaces a warning in `plan`/`apply` whenever `password` is set) and removal waits for the next major provider release. Consumers should:

- For new resources, use `password_wo` + `password_wo_version`. Source the value from a sensitive variable or ephemeral resource so it never lands on disk.
- For existing resources that use `password`, plan a migration. The cleanest path is `terraform state rm cockroach_sql_user.<name>` → swap `password` for `password_wo` (and add `password_wo_version`) in the config → `terraform import` the resource back. The SQL user on the server is untouched; only state is rewritten. Accepting a recreate is the alternative but typically not acceptable for a live DB user.
- Rotating `password_wo` requires bumping `password_wo_version` in the same apply. Without the bump the change is invisible to Terraform — by design.

## Why remove `password` (eventually)

`password` persists the cleartext credential in `terraform.tfstate` for the lifetime of the resource. Any reader of state — local files, remote backends, CI artifacts — can recover it. The terraform-plugin-framework `WriteOnly` mechanism (added in Terraform 1.11) is the supported way to accept a secret on apply without ever writing it to state. This PR ships the new attribute now and deprecates the old one so consumers can plan the migration; removal will land in a future major release.

## Validation

- `go build ./...` and `go vet ./...` clean.
- `go generate` regenerated the resource doc.
- `TestIntegrationSqlUserResource`, `TestIntegrationSqlUserResource_PasswordWOVersionRequiresPasswordWO`, and `TestIntegrationSqlUserResource_ClearLegacyPasswordDoesNotLoop` all pass against Terraform 1.11.4. The clear-password regression test was verified to fail on the buggy early-return shape (reproducing "Provider produced inconsistent result after apply") before the fix landed.
- Skips cleanly on Terraform 1.5.7.
- Wider `TestIntegration*` sweep unchanged from main (the one pre-existing failure, `TestIntegrationDedicatedClusterResource`, reproduces on `main` without this change).


internal/provider/sql_user_resource.go
```diff
@@ -152,15 +159,14 @@ func (r *sqlUserResource) Create(
 
 	var sqlUserRequest client.CreateSQLUserRequest
 	sqlUserRequest.Name = sqlUserSpec.Name.ValueString()
-	if sqlUserSpec.Password.IsNull() {
+	if passwordWO.IsNull() {
 		sqlUserRequest.Password, err = generateRandomPassword()
-		sqlUserSpec.Password = types.StringNull()
 		if err != nil {
```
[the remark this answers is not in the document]
souravcrl: Good catch. Added an explicit `IsUnknown()` guard that returns a diagnostic error before `ValueString()` is called, so an unknown `password_wo` value can't silently become an empty password. See 42435ff.

internal/provider/sql_user_resource.go
```diff
@@ -68,9 +79,26 @@ func (r *sqlUserResource) Schema(
...
+			"password_wo": schema.StringAttribute{
+				Optional:  true,
+				Sensitive: true,
+				WriteOnly: true,
+				Description: "Write-only password for the SQL user (Terraform CLI 1.11+ required). " +
+					"The value is sent on create and on rotation but never stored in Terraform state. " +
+					"To rotate, change the value and bump `password_wo_version`; without a version bump, " +
+					"Terraform cannot detect changes to a write-only attribute. Mutually exclusive with `password`.",
+			},
+			"password_wo_version": schema.Int64Attribute{
```
abhishek-kedia, line 121: `password_wo_version` without `password_wo` is silently accepted. The ConfigValidators enforce password vs password_wo mutual exclusion, but there's no validation that password_wo_version requires password_wo. A user who writes:

```
  resource "cockroach_sql_user" "x" {
    name                = "bob"
    password_wo_version = 1
    cluster_id          = var.cluster_id
  }
```

gets no error, and the user gets a random password with no indication their config is wrong. We can add a RequiredTogether validator for password_wo and password_wo_version.

internal/provider/sql_user_resource.go
```diff
@@ -150,17 +199,31 @@ func (r *sqlUserResource) Create(
...
+		resp.Diagnostics.AddError(
+			"Unknown password_wo value",
+			"The password_wo attribute must be known at apply time. Ensure the value is not derived from an unknown expression.",
+		)
+		return
+	case !passwordWO.IsNull():
+		sqlUserRequest.Password = passwordWO.ValueString()
+	case !sqlUserSpec.Password.IsNull():
+		sqlUserRequest.Password = sqlUserSpec.Password.ValueString()
+	default:
```
abhishek-kedia, line 243: The old code had: 
```
sqlUserSpec.Password = types.StringNull()
```
in the random-password branch, now is removed. Is that intentional?

internal/provider/sql_user_resource.go
```diff
@@ -248,35 +311,43 @@ func (r *sqlUserResource) Read(
...
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	newPassword, shouldUpdate, diag := resolveUpdatedPassword(plan, state, passwordWO)
+	if diag != nil {
+		resp.Diagnostics.Append(diag)
+		return
+	}
```
abhishek-kedia: When a user clears the legacy password to null (a legitimate migration step toward password_wo), resolveUpdatedPassword returns a warning diagnostic. Because the check here is diag != nil followed by return, Update exits before resp.State.Set(ctx, plan) is reached. This means the null value is never persisted to state, so on the next terraform plan Terraform sees the stale password in state, diffs again, and triggers another Update, creating an nfinite plan/apply loop.

internal/provider/sql_user_resource.go
```diff
@@ -288,9 +359,49 @@ func (r *sqlUserResource) Update(
...
+//  2. `password_wo` + `password_wo_version`: the write-only value is absent
+//     from state, so the integer trigger is the only signal of rotation. A
+//     bump (planned version != state version, with `password_wo` set in
+//     config) triggers a push of the current config value.
+//
+// When `password` is cleared (planned null while state had a value) a warning
+// is returned: clearing the attribute does not change the underlying password,
+// it only removes the cleartext from state. This mirrors the historical
+// behavior of the resource so that consumers can stop tracking the legacy
+// attribute without first rotating the user.
```
abhishek-kedia: nit: comments need not be this long.

internal/provider/sql_user_resource.go
```diff
@@ -288,9 +359,49 @@ func (r *sqlUserResource) Update(
...
+			"Setting the password field to null will not change the password. It will simply remove it from Terraform state.",
+		)
+	}
+
+	// Write-only rotation: password_wo_version bumped while password_wo is
+	// set in config. A first-time set of password_wo on an existing resource
+	// (state version null, plan version non-null) is also treated as a bump.
+	versionChanged := !plan.PasswordWOVersion.Equal(state.PasswordWOVersion)
+	if versionChanged && !passwordWO.IsNull() && !passwordWO.IsUnknown() {
+		return passwordWO.ValueString(), true, nil
```
abhishek-kedia: The Create path explicitly errors when passwordWO.IsUnknown(), but here in the update path the same scenario, version bumped while password_wo is unknown — silently returns ("", false, nil). The rotation doesn't happen and the user gets no feedback.                                                                                                                                                                                                

This should return an error diagnostic to match the create behavior. Something like:                                                                                                                                                   
```                                                                                                                                                                                                                                      
if versionChanged && !passwordWO.IsNull() {                                                                                                                                                                                            
    if passwordWO.IsUnknown() {                                                                                 
        return "", false, diag.NewErrorDiagnostic(                                                                                                                                                                                     
            "Unknown password_wo value",                                                                                                                                                                                               
            "The password_wo attribute must be known at apply time. ...",                                                                                                                                                              
         )                                                                                                                                                                                                                              
      }                                                                                                                                                                                                                                
     return passwordWO.ValueString(), true, nil                                                                                                                                                                                         
}
```

internal/provider/sql_user_resource.go
```diff
@@ -68,9 +79,26 @@ func (r *sqlUserResource) Schema(
...
 			"password": schema.StringAttribute{
-				Optional:    true,
-				Sensitive:   true,
-				Description: "If provided, this field sets the password of the SQL user when created. If omitted, a random password is generated, but not saved to Terraform state. The password must be changed via the CockroachDB cloud console.",
+				Optional:           true,
+				Sensitive:          true,
+				DeprecationMessage: passwordDeprecationMessage,
+				Description: "Deprecated. If provided, this field sets the password of the SQL user when created. " +
+					"The value is persisted in Terraform state, which is the reason for deprecation; " +
+					"prefer `password_wo`. If omitted, a random password is generated and not saved to state.",
```
abhishek-kedia: "The password must be changed via the CockroachDB cloud console." was removed from the original comment.

internal/provider/sql_user_resource_test.go
```diff
@@ -22,36 +22,84 @@ import (
...
+// cockroach_sql_user).
+var minTerraformVersionForWriteOnly = version.Must(version.NewVersion("1.11.0"))
+
+var terraformVersionRegex = regexp.MustCompile(`Terraform v(\d+\.\d+\.\d+(?:-[\w.]+)?)`)
+
+// skipIfTerraformBelowWriteOnly skips the test when the local Terraform CLI
+// (the same binary the plugin testing framework will invoke) predates
+// write-only attribute support. We perform this check before any test
+// scaffolding (e.g. gomock expectations) so the skip is clean.
+func skipIfTerraformBelowWriteOnly(t *testing.T) {
```
abhishek-kedia, line 53: The Terraform version is checked twice: once via skipIfTerraformBelowWriteOnly(t) at the top of each test function, and again via tfversion.SkipBelow inside the TestCase.

If only tfversion.SkipBelow is enough, we can consider removing this function.

internal/provider/sql_user_resource_test.go
```diff
@@ -127,23 +191,45 @@ func testSqlUserResource(
...
+				// Step 2 exercises both rotation paths simultaneously.
+				Config: getTestSqlUserResourceConfig(
+					clusterName, sqlUserNamePass, sqlUserNameWO, sqlUserNameNoPass,
+					"cockroach@legacy-2", "cockroach@wo-2", 2,
+				),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceNameWO, "password_wo_version", "2"),
 				),
 			},
 			{
```
abhishek-kedia, line 250: Import state test no longer covers the legacy password path. We should still have the tests until we stop supporting it.

internal/provider/sql_user_resource_test.go
```diff
@@ -127,23 +191,45 @@ func testSqlUserResource(
...
+					resource.TestCheckResourceAttr(resourceNameWO, "password_wo_version", "1"),
+				),
+			},
+			{
+				// Step 2 exercises both rotation paths simultaneously.
+				Config: getTestSqlUserResourceConfig(
+					clusterName, sqlUserNamePass, sqlUserNameWO, sqlUserNameNoPass,
+					"cockroach@legacy-2", "cockroach@wo-2", 2,
+				),
+				Check: resource.ComposeTestCheckFunc(
```
abhishek-kedia, line 233: Step 2 claims to exercise both rotation paths, but the Check only verifies the write-only path (password_wo_version bumped to 2). There's no assertion on resourceNamePass to confirm the legacy password rotation was applied

examples/resources/cockroach_sql_user/resource.tf
```diff
@@ -2,17 +2,34 @@ variable "cluster_id" {
...
+# The legacy `password` attribute remains available for backwards
+# compatibility but is deprecated: it persists the cleartext password in
+# `terraform.tfstate`. New consumers should use `password_wo` above; existing
+# consumers should plan a migration.
+#
+# resource "cockroach_sql_user" "legacy" {
+#   name       = "legacy-sql-user"
+#   password   = var.sql_user_password
+#   cluster_id = var.cluster_id
+# }
```
abhishek-kedia, line 35: Since we are deprecating legacy passwords, should we consider removing it?

abhishek-kedia (commented):
Please ensure PR description follows same format as other PRs in this repo. Also update the CHANGELOG file.

examples/resources/cockroach_sql_user/resource.tf
souravcrl: Not yet — this PR only deprecates `password` (soft `DeprecationMessage` warning at plan/apply); the attribute still works end-to-end and existing consumers can keep using it through the deprecation window. Removal is planned for the next major release. The commented-out block is intentional — it gives existing consumers a side-by-side reference for what the legacy syntax looked like so they can recognize their own configs while migrating. Happy to drop it once we cut the major release that removes the attribute.

internal/provider/sql_user_resource.go
souravcrl: I did not understand why we had it her in this file in the first place? Can you clarify a bit on that? I can re add it if required.

internal/provider/sql_user_resource.go
souravcrl: Good catch — added `resourcevalidator.RequiredTogether(password_wo, password_wo_version)` plus a dedicated test (`TestIntegrationSqlUserResource_PasswordWOVersionRequiresPasswordWO`) that asserts the validator fires on the exact config you posted.

I also considered a custom asymmetric validator that would only error in the "version without wo" direction — that matches your concern most precisely and would let consumers set `password_wo` once without the bookkeeping of an unused `password_wo_version`. I went with the symmetric `RequiredTogether` instead because (a) it's the canonical write-only + trigger idiom in the ecosystem — `random_password_wo`, AWS RDS, and HashiCorp's own examples all pair the two attributes together with `RequiredTogether`; (b) consumers who initially set `password_wo` alone tend to discover the trigger only when their first rotation silently no-ops, which is a worse failure mode than being prompted to add `password_wo_version = 1` up front; and (c) it's one line of standard framework plumbing rather than a custom resource ConfigValidator that we'd have to maintain. The mild cost — set the integer to `1` on first use even if you don't intend to rotate — feels worth the stronger contract.

internal/provider/sql_user_resource.go
souravcrl: I think this was dead code as we check this condition `if sqlUserSpec.Password.IsNull() {` before the assignment

internal/provider/sql_user_resource.go
souravcrl: Good catch — fixed. Pinning regression with `TestIntegrationSqlUserResource_ClearLegacyPasswordDoesNotLoop`, which I confirmed fails on the buggy early-return shape (reproducing "Provider produced inconsistent result after apply") before the fix.

I went with refactoring `resolveUpdatedPassword` to return `diag.Diagnostics` rather than the simpler "just delete the `return`" because the underlying bug is a *type-shape* problem, not a single missing line: returning `diag.Diagnostic` (singular) collapses warning and error into the same out-channel, and any caller — present or future — that reaches for the natural `if d != nil { ...; return }` idiom will land on the same loop. Returning `diag.Diagnostics` (the slice the framework's `Append` already accepts) makes the caller use the framework's own `HasError` predicate to decide whether to stop, which is the contract every other handler in this provider follows. It costs three lines vs. one, but it removes the foot-gun rather than just stepping around it, and the helper now composes the same way the rest of the codebase does.

internal/provider/sql_user_resource.go
souravcrl: fixed

internal/provider/sql_user_resource_test.go
souravcrl: Verified empirically — not dead code. The two checks fire at different lifecycle points and only the early one prevents a false failure on TF < 1.11. Concretely: `gomock.NewController(t)` registers a `t.Cleanup` (in vendor/github.com/golang/mock/gomock/controller.go) that calls `ctrl.finish(true, nil)`, which walks every unmet expectation and calls `t.Errorf("missing call(s) to ...")`. So if we relied only on `tfversion.SkipBelow` inside the TestCase, the sequence on TF 1.5.7 would be: (1) `gomock.NewController(t)` runs and registers the Finish cleanup, (2) ~15 `s.EXPECT()...` calls register expectations, (3) `resource.Test` runs and `tfversion.SkipBelow` calls `t.Skip` so no steps execute, (4) test cleanup runs `ctrl.finish` which calls `Errorf` for every unmet expectation — flipping the test from skipped to failed. I confirmed this by temporarily removing `skipIfTerraformBelowWriteOnly` and running on TF 1.5.7: the test failed with 11 "missing call(s) to ..." errors despite the SkipBelow message appearing in the output. `skipIfTerraformBelowWriteOnly` runs at the top of the test function — before the gomock controller is even constructed — so the cleanup never sees unmet expectations and the skip is clean. For `TestAccSqlUserResource` (no mocks) the early call is technically redundant, but I kept it for consistency so every write-only-dependent test follows the same single-line pattern at the top.

internal/provider/sql_user_resource.go
souravcrl: fixed. `resolveUpdatedPassword` now errors in exactly the same shape as Create when a rotation is requested (version bumped, `password_wo` set in config) but the value is unknown at apply time. To keep the two diagnostics from drifting, I extracted `unknownPasswordWODiagSummary` and `unknownPasswordWODiagDetail` constants and reused them on both paths, so a consumer who hits the problem on a rotation gets the same title and guidance as one who hits it on a fresh create. Coverage is pinned by a new table-driven `TestResolveUpdatedPassword` that exercises all eight decision arms of the helper (no-op, legacy edit, legacy no-op, legacy cleared → warning, version-bump rotation, first-time set rotation, version-unchanged no-op, and the new unknown-`password_wo` → error). The only behavior change from your suggested snippet is that I kept the unknown check inside the `versionChanged && !passwordWO.IsNull()` arm — if version didn't change we're not rotating at all, so an unknown `password_wo` floating around in config (e.g., from a computed expression that resolves elsewhere) shouldn't surface here.

internal/provider/sql_user_resource_test.go
souravcrl: Added. Step 1 now asserts `password = "cockroach@legacy-1"` on `cockroach_sql_user.with_pass` (confirming Create wrote the supplied value to state) and step 2 asserts `password = "cockroach@legacy-2"` (confirming the rotation landed in state). The gomock `UpdateSQLUserPassword` expectations registered above already match on the exact `Password` field, so a wrong-value Update would have failed gomock; the new TestCheckResourceAttr lines are the state-side counterpart so a future regression that drops `resp.State.Set` or writes a stale value would be caught even if the API call still fired correctly. (`password_wo` is write-only and absent from state, so the WO path stays observable only via the `password_wo_version` integer.)

internal/provider/sql_user_resource_test.go
souravcrl: fixed

internal/provider/sql_user_resource.go
abhishek-kedia: Mainly because of password being visible in .tfstate. It will be good to add it back.

More context: https://github.com/cockroachdb/terraform-provider-cockroach/pull/77

internal/provider/sql_user_resource.go
```diff
@@ -83,6 +121,37 @@ func (r *sqlUserResource) Schema(
...
+//   - `password` and `password_wo_version` are likewise mutually exclusive —
+//     the version trigger has no meaning on the legacy path.
+//   - `password_wo` and `password_wo_version` must be configured together.
+//     The asymmetric "if version then wo" reading would also be sound (and
+//     marginally less ceremonial for set-once consumers), but we use the
+//     symmetric `RequiredTogether` here because it matches the canonical
+//     write-only + trigger idiom in the ecosystem (`random_password_wo`,
+//     AWS RDS, etc.) and forecloses the "I set `password_wo` once and now
+//     can't rotate without a state surgery" gotcha — every consumer
+//     starts with the trigger in place.
```
abhishek-kedia: nit: would be good to have straight forward comments.

internal/provider/sql_user_resource.go
abhishek-kedia: It wasn't a dead code.

```
if sqlUserSpec.Password.IsNull() {
 --- Generate random password ---
		sqlUserRequest.Password, err = generateRandomPassword()
		sqlUserSpec.Password = types.StringNull()
		if err != nil {
			resp.Diagnostics.AddError("Error generating password", err.Error())
			return
		}
	} else {
--- assign the user given password ---
		sqlUserRequest.Password = sqlUserSpec.Password.ValueString()
	}
```

abhishek-kedia: Also, what happens if passwordWo or legacy password is ""?

abhishek-kedia (commented):
PR description and CHANGELOG still needs updation.

internal/provider/sql_user_resource.go
souravcrl: Good catch — empty string is a known non-null value so it slipped past the `IsNull()/IsUnknown()` guards and would've been sent as-is to the
  API; added `stringvalidator.LengthAtLeast(1)` to both password and password_wo so it now fails at plan time.

internal/provider/sql_user_resource.go
souravcrl: I understand now and I have added it back.

internal/provider/sql_user_resource.go
souravcrl: done

abhishek-kedia:
Raised another PR with the fixes applied: https://github.com/cockroachdb/terraform-provider-cockroach/pull/353

PR #352, opened 2026-07-07
provider: allow setting per-region vCPU/machine type

mwstobo:
This commit adds the ability to set either num_virtual_cpus or machine_type on individual regions in a cockroach_cluster resource. This feature is not yet fully public, and requires specific access enabled in your Cockroach Cloud organization.

Note, this feature is mutually exclusive with setting dedicated.num_virutal_cpus or dedicated.machine_type. Using both will result in an error.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `num_virtual_cpus` and `machine_type` attributes to the `regions` block on
+  `cockroach_cluster`, enabling heterogeneous Advanced clusters whose regions use
+  different machine types. These are mutually exclusive with the cluster-wide
+  `dedicated.num_virtual_cpus`/`dedicated.machine_type` and with each other within a
+  region, and must be set on every region when used. Requires a feature flag to be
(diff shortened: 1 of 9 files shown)


mwstobo:
Acceptance test run: https://github.com/cockroachdb/terraform-provider-cockroach/actions/runs/28898296144

fantapop (approved):
This all looks well thought out to me.  The one thing I was wondering about is the drift detection.  I know we were talking about automatically updating the machine type for customers at some point and how this could cause a conflict in their terraform configurations if they were setting it.  Did we move forward with this same solution elsewhere? i.e. Let it break and force them to update their configs? or is that a new concept being added here?

mwstobo:
Yeah, I was thinking about that too. This commit uses the same functionality we already have for cluster-wide machine types. If the machine type changes while preserving the vCPUs (which is always what happens), users will get a warning, but no error. @sagar-pai-crl added this functionality originally, and it's preserved here for per-region machine types as well.

mwstobo:
See: https://github.com/cockroachdb/terraform-provider-cockroach/commit/0cd3172c74b6737642f30b1150525dd590a8e5f1

PR #353, opened 2026-07-15
[CC-36355]: add write-only password attributes to cockroach_sql_user resource

abhishek-kedia:
- Introduced `password_wo` and `password_wo_version` attributes for the `cockroach_sql_user` resource, allowing for secure password management without persisting sensitive data in Terraform state.
- Deprecated the existing `password` attribute, which will be removed in a future major release. Updated documentation to reflect these changes and provide migration guidance.
- Updated tests to validate the new functionality and ensure compatibility with Terraform CLI 1.11+.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)


internal/provider/sql_user_resource.go
```diff
@@ -68,9 +87,29 @@ func (r *sqlUserResource) Schema(
...
+				Description: "Deprecated. If provided, this field sets the password of the SQL user when created. " +
+					"The value is persisted in Terraform state, which is the reason for deprecation; " +
+					"prefer `password_wo`. If omitted, a random password is generated and not saved to state. " +
+					"The password must be changed via the CockroachDB cloud console.",
+			},
+			"password_wo": schema.StringAttribute{
+				Optional:   true,
+				Sensitive:  true,
+				WriteOnly:  true,
+				Validators: []validator.String{stringvalidator.LengthBetween(12, 500)},
```
sanchit-CRL, line 103: Existing configs whose password falls outside 12–500 will now fail at terraform plan after upgrading, rather than at apply/server-side.

sanchit-CRL (approved):

internal/provider/sql_user_resource.go
abhishek-kedia: I think we should keep it. This restriction is already there in our servers, it will prevent users from having a bad config. For existing users which don't meet the criteria, it will fail on the server and it won't affect those who already met it.

PR #358, opened 2026-08-06
[CNSL-2431] reduce cockroach_cluster plan noise on updates

devlaneyr:
Changing one attribute on an existing cluster reported roughly ten unrelated attributes as "known after apply", burying the real change. The framework marks every Computed attribute with a null config value unknown on each plan, and UseStateForUnknown had been applied to only some of the cluster schema's Computed attributes.

Pin the ones that are stable across an update: account_id, parent_id, delete_protection, dedicated.memory_gib, dedicated.disk_iops, and the regions block's ui_dns, private_endpoint_dns and s3_vpc_endpoint_id.

The two dedicated fields are derived server-side, so each is paired with a reset to unknown in coordinateDedicatedMachinePlan. memory_gib depends on the machine type alone and reuses the existing `resizing` flag; disk_iops depends on the machine type and the storage size, so it gets its own condition. Folding storage into `resizing` would unsettle memory_gib on a pure storage change, reintroducing the noise.

Vendors float64planmodifier; go.mod and go.sum are unchanged.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bumped version of cockroach-cloud-sdk-go from v8 to v9.
 
+### Fixed
+
+- Reduced `cockroach_cluster` plan noise: `account_id`, `parent_id`,
+  `delete_protection`, `dedicated.memory_gib`, `dedicated.disk_iops`, and the
+  `regions` block's `ui_dns`, `private_endpoint_dns`, and `s3_vpc_endpoint_id` are
+  no longer reported as "known after apply" on updates that don't change them.
+
 ## [1.22.0] - 2026-07-16
 
 ### Added
--- a/internal/provider/cluster_resource.go
+++ b/internal/provider/cluster_resource.go
@@ -38,6 +38,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/boolplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/float64planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/listplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/objectplanmodifier"
@@ -89,6 +90,9 @@ var regionSchema = schema.NestedAttributeObject{
 		"ui_dns": schema.StringAttribute{
 			Computed:    true,
 			Description: "DNS name used when connecting to the DB Console for the cluster.",
+			PlanModifiers: []planmodifier.String{
+				stringplanmodifier.UseStateForUnknown(),
+			},
 		},
 		"internal_dns": schema.StringAttribute{
 			Computed:    true,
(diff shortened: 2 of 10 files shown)


devlaneyr:
I reproduced by using the main.tf file: 

```

terraform {
    required_providers {
      cockroach = { source = "cockroachdb/cockroach" }
    }
  }

  variable "environment" {
    type    = string
    default = "dev"
  }

  resource "cockroach_cluster" "repro" {
    name           = "cnsl-2431"
    cloud_provider = "GCP"
    plan           = "STANDARD"
    serverless     = { usage_limits = { provisioned_virtual_cpus = 2 } }
    regions        = [{ name = "us-central1" }]
    labels         = { repro = "one" }
  }

```

which I performed `terraform apply` on. 

I then modified the labels to: 

` labels         = { repro = "two" }`

which yielded noisy output from `terraform plan` as follows (I've redacted potentially sensitive fields): 

```
 # cockroach_cluster.repro will be updated in-place
  ~ resource "cockroach_cluster" "repro" {
      ~ account_id        = "redacted" -> (known after apply)
      ~ delete_protection = false -> (known after apply)
      ~ full_version      = "redacted" -> (known after apply)
        id                = "redacted"
      ~ labels            = {
          ~ "repro" = "one" -> "two"
        }
        name              = "cnsl-2431"
      ~ operation_status  = "UNSPECIFIED" -> (known after apply)
      ~ parent_id         = "root" -> (known after apply)
      ~ regions           = [
          ~ {
              + machine_type         = (known after apply)
                name                 = "us-central1"
              + num_virtual_cpus     = (known after apply)
              ~ primary              = true -> (known after apply)
              ~ private_endpoint_dns = "redacted" -> (known after apply)
              + s3_vpc_endpoint_id   = (known after apply)
              + ui_dns               = (known after apply)
                # (3 unchanged attributes hidden)
            },
        ]
      ~ serverless        = {
          + with_empty_ip_allowlist = (known after apply)
            # (3 unchanged attributes hidden)
        }
      ~ state             = "CREATED" -> (known after apply)
      ~ upgrade_status    = "FINALIZED" -> (known after apply)
        # (5 unchanged attributes hidden)
    }

```
After the updates in this commit the same `terraform apply` yields:

```
# cockroach_cluster.repro will be updated in-place
  ~ resource "cockroach_cluster" "repro" {
      ~ full_version      = "redacted" -> (known after apply)
        id                = "redacted"
      ~ labels            = {
          ~ "repro" = "one" -> "two"
        }
        name              = "cnsl-2431"
      ~ operation_status  = "UNSPECIFIED" -> (known after apply)
      ~ regions           = [
          ~ {
              + machine_type         = (known after apply)
                name                 = "us-central1"
              + num_virtual_cpus     = (known after apply)
              ~ primary              = true -> (known after apply)
                # (6 unchanged attributes hidden)
            },
        ]
      ~ serverless        = {
          + with_empty_ip_allowlist = (known after apply)
            # (3 unchanged attributes hidden)
        }
      ~ state             = "CREATED" -> (known after apply)
      ~ upgrade_status    = "FINALIZED" -> (known after apply)
    }

```



A slight improvement but probably not the reduction we want to see. Fixing these one attribute at a time only goes so 
far, since the rest of them do actually change on some updates, so we can't just pin them. In a follow up PR I could 
expand on this solution by looking at the plan as a whole. If the only thing changing is metadata that can't affect 
the cluster then we can show the plan as settled so a label change would then show nothing as (known after apply).

devlaneyr:
Also the checks are failing because github actions is having an incident it seems https://www.githubstatus.com/

fantapop (approved):
This seems reasonable to me.  For the primary value, maybe we could eventually deprecate the bool and instead store the value as a region name.

PR #361, opened 2026-08-10
Raise the unit test timeout from 5 to 10 minutes

devlaneyr:
Recent runs have been bumping up against the 5 minute limit and timing out.

**Commit checklist**
- [ ] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ clean:
 	go clean -testcache -i -x
 
 test:
-	go test ./... -v $(TESTARGS) -timeout 5m
+	go test ./... -v $(TESTARGS) -timeout 10m
 
 testacc:
 	TF_ACC=1 go test ./... -v $(TESTARGS) -timeout 120m -parallel 100


fantapop (approved):
This seems reasonable to me.  I can see that previously succeeding runs were talking almost 5 minutes.   btw, I had claude check the test log for the longest running tests.  

```
The big one

TestIntegrationServerlessClusterResource (68.7s) is by far the dominant cost — 26% of the total runtime. It’s a table-driven test with ~36 subtests. Its own slowest subtests:

13 subtests at ~3.3s each (the single-region_serverless_BASIC_cluster_* conversion cases and similar) — these alone total ~43s
Another cluster at ~1.5–1.6s each (the setting_cockroach_version_* cases)
The rest are fast (0.5s validation-only cases)
TestIntegrationAllowlistEntryResource (21s) is second, split as dedicated (12.0s) + serverless (9.1s).
```

It looks like we could optimize that a bit but probably not worth it for now.  increasing the timeout sounds good.

PR #362, opened 2026-08-11
[CNSL-2431] remove plan noise from metadata-only cluster updates

devlaneyr:
Terraform marks every Computed attribute the config leaves unset as "known after apply", so changing a single label on a cockroach_cluster planned a dozen unrelated attributes as pending changes.

A prior commit e66893f added UseStateForUnknown to some of them. That modifier copies prior state into the plan unconditionally, so it only fits attributes that never change on an update. The rest do change on some updates, and stayed noisy on all of them.

Whether those attributes can change depends on what else the update is doing, which a per-attribute modifier can't see. This adds a resource-level check over the whole plan: when every difference between plan and state is confined to metadata that doesn't start a cluster operation, the remaining unknowns are filled in from prior state. Otherwise the plan is left alone. Only unknowns are ever replaced, so a plan can get quieter but never hide a change.

**Commit checklist**
- [ ] Changelog
- [ ] Integration test(s)
- [ ] Acceptance test(s)
- [ ] Example(s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `delete_protection`, `dedicated.memory_gib`, `dedicated.disk_iops`, and the
   `regions` block's `ui_dns`, `private_endpoint_dns`, and `s3_vpc_endpoint_id` are
   no longer reported as "known after apply" on updates that don't change them.
+- A `cockroach_cluster` update that changes only `labels`, `delete_protection`,
+  `parent_id`, or `backup_config` now plans with nothing marked "known after
+  apply". Fields like `full_version`, `state`, and the `regions` block's
+  `primary` used to show up on every update. Clusters with an operation in
+  flight are unchanged.
 
 ## [1.22.0] - 2026-07-16
 
--- a/internal/provider/cluster_resource.go
+++ b/internal/provider/cluster_resource.go
@@ -47,6 +47,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 )
@@ -1081,7 +1082,13 @@ func (r *clusterResource) ModifyPlan(
 		// leave stale-but-known values that a resize/add/remove would invalidate.
 		if coordinateDedicatedMachinePlan(config, plan, state) {
 			resp.Diagnostics.Append(resp.Plan.Set(ctx, plan)...)
+			if resp.Diagnostics.HasError() {
+				return
+			}
 		}
+
+		// Last, because it reads the plan the steps above have finished writing.
+		resolveUnknownsForMetadataOnlyUpdate(ctx, state, req, resp)
 	}
 
 	if req.Plan.Raw.IsNull() {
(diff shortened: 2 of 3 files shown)


devlaneyr:
I reproduced by using the main.tf file:

```
terraform {
    required_providers {
      cockroach = { source = "cockroachdb/cockroach" }
    }
  }

  variable "environment" {
    type    = string
    default = "dev"
  }

  resource "cockroach_cluster" "repro" {
    name           = "cnsl-2431"
    cloud_provider = "GCP"
    plan           = "STANDARD"
    serverless     = { usage_limits = { provisioned_virtual_cpus = 2 } }
    regions        = [{ name = "us-central1" }]
    labels         = { repro = "one" }
  }
```
which I performed terraform apply on.

I then modified the labels to:

 `labels         = { repro = "two" }`

which on main still yields noisy output from terraform plan as follows (I've redacted potentially sensitive fields):

```
 # cockroach_cluster.repro will be updated in-place
  ~ resource "cockroach_cluster" "repro" {
      ~ full_version      = "redacted" -> (known after apply)
        id                = "redacted"
      ~ labels            = {
          ~ "repro" = "one" -> "two"
        }
        name              = "cnsl-2431"
      ~ operation_status  = "UNSPECIFIED" -> (known after apply)
      ~ regions           = [
          ~ {
              + machine_type         = (known after apply)
                name                 = "us-central1"
              + num_virtual_cpus     = (known after apply)
              ~ primary              = true -> (known after apply)
                # (6 unchanged attributes hidden)
            },
        ]
      ~ serverless        = {
          + with_empty_ip_allowlist = (known after apply)
            # (3 unchanged attributes hidden)
        }
      ~ state             = "CREATED" -> (known after apply)
      ~ upgrade_status    = "FINALIZED" -> (known after apply)
    }
```
After the updates in this commit the same terraform plan yields:

```
 # cockroach_cluster.repro will be updated in-place
  ~ resource "cockroach_cluster" "repro" {
        id                = "redacted"
      ~ labels            = {
          ~ "repro" = "one" -> "two"
        }
        name              = "cnsl-2431"
        # (14 unchanged attributes hidden)
    }
 ```

Nothing is marked known after apply anymore, so the plan shows only the label change.

**Comment by devlaneyr** [comment]
@fantapop This is a follow-up PR to (https://github.com/cockroachdb/terraform-provider-cockroach/pull/358) 

Can you confirm that the fields in `metadataOnlyAttributes` truly are metadata only?

File: internal/provider/cluster_resource.go [hunk 1]
```diff
@@ -1926,6 +1933,136 @@ func coordinateDedicatedMachinePlan(config, plan, state *CockroachCluster) bool
 	return changed
 }
 
+// metadataOnlyAttributes are the attributes an update can change without
+// causing other attributes to change.
+var metadataOnlyAttributes = map[string]struct{}{
+	"labels":            {},
+	"delete_protection": {},
+	"parent_id":         {},
+	"backup_config":     {},
```
fantapop, line 1942: There are some weird rules governing backup_config.  It can only be updated once without a support ticket.  See here: https://github.com/cockroachdb/terraform-provider-cockroach/blob/3440b333e20a638ac142fe5a2dc1ee73040c2229/docs/guides/updating-backup-retention.md

I'm wondering if this might cause a problem if you're doing something like using state for unknown and expecting the update to work.  It's worth testing at least.

internal/provider/cluster_resource.go
```diff
@@ -1926,6 +1933,136 @@ func coordinateDedicatedMachinePlan(config, plan, state *CockroachCluster) bool
...
+	}
+
+	// Diff below requires both sides to have the same type.
+	if planRaw.Type() == nil || stateRaw.Type() == nil || !planRaw.Type().Is(stateRaw.Type()) {
+		return
+	}
+
+	// Only resolve unknowns for a cluster in the CREATED state. Anything else
+	// isn't settled, and the apply would return different values than the plan
+	// promised, failing with "inconsistent result after apply".
```
fantapop, line 1976: I'm slightly surprised by this.   Here are the states:

```
message ClusterState {
  enum Type {
    UNSPECIFIED = 0;
    CREATING = 1;
    CREATED = 2;
    CREATION_FAILED = 3;
    DELETED = 4;
    // An exclusive operation is being performed on this cluster.
    // Other operations should not proceed if they did not set a cluster into the LOCKED state.
    LOCKED = 5;
  }
```
It seems plausible we wouldn't allow updates while in CREATING, CREATION_FAILED or DELETED.  I'm unclear about LOCKED.  I think we might still allow certain metadata updates in which case this would be fine.  This might be worth digging into a bit more.

internal/provider/cluster_resource.go
```diff
@@ -1926,6 +1933,136 @@ func coordinateDedicatedMachinePlan(config, plan, state *CockroachCluster) bool
...
+	if state.State.ValueString() != string(client.CLUSTERSTATETYPE_CREATED) {
+		return
+	}
+
+	if opStatus := state.OperationStatus.ValueString(); IsKnown(state.OperationStatus) &&
+		opStatus != "" && opStatus != string(client.CLUSTERSTATUSTYPE_UNSPECIFIED) {
+		return
+	}
```
fantapop, line 1984: Here again, it seems like we're not allowing updates during long running operations but I think that might actually be fine?   Although sometimes terraform might be polling for a long running operation to finish such as during an update.

internal/provider/cluster_resource.go
```diff
@@ -1926,6 +1933,136 @@ func coordinateDedicatedMachinePlan(config, plan, state *CockroachCluster) bool
...
+
+	diffs, err := planRaw.Diff(stateRaw)
+	if err != nil {
+		// Can't prove the update is inert, so leave the plan alone.
+		tflog.Trace(ctx, "skipping metadata-only unknown resolution", map[string]any{"error": err.Error()})
```
fantapop, line 1989: I agree with the Trace level logging here but it's probably worth a comment about why we would log this error at trace rather than error.

internal/provider/cluster_resource.go
```diff
@@ -1926,6 +1933,136 @@ func coordinateDedicatedMachinePlan(config, plan, state *CockroachCluster) bool
...
+		// did set is a real pending change and has to stay unknown.
+		if diff.Value1 != nil && !diff.Value1.IsKnown() && configIsNullAt(configRaw, diff.Path) {
+			continue
+		}
+		if !isMetadataOnlyPath(diff.Path) {
+			return
+		}
+	}
+
+	resolved, err := tftypes.Transform(planRaw, func(
```
fantapop, line 2006: It's probably worth a comment above here about what this is doing.  It's pretty incoherent (at least to me)

internal/provider/cluster_resource.go
```diff
@@ -1926,6 +1933,136 @@ func coordinateDedicatedMachinePlan(config, plan, state *CockroachCluster) bool
...
+		attrPath *tftypes.AttributePath, val tftypes.Value,
+	) (tftypes.Value, error) {
+		if val.IsKnown() {
+			return val, nil
+		}
+		// A config-driven unknown is the user's edit.
+		if !configIsNullAt(configRaw, attrPath) {
+			return val, nil
+		}
+		stateValI, _, err := tftypes.WalkAttributePath(stateRaw, attrPath)
```
fantapop, line 2016: what does this var name mean?  Why the "I" at the end?  Normally i would expect that to be short for "index" but I see we're casting it to the state value below.

internal/provider/cluster_resource.go
fantapop: It also might be worth trying to tuck some of this logic that's not specific to the cluster_resource into a separate file or package to keep it out of this file.  the cluster_resource file is already large.

CHANGELOG.md
```diff
@@ -23,6 +23,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `delete_protection`, `dedicated.memory_gib`, `dedicated.disk_iops`, and the
   `regions` block's `ui_dns`, `private_endpoint_dns`, and `s3_vpc_endpoint_id` are
   no longer reported as "known after apply" on updates that don't change them.
+- A `cockroach_cluster` update that changes only `labels`, `delete_protection`,
```
fantapop, line 26: Changelog entries should be placed on top like a stack.

PR #363, opened 2026-08-31
[CC-37027]: Fix cockroach_cmek Read panic on region-count drift

abhishek-kedia:
loadCMEKToTerraformState matched the API's region list to the plan/state region list by slice index, which panicked (index out of range) whenever the API returned more regions than Terraform tracked — e.g. after a region was added out of band. Because the crash is on the Read path, every subsequent plan/apply/destroy also crashed, bricking the workspace.

Match API regions to plan regions by name instead, nil-check the key URI before dereferencing, and sort plan-absent regions to the end so drift surfaces as a normal diff. Adds regression unit tests.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)


internal/provider/cmek_resource.go
```diff
@@ -428,42 +428,63 @@ func (r *cmekResource) ImportState(
...
+	sort.SliceStable(*cmekObj.RegionInfos, func(i, j int) bool {
+		return ordinalFor(cmekObj.GetRegionInfos()[i].GetRegion()) < ordinalFor(cmekObj.GetRegionInfos()[j].GetRegion())
 	})
 }
 
 func loadCMEKToTerraformState(
 	cmekObj *client.CMEKClusterInfo, state *ClusterCMEK, plan *ClusterCMEK,
 ) {
 	sortCMEKRegionsByPlan(cmekObj, plan)
+	plannedURIs := map[string]string{}
```
sanchit-CRL, line 458: var plannedURIs map[string]string (nil maps are safe to read) with allocation only inside the if plan != nil block avoids the throwaway allocation and reads more clearly.

internal/provider/cmek_resource.go
```diff
@@ -428,42 +428,63 @@ func (r *cmekResource) ImportState(
...
 	for i, region := range plan.Regions {
 		regionOrdinals[region.Region.ValueString()] = i
 	}
-	sort.Slice(*cmekObj.RegionInfos, func(i, j int) bool {
-		return regionOrdinals[cmekObj.GetRegionInfos()[i].GetRegion()] < regionOrdinals[cmekObj.GetRegionInfos()[j].GetRegion()]
+	ordinalFor := func(name string) int {
+		if o, ok := regionOrdinals[name]; ok {
+			return o
+		}
+		return len(plan.Regions)
```
sanchit-CRL, line 447: On the Read path (plan == state == current), this continue drops any CMEK region the server has but config doesn't declare, e.g. a region enabled out of band. The region silently disappears from state and never surfaces as drift, so an operator gets no signal that CMEK is active on a region they didn't declare.

This can be handled by raising an error: "region X has CMEK enabled but is not in regions; CMEK cannot be removed, add it to your config." This fails closed — apply is blocked until config matches reality, and once the region is added, Update is a no-op for it (it's already enabled server-side). This keeps drift visible and actionable instead of hidden.

This also contradicts the comment "placed at the end so drift surfaces as a normal diff", they're dropped, not surfaced.

abhishek-kedia: Fixed.

sanchit-CRL (approved):
LGTM

PR #365, opened 2026-09-08
[CC-37028]: Add configurable timeouts

abhishek-kedia:
- Introduced optional `timeouts` blocks for both `cockroach_cluster` and `cockroach_cmek` resources, allowing users to specify custom durations for `create` and `update` operations.
- Updated validation to reject timeout values shorter than the default settings (1h for create and 2h for update in `cockroach_cluster`; 2h for both in `cockroach_cmek`).
- Enhanced documentation to include details about the new `timeouts` block and its usage.
- Added tests to ensure the correct handling of timeout configurations and validation logic.
- Updated CHANGELOG to reflect these changes and their implications for users.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [ ] Example(s)


internal/provider/cluster_resource.go
```diff
@@ -1145,14 +1168,33 @@ func (r *clusterResource) Update(
...
+	// When the practitioner explicitly sets timeouts.update, treat it as an
+	// end-to-end budget for the whole apply so they can size it to their own
+	// CI/CD wall-clock limits. When it's omitted, keep the historical per-phase
+	// behavior: an update can run several waits in sequence (lock wait, a version
+	// upgrade, then a resize), and each has always received the full default
+	// budget. Collapsing the default would silently shrink it for combined
+	// applies on large clusters, so we only do so on explicit opt-in.
+	if timeoutAttrConfigured(plan.Timeouts, "update") {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, updateTimeout)
```
fantapop: It seems error prone to check if the value is configured by string name and then pick the value from a local var.  Can we have the value to use come back from the configured check?

internal/provider/cluster_resource.go
```diff
@@ -1145,14 +1168,33 @@ func (r *clusterResource) Update(
 		return
 	}
 
+	updateTimeout, diagsTimeout := plan.Timeouts.Update(ctx, clusterUpdateTimeout)
```
fantapop: reusing the same variable for the total operation timeout vs a single phase timeout makes this a bit confusing to follow.  Can you try splitting these out into more clearly labeled variable names?

internal/provider/cmek_resource.go
```diff
@@ -229,8 +241,29 @@ func (r *cmekResource) Create(
...
+	createTimeout, diagsTimeout := plan.Timeouts.Create(ctx, clusterUpdateTimeout)
+	resp.Diagnostics.Append(diagsTimeout...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Only treat an explicitly-set timeouts.create as an end-to-end budget for
+	// the whole enable + readiness sequence. When omitted, keep the historical
+	// per-phase behavior so existing configs aren't silently given a smaller
+	// budget.
```
fantapop: It seems like these comment blocks with similar content are all over.  It would be nice to eliminate or trim these down so they aren't so repetitive and large.

internal/provider/timeouts.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (v minTimeoutValidator) MarkdownDescription(ctx context.Context) string {
+	return v.Description(ctx)
+}
+
+func (v minTimeoutValidator) ValidateString(
+	ctx context.Context, req validator.StringRequest, resp *validator.StringResponse,
+) {
+	value := req.ConfigValue
+	if value.IsNull() || value.IsUnknown() {
```
fantapop: we have a helper for this checkout `IsKnown(value)`

internal/provider/timeouts.go
```diff
@@ -0,0 +1,125 @@
...
+		return false
+	}
+	return !v.IsNull() && !v.IsUnknown()
+}
+
+func appendMinimumDescription(description string, minimum time.Duration) string {
+	sentence := fmt.Sprintf(
+		"Must be at least %s. When set, this value bounds the whole operation "+
+			"end-to-end (so it can be sized to a CI/CD wall-clock limit); when "+
+			"omitted, the default budget is applied to each operation phase.",
```
fantapop: It seems like this is a docs string.  Maybe @jhlodin can take a look?   I feel like referring to parts of the processes as an "operation phase" may not be clear.  Also does this show up in the output docs?

docs/resources/cmek.md
```diff
@@ -93,6 +94,15 @@ Read-Only:
...
 - `ui_dns` (String) DNS name used when connecting to the DB Console for the cluster.
 
+
+<a id="nestedblock--timeouts"></a>
+### Nested Schema for `timeouts`
+
+Optional:
+
+- `create` (String) Must be at least 2h. When set, this value bounds the whole operation end-to-end (so it can be sized to a CI/CD wall-clock limit); when omitted, the default budget is applied to each operation phase.
+- `update` (String) Must be at least 2h. When set, this value bounds the whole operation end-to-end (so it can be sized to a CI/CD wall-clock limit); when omitted, the default budget is applied to each operation phase.
```
fantapop: @jhlodin can you take a look here?  I'm not sure it's clear how the default budget works when reading this.  Maybe it doesn't matter since we don't really explain it today.

internal/provider/timeouts.go
abhishek-kedia: Simplified it.

internal/provider/cluster_resource.go
abhishek-kedia: Fixed.

internal/provider/cluster_resource.go
abhishek-kedia: Done

internal/provider/cmek_resource.go
abhishek-kedia: Done

jhlodin (approved):
Docs LGTM

fantapop (approved):
Nice.  Looks clean.  Thank you

PR #366, opened 2026-09-08
[CNSL-2800][CNSL-2799] identify the provider on CC API requests

fantapop:
## Summary

API requests carried a bare `terraform-provider-cockroach` User-Agent with no
version, and nothing else identified the caller.

- Report the provider version and, when Terraform supplies one, the CLI
  version: `terraform-provider-cockroach/1.22.0 terraform/1.9.5`. The CLI
  version comes from `ConfigureRequest` and is dropped when empty.
- Send `Cc-Client: terraform`, which the CC API usage log prefers over the
  User-Agent heuristic that callers wrapping the SDK can override.

Locally built binaries keep the existing `dev` fallback from `main.go`, so the
version token is not guaranteed to be parseable as semver.

Second commit is unrelated tooling: `tfplugindocs` derives the provider name
from the working directory, so `make generate` fails in a git worktree. Passing
`-provider-name` fixes it. Docs regenerate unchanged.

**Commit checklist**
- [x] Changelog
- [x] Doc gen (`make generate`) — no schema change; regenerates identical
- [x] Unit test(s) — User-Agent format and the `Cc-Client` header
- [ ] Acceptance test(s) — n/a
- [ ] Example(s) — n/a


jaywritescode (approved):
LGTM!

PR #367, opened 2026-09-15
Update cockroach-cloud-sdk-go dependency to v10

jaywritescode:
Bumps the SDK from v9 to v10, which is a major version because `Cluster.Plan` became optional and is now `*PlanType`. Clusters on a Continuum edition report an edition instead of a plan, so the two are mutually exclusive and neither is always present.

Behavior change: `Read` maps a nil plan to a null `plan` attribute rather than the empty string. Outside a Continuum organization every cluster still reports a plan, so nothing changes there.

Mechanical parts of the diff:

- `/v9` to `/v10` import path across the tree, `go.mod`, `Makefile`, and `vendor/`.
- Test fixtures build `client.Cluster` literals with a plain `Plan` value, so those are wrapped in `ptr(...)`. The spec types already took a pointer in v9 and are unchanged. Where a fixture used a bare string it now uses the matching `PLANTYPE_` constant, since the line had to change regardless.
- `mock/service.go` regenerated. v10 adds methods to `client.Service`, so the old mock no longer satisfied the interface.
- `docs/resources/user_role_grant{,s}.md` regenerated; the SDK's role enum gained `AUDITOR` and `CLUSTER_DATA_ACCESSOR`.

`go build`, `go vet`, `make install`, `make generate`, and the unit suite all pass.

**Commit checklist**
- [x] Changelog
- [x] Integration test(s) - existing tests updated for the pointer change
- [ ] Acceptance test(s) - n/a, no behavior to exercise against a live cluster
- [ ] Example(s) - n/a


CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bumped version of cockroach-cloud-sdk-go from v8 to v9.
 
+- Bumped version of cockroach-cloud-sdk-go from v9 to v10.
+
```
fantapop: changelog entries should stack top down.

internal/provider/blackout_window_resource_test.go
```diff
@@ -47,7 +47,7 @@ func TestIntegrationBlackoutWindowResource(t *testing.T) {
 		Id:               standardClusterID,
 		Name:             clusterName,
 		CockroachVersion: "v23.1.0",
-		Plan:             "STANDARD",
+		Plan:             ptr(client.PLANTYPE_STANDARD),
```
fantapop, line 50: It would be nice if we could make some vars somewhere useful so we don't have to wrap these in pointers everywhere they're referenced.  But let's worry about that some other time.

fantapop (approved):
This is looking good.  The one change I think we should make right now is to fix the Changelog entry.

CHANGELOG.md
jaywritescode: Changed.

PR #368, opened 2026-09-15
[CNSL-2648] Add an edition attribute to the cluster resource

jaywritescode:
## Summary

Adds `edition` to the `cockroach_cluster` resource and data source.

Clusters in Cockroach Continuum organizations report an `edition`; clusters in organizations that are not on Continuum report a `plan`. A cluster never reports both, so the two are declared mutually exclusive, and whichever the API does not report is kept null so it does not produce a perpetual diff.

Changing the edition of an existing cluster is not currently supported, and is rejected in `ModifyPlan` — matching how `name`, `cloud_provider`, and plan type are handled here. `RequiresReplace` would destroy the cluster's data.

Shape validation runs at plan time rather than letting the apply fail: `STANDARD` is serverless, `MISSION_CRITICAL` is dedicated. The server enforces this too; this only moves the rejection earlier.

At apply time, a create that fails because the organization is not on Continuum is matched on the server's exact error message and reported as "Edition not enabled for this organization", explaining that other organizations set a `plan` instead. Every other failure falls through to the generic create error, which quotes the server verbatim.

A second commit adds the `1.23.0` changelog header, per `RELEASE.md`.

## Commit checklist
- [x] Changelog
- [x] Integration test(s)
- [x] Acceptance test(s)
- [x] Example(s)

`TestAccClusterEditionNotEnabled` runs in the existing acceptance organization, which is not on Continuum. It applies a config with an `edition`, expects the server to reject it, and asserts the provider reports the specific diagnostic rather than the generic one. No cluster is created. This pins the exact server message that Create matches on — if the message is reworded, Create falls through to the generic error and the test fails.

The success paths — round trip, and `MISSION_CRITICAL` on dedicated hardware — have no acceptance coverage, since they need an organization on Continuum. Those stay covered by mocked integration tests, along with the plan-time rejection of edition changes, shape validation, and the create-error path.

`examples/resources/cockroach_cluster/resource.tf` gains a `STANDARD` and a `MISSION_CRITICAL` edition cluster, which show up in the generated resource docs. `examples/workflows/cockroach_continuum_cluster` is a runnable workflow for an organization on Continuum, mirroring the existing per-plan workflow examples.

Manually verified against staging in both kinds of org:

- **Continuum** — created a `STANDARD` serverless cluster with `edition`; re-plan clean; data source agreed; removing `edition` from config produced no diff; changing it was refused at plan time; a legacy `plan` was refused by the server.
- **Non-Continuum** — `edition` was refused with the new diagnostic; `plan` created the cluster normally with `edition` null; setting both was refused by the conflict validator.
- **Acceptance org** — the new acceptance test was run against it directly and passes; no cluster created.


docs/resources/cluster.md
```diff
@@ -158,6 +158,9 @@ resource "cockroach_allow_list" "home" {
 - `customer_cloud_account` (Attributes) Cloud-specific details required to host the cluster in your own cloud account. Only one of `aws`, `gcp`, or `azure` may be specified. This feature is available in [Private Preview](https://www.cockroachlabs.com/docs/stable/cockroachdb-feature-availability). Contact your Cockroach Labs account team to enable this feature. (see [below for nested schema](#nestedatt--customer_cloud_account))
 - `dedicated` (Attributes) (see [below for nested schema](#nestedatt--dedicated))
 - `delete_protection` (Boolean) Set to true to enable delete protection on the cluster. If unset, the server chooses the value on cluster creation, and preserves the value on cluster update.
+- `edition` (String) Denotes the cluster's edition. Set at creation and immutable thereafter. Mutually exclusive with `plan`. `STANDARD` clusters require a `serverless` block and `MISSION_CRITICAL` clusters require a `dedicated` block. Allowed values are:
+  * STANDARD
+  * MISSION_CRITICAL
```
fantapop: I think we need to be more clear here that edition is for continuum orgs and plans are for non-continuum orgs.  The current wording makes it sound like there could be a mixed state. 

One other thing to be cautious of here.  I don't think edition is going to immutable forever.  I'm not certain but I think once we have Enterprise you might be able to switch between Enterprise and Mission Critical.  We may not want to bake wording like this into these docs.

internal/provider/cluster_resource.go
```diff
@@ -836,8 +887,22 @@ func (r *clusterResource) Create(
...
 	traceAPICall("CreateCluster")
-	clusterObj, _, err := r.provider.service.CreateCluster(ctx, clusterReq)
+	clusterObj, httpResp, err := r.provider.service.CreateCluster(ctx, clusterReq)
 	if err != nil {
+		// A 403 alone is ambiguous: an API key lacking cluster-create permission
+		// returns one too. If the wording ever changes, this falls through to the
+		// generic error, which quotes the server message verbatim anyway.
+		apiMessage := formatAPIErrorMessage(err)
+		if IsKnown(plan.Edition) && httpResp != nil && httpResp.StatusCode == http.StatusForbidden &&
+			strings.Contains(strings.ToLower(apiMessage), "edition") {
```
fantapop: This seems too loose to me.  How about checking for the exact error string?

internal/provider/cluster_resource.go
```diff
@@ -836,8 +887,22 @@ func (r *clusterResource) Create(
...
+			resp.Diagnostics.AddError(
+				"Edition not enabled for this organization",
+				fmt.Sprintf("Could not create cluster: %v\n\nThe edition attribute is only accepted for organizations "+
+					"enrolled in CockroachDB editions. Contact your Cockroach Labs account team to enable it, or remove "+
+					"the edition attribute to create a cluster on a plan.", apiMessage),
```
fantapop: This error message doesn't really make sense.  

> enrolled in CockroachDB editions

> Contact your Cockroach Labs account to enable it.  

Let's keep it simple. think we'd want something more like:

Could not create cluster.  %v\n\n The edition attribute is only accepted in Cockroach Continuum organizations.  Non-continuum organizations require a plan attribute on every cluster instead.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added the `cockroach_cluster_runtime_scanning` resource to enable or disable per-cluster runtime scanning (HBTD, host-based threat detection) enforcement.
 
+- Added an `edition` attribute to the `cockroach_cluster` resource and data source. It is set at creation and is immutable; changing it is refused at plan time. `edition` and `plan` are mutually exclusive, and the edition must match the cluster's shape: `STANDARD` requires a `serverless` block and `MISSION_CRITICAL` requires a `dedicated` block.
```
fantapop: similar comment as above.  edition may not be mutable. In addition, we should be explaining that plan is for non-continuum orgs and editoin is for continuum orgs.

fantapop (commented):
I left a handful of comments here.

docs/resources/cluster.md
fantapop: Reworded to frame this per-organization: Continuum orgs set an `edition`, orgs that aren't on Continuum set a `plan`, so the two can't both be set. Dropped "immutable thereafter" in favor of "Changing the edition of an existing cluster is not currently supported" — describes today's behavior without claiming it holds forever.

Made the matching change in three other places that had the same problem: the data source description, the CHANGELOG, and the `ModifyPlan` error, which said changing edition "isn't allowed" and borrowed the "to prevent accidental deletion of data" rationale from the neighboring guards. That reason doesn't really apply here — the change just isn't supported yet — so it now says so plainly. 🤖

internal/provider/cluster_resource.go
fantapop: Switched to an exact match against a named constant. Also removed the 403 check, per your other note about it being an auth code.

Confirmed against staging: a non-Continuum org returns exactly `the requested edition is not enabled for this organization`.

The substring check turned out to be wrong rather than just loose. A real shape error comes back as `invalid argument: Mission Critical edition clusters require dedicated hardware`, which contains "edition" — so someone who forgot the `dedicated` block would have been told their organization isn't on Continuum. That message is now a test case. 🤖

internal/provider/cluster_resource.go
fantapop: Took your wording with one edit: "Non-continuum organizations require a plan attribute on every cluster instead" became "Organizations that are not on Continuum set a plan on every cluster instead". Say the word if you'd rather have yours verbatim.

Output from a real non-Continuum staging org:

```
Error: Edition not enabled for this organization

Could not create cluster: the requested edition is not enabled for this organization

The edition attribute is only accepted in Cockroach Continuum organizations.
Organizations that are not on Continuum set a plan on every cluster instead.
```
🤖

CHANGELOG.md
fantapop: Rewritten along the same lines — leads with the org split and says "not currently supported" rather than immutable:

> Added an `edition` attribute to the `cockroach_cluster` resource and data source. Clusters in Cockroach Continuum organizations set an `edition`; clusters in organizations that are not on Continuum set a `plan` instead, so the two attributes cannot both be set. The edition must match the cluster's shape: `STANDARD` requires a `serverless` block and `MISSION_CRITICAL` requires a `dedicated` block. Changing the edition of an existing cluster is not currently supported and is refused at plan time.

🤖

fantapop (approved):
LGTM

PR #369, opened 2026-09-15
Reset account_id and plan when a serverless plan type changes

fantapop:
`account_id` is pinned to the prior state by `UseStateForUnknown`, but the
value tracks the hosting cloud project and differs between plan types — BASIC
reports an empty string, STANDARD reports the shared serverless project. Moving
a serverless cluster between the two planned the old `account_id` and failed the
apply with "Provider produced inconsistent result after apply". `plan` has the
same problem when the configuration leaves it unset.

Both are now marked unknown at plan time when the plan type changes. The new
`planTypeChanging` helper reads the target from config, since the plan can carry
a pinned value.

This is a regression from e66893f (Aug 4), which added the `account_id` plan
modifier. v1.23.0 is the first tag that contains it. It reproduces on the
`TestAccServerlessClusterResource` and `TestAccServerlessUpgradeType` acceptance
tests, both of which failed on the v1.23.0 pre-release run.

The changelog entry is under `Unreleased` — v1.23.0 is shipping without this fix.

**Commit checklist**
- [x] Changelog
- [ ] Doc gen (`make generate`) — no schema changes
- [x] Integration test(s) — `TestIntegrationServerlessPlanTypeChangeAccountId`
- [x] Acceptance test(s) — covered by the two existing tests above
- [ ] Example(s) — n/a

🤖


devlaneyr (approved):
LGTM

PR #370, opened 2026-09-15
Fix the expected error message in the maintenance window test

fantapop:
Step 1 of `TestAccMaintenanceWindowResource` asserts that setting a maintenance window on a standard cluster is refused. The API message is now "maintenance windows are not supported for standard clusters", so the `ExpectError` pattern stopped matching and the acceptance test failed. This updates that pattern and the mocked error in the integration test.

The refusal itself still works — only the assertion drifted. The message is produced entirely server-side; nothing in this repo emits the old wording.

**Commit checklist**
- [ ] Changelog — not needed, test-only
- [ ] Doc gen (`make generate`) — n/a
- [x] Integration test(s) — `TestIntegrationMaintenanceWindowResource` passes
- [ ] Acceptance test(s) — needs a CI run, not verified locally
- [ ] Example(s) — n/a


jaywritescode (approved):
LGTM

PR #375, opened 2026-09-16
[CNSL-2895] Add acceptance tests for Continuum edition clusters

linhcrl:
Existing acceptance tests only cover the legacy `plan` flow; edition (STANDARD / MISSION_CRITICAL) had only mocked or negative coverage. Add two tests that create real edition clusters against a live Continuum org and verify `edition` round-trips with `plan` null:

  - TestAccStandardEditionClusterResource (serverless)
  - TestAccMissionCriticalEditionClusterResource (dedicated)

They read a new COCKROACH_CONTINUUM_API_KEY, injected into a provider block so the Continuum org stays isolated from the rest of the suite, and skip cleanly when it is unset. The key belongs to the "Cloud API Terraform Testing - Continuum" org (ID 959e7472-db19-469f-bb9e-0cdf25ee25e9, label org-3bsq8).


internal/provider/provider_test.go
```diff
@@ -60,6 +61,36 @@ func testAccPreCheck(t *testing.T) {
...
+		t.Skipf("%s must be set to run Continuum acceptance tests", CockroachContinuumAPIKey)
+	}
+}
+
+// continuumProviderBlock emits a provider block whose API key targets a
+// Cockroach Continuum-enabled organization. Continuum acceptance tests prepend
+// it so they authenticate against a Continuum org: the block's apikey overrides
+// the COCKROACH_API_KEY env var the rest of the suite uses (see provider
+// Configure). The key is only read when the test runs, which is gated by
+// testAccContinuumPreCheck.
```
fantapop, line 85: I think we can remove all the comments.  It's all pretty self explanatory looking at the code.

internal/provider/cluster_resource_test.go
```diff
@@ -5777,3 +5663,153 @@ func TestIntegrationClusterEditionImmutable(t *testing.T) {
...
+// real STANDARD edition (serverless) cluster in a Cockroach Continuum
+// organization. It is skipped unless TF_ACC and COCKROACH_CONTINUUM_API_KEY are
+// set. This is the Continuum edition counterpart to
+// TestAccServerlessClusterResource, which only exercises the legacy plan flow.
+func TestAccStandardEditionClusterResource(t *testing.T) {
+	t.Parallel()
+	clusterName := fmt.Sprintf("%s-std-edition-%s", tfTestPrefix, GenerateRandomString(2))
+	resource.Test(t, resource.TestCase{
+		IsUnitTest:               false,
+		PreCheck:                 func() { testAccPreCheck(t); testAccContinuumPreCheck(t) },
```
fantapop, line 5727: can we incorporate the testAccPreCheck into the continuum precheck function so we just have to call it?

internal/provider/cluster_resource_test.go
```diff
@@ -5777,3 +5663,153 @@ func TestIntegrationClusterEditionImmutable(t *testing.T) {
...
+		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			// Create with edition STANDARD and verify it round-trips into state
+			// (and through the data source) with plan left null.
+			{
+				Config: continuumStandardConfig(clusterName, 2),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckCockroachClusterExists(serverlessResourceName),
```
fantapop, line 5735: From claude:

testCheckCockroachClusterExists authenticates against the wrong org. Both tests call it in their create step. That helper uses the package-level cl (internal/provider/provider_test.go:34-42), built in init() from COCKROACH_API_KEY — the non-Continuum org. Cloud API keys are org-scoped, so GetCluster on a Continuum cluster ID returns 404/403 and the check fails with “cluster does not exist” even on a successful create. The provider block only redirects the provider’s client, not the test helper’s.

suggested fix: build a Continuum client from COCKROACH_CONTINUUM_API_KEY and use a variant of the helper that takes a client, or drop the existence check and rely on the attribute assertions plus the data source (which does go through the provider).

internal/provider/cluster_resource_test.go
```diff
@@ -5777,3 +5663,153 @@ func TestIntegrationClusterEditionImmutable(t *testing.T) {
...
+resource "cockroach_cluster" "test" {
+    name           = "%s"
+    cloud_provider = "GCP"
+    edition        = "MISSION_CRITICAL"
+    dedicated = {
+        num_virtual_cpus = 4
+        storage_gib      = %d
+    }
+    regions = [{
+        name       = "us-central1"
```
fantapop, line 5706: from claude:

2. Hardcoded us-central1 conflicts with main. The branch is behind 99a8362, which moved every cluster-creating test onto testRegion/testSecondRegion/testThirdRegion because us-central1 runs out of capacity and flakes. DEVELOPMENT.md:152 explicitly says new tests should use these. Rebase and use testRegion (us-east1 works for both serverless and dedicated shapes).

internal/provider/cluster_resource_test.go
```diff
@@ -5777,3 +5663,153 @@ func TestIntegrationClusterEditionImmutable(t *testing.T) {
...
+}
+`, clusterName, provisionedVCPUs)
+}
+
+// continuumMissionCriticalConfig builds a MISSION_CRITICAL edition (dedicated)
+// cluster config plus a data source that reads it back, targeting a Continuum
+// org. node_count and storage_gib are parameterized so an update step can change
+// them in place.
+func continuumMissionCriticalConfig(clusterName string, nodeCount, storageGib int) string {
+	return continuumProviderBlock() + fmt.Sprintf(`
```
fantapop, line 5696: Embedding the continuum provider block feels wrong to me considering there is a provider factory as part of the test framework and it injects the the non continuum provider.

i.e.
```go
ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
```

At first claude thought specifying the provider inline would be okay but now it's saying it's not.   Here is the full transcript:

```
How the Continuum credential is wired

A provider block in HCL isn’t unidiomatic; plenty of providers configure credentials that way. The problem is narrower. testAccProtoV6ProviderFactories evaluates New("test")() once at package init and NewProtocol6WithError captures that pointer, so every test in the binary shares one *provider. Configure writes p.service and passes p itself to resources, which read r.provider.service at call time.

That’s been harmless because every test has filled that field with the same key. This is the first test to fill it with a different org’s key while holding t.Parallel(), so a Continuum Configure can land between a sibling test’s plan and apply. The result is a cross-org API call, nondeterministically.

Giving the Continuum tests their own factory map fixes it, because they get their own provider instance:

type continuumProvider struct {
	tf_provider.Provider
	apiKey string
}

func (p *continuumProvider) Configure(
	ctx context.Context, req tf_provider.ConfigureRequest, resp *tf_provider.ConfigureResponse,
) {
	var attrs map[string]tftypes.Value
	if err := req.Config.Raw.As(&attrs); err != nil {
		resp.Diagnostics.AddError("continuum test provider", err.Error())
		return
	}
	attrs["apikey"] = tftypes.NewValue(tftypes.String, p.apiKey)
	req.Config.Raw = tftypes.NewValue(req.Config.Raw.Type(), attrs)
	p.Provider.Configure(ctx, req, resp)
}

func continuumProviderFactories() map[string]func() (tfprotov6.ProviderServer, error) {
	p := &continuumProvider{Provider: New("test")(), apiKey: os.Getenv(CockroachContinuumAPIKey)}
	return map[string]func() (tfprotov6.ProviderServer, error){
		"cockroach": providerserver.NewProtocol6WithError(p),
	}
}
The provider block then comes out of the config entirely and the key never reaches HCL, which also handles a smaller issue: continuumProviderBlock writes the key in cleartext into the config file the framework generates. apikey is Sensitive, so plan and apply output is redacted, but the config file isn’t, and Terraform echoes source lines when reporting a config error.

A few notes on the snippet. tfsdk.Config is read-only — no SetAttribute, unlike State and Plan — so rewriting Raw is the available route; Config.Raw is exported and Value.As accepts *map[string]tftypes.Value for objects. Embedding the interface would drop optional provider interfaces, but this provider implements exactly the five core methods, so nothing is lost. The inner Configure still sets resp.ResourceData to the concrete *provider, so clusterResource.Configure’s type assertion keeps working. It compiles against framework v1.15.0.

Shaping it for what comes next. We expect Continuum to carry a growing share of acceptance coverage, with the non-Continuum set shrinking to a deliberate subset. If that’s the direction, it’s worth making the org an explicit parameter rather than a Continuum-specific helper sitting against an implicit default — a testOrg value carrying the credential, its factory map, its API client, and whether the org expresses tier as plan or edition. Moving a test between orgs then becomes a one-word change, and we avoid renaming the default in both directions when the majority flips. Not asking you to build that here, just suggesting these two helpers land in roughly that shape so the next test doesn’t have to generalize them.
```

fantapop:
There were some other smaller things my local agent reviewer flagged that you may want to consider:

* These will skip silently in CI. Neither acceptance-tests.yml nor pre-release-ci.yaml passes COCKROACH_CONTINUUM_API_KEY, so the coverage never runs and nothing reports that. Add the secret, or state in DEVELOPMENT.md that these are local-only.

* Document the env var. DEVELOPMENT.md documents the other test env vars. This one should cover which org, what permissions, and why it’s separate — the org ID and label from the PR description belong in the repo.

* ImportStateVerify: false is attributed to matching the dedicated test, but that one disables it for a specific upgrade_status / finalize_version_upgrade conflict that neither new test has. Since the point is that edition survives a round trip, ImportStateVerify: true would actually check it. Worth trying before turning it off.
nodeCount is unused variance. Both call sites pass 1, and the comment says it’s parameterized for an update step that never changes it. Drop it or use it.

* node_count = 1 for MISSION_CRITICAL — does the API accept a single-node cluster at that edition? Just want to be sure it won’t hit a server-side minimum.

* Const placement. CockroachContinuumAPIKey is exported but test-only; its siblings live in models.go.

* Naming consistency. The mission-critical test declares local resourceName / dataSourceName with values identical to the package-level serverlessResourceName / serverlessDataSourceName the standard test uses. Worth picking one.

